Issue #155, opened 2023-01-18
Measurements that error won't response to the client


pbirkhol-ni:
# Bug Report

If you run a measurement that errors, the error will not get passed back to the client. Instead, nothing will get passed to the client and the call will appear to be hung.

## Repro or Code Sample

None of the examples currently display this error. This is an issue with a relatively new VI (Invoke Measure Logic.vi)

## Expected Behavior

When a measurement errors, a gRPC status should be sent back to the client.

## Current Behavior

No response is sent to the client

## Possible Solution

Properly handle errors in Invoke Measure Logic.vi. Currently, if there is an error, the call to Set Measurement Service Measure Response.vi effectively gets skipped because it doesn't run upon incoming error. We should consider sending the status back upon incoming error.

[AB#2272710](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2272710)

aepete:
Fixed by #156

PR #156, opened 2023-01-19
Fix for Bug #155 - error reporting over grpc when G measurement errors

aepete:

This change fixes bug #155 by adjusting error handling during the Measure event.  This change also applies similar error handling to the GetMetadata event, which has evidently lacked any error reporting over grpc, previously.

- Restores a recent regression in error reporting for Measure event
- Improves error reporting for GetMetadata event in a way that's consistent with errors in Measure

Created a basic measurement and loaded it within InstrumentStudio.

Default state:
- Exercised measurement in its default state; functioned properly.

Testing error handling for GetMetadata codepath:
- Exercised measurement with having injected an error into `Get UI Details.vi`'s error rail.  InstrumentStudio enumerates the measurement but fails to load its UI, showing this banner:
image

Testing error handling for Measure codepath:
- Exercised measurement with having injected an error into `Measurement Logic.vi`'s error rail; unexpected error surfaced in InstrumentStudio:
image

- Exercised measurement with having a broken run arrow on `Measurement Logic.vi`; unexpected error surfaced in InstrumentStudio.
image

(Though the banner text is fairly generic, more details are available by hovering over the banner, as shown in screenshots.  We might consider providing better details in the tooltip for the `Get UI Details.vi` case.)

[AB#2273804](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2273804)


aepete:
Start Sync was updated (both for GetMetadata and Measure events) to inject errors into the subsequent gprc response call.  After translating the error into a grpc status, it's deliberately ignored, so as to not interfere with the code that subsequently sets the response over gprc.

(The new `No Error` cases are merely pass-through.)
 
_Measure event_:
image
- I've moved the call to `Set MeasurementService Measure Response.vi` from the subVI up into the calling event structure.
- The last piece of `Invoke Measure Logic.vi` now just merges the errors and outputs the Measure response, and it allows the caller to handle the error and send the response.
 image

_Get Metadata event_:
image
- I've updated this codepath to now call `Set MeasurementService GetMetadata Response.vi`, a wrapper autogenerated by grpc-labview, rather than directly calling `Set ni_measurementlink_measurement_v1_GetMetadataResponse.vi`, so as to be congruent to the Measure case.

dixonjoel (approved):

pbirkhol-ni (approved):

PR #163, opened 2023-01-25
Include information about the cancellable measurement example

dixonjoel:
Signed-off-by: Joel Dixon <joel.dixon@ni.com>

Adds basic information about the Cancellable measurement example. Also fixes one typo to remove 'latest'.

Give each example its own README and move the information from the root README about examples down into the README under Example Measurements.

Get the documentation up to date with this newly added example.

None - documentation only.


dixonjoel:
> Can we just delete the example blurbs here and add a README.md for every example? That is what we did for python and I think we should be consistent. For now, I would only add a README.md for the Cancellable Measurement. I am planning on deleting the other two examples. :)

Sure, I'll try to match Python. I think we listed the examples in the top-level read me, but only as simple bullet points.

dixonjoel:
> Can we just delete the example blurbs here and add a README.md for every example? That is what we did for python and I think we should be consistent. For now, I would only add a README.md for the Cancellable Measurement. I am planning on deleting the other two examples. :)

I updated this main README. I hope we can postpone the actual additions of the READMEs for the examples to other PRs

bkeryan:
> Can we just delete the example blurbs here and add a README.md for every example? That is what we did for python and I think we should be consistent. For now, I would only add a README.md for the Cancellable Measurement. I am planning on deleting the other two examples. :)

I agree that detailed information about the example should be in a per-example README.md.

I think it's ok to provide a list of examples in the top-level README.md, but we should consider moving that to examples/README.md. In the Python repo, we upload the top-level README.md to PyPI as the package description.

Source/Example Measurements/NI-DMM Measurement/README.md
```diff
@@ -0,0 +1,21 @@
+# NI-DMM Measurement
```
dixonjoel, line 1: @aepete @bkeryan I stole this from the NI-DMM Python example and removed the information about TestStand sequences. I don't know if they have the same features, so let me know how this might need to change.

docs/README.md
```diff
@@ -113,49 +113,23 @@ To run the measurements, follow the below steps
...
-2. Start the discovery service if the measurement is to be used in InstrumentStudio Measurement Framework.
-
-3. Run the following VI to run the measurement service from `LDOMeasurements.lvproj`
-    -To run Line Regulation, run `Run Service.vi` from `LineRegulation.lvlib`
-    -To run Load Regulation, run `Run Service.vi` from `LoadRegulation.lvlib`
-
-4. Use the `Measurement Framework` and run the measurements interactively in InstrumentStudio. For more details, please refer to the section [Using the measurement in InstrumentStudio](#using-the-measurement-in-instrumentstudio).
-
-Note: The Line and Load regulation measurements have not been tested with real hardware. They have only been tested with simulated instruments.
+The `Source\Example Measurements` directory contains example measurement services. See the [README.md](https://github.com/ni/measurementlink-labview/blob/main/Source/Example%20Measurements/README.md) file there for more information.
```
bkeryan, line 126: If this markdown is only served up from GitHub, you can use relative paths.

When it's uploaded to PyPI or some other service, you can't.

Source/Example Measurements/NI-DMM Measurement/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+This is a MeasurementLink example that performs a measurement using an NI DMM.
+
+## Features
+
+- Pin-aware, supporting one session, one pin, and one selected site
+- Uses the NI gRPC Device Server to allow sharing instrument sessions with other
```
bkeryan, line 8: This isn't supported in LV yet

Source/Example Measurements/CancellableMeasurement/README.md
```diff
@@ -0,0 +1,21 @@
+# CancellableMeasurement
+
+This is a MeasurementLink example that demonstrates a measurement
+that takes a specified length of time and can be cancelled while it is running.
+
+It will run for the amount of time specified with the `Delay (sec)` in the UI unless the `STOP` button is clicked.
+
+Once the measurement completes or is cancelled, it will return a random number.
```
bkeryan, line 8: The return values are ignored when the measurement is canceled.

Source/Example Measurements/NI-DMM Measurement/README.md
pbirkhol-ni: FYI, I plan on deleting this example. It does not represent best practice. It may get revived at some point. I'm okay with you adding this in case it gets revived, but am also okay with you not adding it.

Source/Example Measurements/CancellableMeasurement/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Features
+
+- Runs for the number of seconds defined in the `Delay(sec)` control
```
bkeryan, line 12:
```suggestion
- Runs for the number of seconds defined in the `Delay (sec)` control
```

Source/Example Measurements/NI-DMM Measurement/README.md
pbirkhol-ni, line 8: This is not true. It does not use grpc-device.

Source/Example Measurements/README.md
```diff
@@ -1,52 +1,32 @@
...
-
-1. To understand the LabVIEW Service and how a service function is called on client call, please refer: https://github.com/ni/grpc-labview/blob/master/docs/BasicsTutorial.md
-
-Measurement Service and the function calls are similar to how the general LabVIEW service works as mentioned in the link above. You can find the `MeasurementService` and `ServiceManagementService` under the `<libraryName>.lvlib\Service\`. 
-
-`MeasurementService` is responsible for responding to the MetaData request and calling the `Measurement Logic.vi` when the DoMeasurement request is sent from the frameworks. These implementation can be found in: `Start Sync.vi`
-
-`ServiceManagementService` is responsible for managing the service like triggering stop event when the framework request to stop the service. These implementation can be found in: `Start Sync.vi`
+Measurement logic for the measurements is contained in `Measurement Logic.vi` measurement service library.
```
pbirkhol-ni, line 27:
```suggestion
Measurement logic for the measurements is contained in `Measurement Logic.vi`.
```

I think mentioning the library is confusing. Just mention the VI name should be obvious enough.

Source/Example Measurements/README.md
```diff
@@ -1,52 +1,32 @@
...
 
 ### How to add new UI elements
 
-User Interface is created in `Measurement UI.vi` under the `<LibraryName>.lvlib\`. The control and indicator names in the User Interface should match the `Measurement Configuration` and `Measurement Results`. If the datatype and name matches, then the data from the UI will sent to the logic before execution and the results will be published to UI after measurement is run.
+The User Interface is defined by `Measurement UI.vi` under `<LibraryName>.lvlib\`. The control and indicator names in the User Interface should match the `Measurement Configuration` and `Measurement Results`. If the datatype and name matches, then the data from the UI will
```
pbirkhol-ni, line 31: Same as above. I wouldn't mention the library name. These names should be unique in the examples.

pbirkhol-ni (changes requested):

Source/Example Measurements/NI-DMM Measurement/README.md
dixonjoel: Not adding the README for NI-DMM.

docs/README.md
dixonjoel: Using a relative path.

Source/Example Measurements/CancellableMeasurement/README.md
dixonjoel: Updated to take out the 'cancelled'.

Source/Example Measurements/NI-DMM Measurement/README.md
dixonjoel: Not adding the README for NI-DMM.

dixonjoel: Removed the NI-DMM readme.

aepete (approved):

PR #167, opened 2023-01-31
Add a DCPower LabVIEW MeasurementLink example

pbirkhol-ni:

Adds a DCPower MeasurementLink example showing best practice for LabVIEW measurements.

Note that I created this example using the MeasurementLink project wizard. I borrowed as much as I could from the DCPower python example (the LabVIEW UI, readme, code flow, etc).

In the comments below, I will add a non comprehensive set of screenshots and also comment on code that I'd like additional input. For a full review, you should really pull down the example and open it in LabVIEW.

Customers will need at least one full featured example to be successful with MeasurementLink.

I installed MeasurementLink 23.0, InstrumentStudio 23.0, LabVIEW 2020SP1, TestStand 2021, and DCPower 23.0 on a clean machine. I installed the VI Packages for MeasurementLink. Then I copied over the example and tested it in both InstrumentStudio and LabVIEW.


pbirkhol-ni:
Type specializations for the pin control:
image

pbirkhol-ni:
Measurement details:
image

pbirkhol-ni:
Measurement Logic:
image

pbirkhol-ni:
Create and Register Sessions code module:
image

pbirkhol-ni:
Unregister and Destroy Sessions code module:
image

pbirkhol-ni:
Update Pin Map code module
image

pbirkhol-ni:
DC Power specific logic. This gets called from the `Measurement Logic.vi`.
image

pbirkhol-ni:
Areas of note:
- When creating sessions in the code module, I create them and then immediately detach from them. I think that is the right behavior, but it is hard to map directly to the python example since python auto handles this.
image

- When reserving and unreserving sessions, I typically wire the Response Data of reserve to the Request Data of unreserve. This cleans up the diagram considerably, but is probably bad practice. Note that there is a coercion dot that shows the type mismatch.
image

- When creating sessions in the code module, I don't create a session if it already exists and I only register the sessions that I created. So for example, if the code module is supposed to create two sessions but one already exists, I will skip it and only create and register the session that doesn't exist. This is different from how python behaves (python assumes that none of the sessions exist). I don't know that it matters either way, but the LabVIEW code seems safer.
image

- In the TestStand sequence, I don't just assign the PinMapPath file global to the PinMapId temporary global like the python examples. Instead, I read the pin map id from the code module, assign that to a PinMapId file global, and then assign that to the PinMapId temporary global. According to Jonathan Mendez, this should be considered best practice and I filed a bug to python to update the examples.
- The code modules are part of the `NI-DCPower Source DC Voltage.lvlib`. This seems okay, but perhaps could cause build problems if/when trying to build the code modules into a PPL/exe?

dixonjoel:
Why do the calls into resolve service have two copies of the same service class?

dixonjoel:
I get this type conflict when opening in TestStand. I suspect it's a problem with my dev machine - I didn't check which MeasurementLink I even have installed, but just wanted to verify you've been able to open the sequence on a test machine and ask which version(s) of MeasurementLink work with this sequence file.

image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Source DC Voltage.lvproj
```diff
@@ -0,0 +1,181 @@
+﻿<?xml version='1.0' encoding='UTF-8'?>
```
dixonjoel, line 1: Why the difference in file name for this .lvproj file? All the others are NIDCPowerSourceDCVoltage*.

dixonjoel:
How much failure testing have you done to see how well the example handles and communicates error states? E.g. what if the discovery service is not running when you run the example? What if the pin map service is not installed? etc.

I don't think this kind of testing needs to hold up the submission of the example, but I think we should try a few of the common errors a user may experience and see how the error reporting is and if it's reasonable to help the user understand what's wrong.

dixonjoel:
Have you tested the example by building it into a .exe to flush out any errors that may occur in the runtime but not in the dev environment?

Source/Example Measurements/README.md
```diff
@@ -14,6 +15,9 @@ To run the example measurement, follow these steps
 
 1. Open the LabVIEW project (.lvproj) file for the example.
 2. Open and Run the following VI to run the measurement service: `Run Service.vi`
+3. For examples that use a pin map, the name of the device in MAX must match the name of the device in the pin map.
```
dixonjoel, line 18: I think this should be 'NI MAX'.

dixonjoel:
Did you intentionally leave out the .measui and .measproj from the Python example? I don't think it has to be included in this initial checkin, but I think it would be nice to show that a LabVIEW service can use .measui or LV UI just like a Python service.

bkeryan:
> Why do the calls into resolve service have two copies of the same service class?
> 
> Edit: It looks like it's the provided interface and service class. Are these always the same for a LV service?

The provided interface and service class are the same for "singleton" services like Session Management.

pbirkhol-ni:
> I get this type conflict when opening in TestStand. I suspect it's a problem with my dev machine - I didn't check which MeasurementLink I even have installed, but just wanted to verify you've been able to open the sequence on a test machine and ask which version(s) of MeasurementLink work with this sequence file.
> 
> image
> 
> Followed by image

I wrote the example targeting the new NI_MeasurementLink.ini file. Once that is in the installers you shouldn't get that message. However, we may start getting that message with the python examples since they target the old name.

pbirkhol-ni:
> Did you intentionally leave out the .measui and .measproj from the Python example? I don't think it has to be included in this initial checkin, but I think it would be nice to show that a LabVIEW service can use .measui or LV UI just like a Python service.

It wasn't intentional. I would be surprised if someone writing a LabVIEW measurement uses a .measui UI, but it is definitely possible. I can add one.

pbirkhol-ni:
> Have you tested the example by building it into a .exe to flush out any errors that may occur in the runtime but not in the dev environment?

No, I haven't. I will test that. It gets a little tricky with the code modules. However, I can test that the measurement works. The strategy for deploying the code modules will be decided later.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Source DC Voltage.lvproj
pbirkhol-ni: I created the example using the measurement template before I started stealing files from the python example. I will go ahead and change the project and library name to be consistent.

pbirkhol-ni:
> > Why do the calls into resolve service have two copies of the same service class?
> > Edit: It looks like it's the provided interface and service class. Are these always the same for a LV service?
> 
> The provided interface and service class are the same for "singleton" services like Session Management.

I followed the python example and provided the interface and service class. Technically it should work with one or the other, but I think it is safer to provide both. Since they are the same string, I could have a single string and bundle it twice into the cluster, but I kind of like the cluster constant better.

bkeryan:
Do not pass timeout=0 to ReserveSessions and ReserveAllRegisteredSessions. Either choose a timeout or pass timeout=-1 for infinite. The Python examples pass timeout=-1.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,34 @@
...
+		<Item Name="MeasurementPlugin.lvclass" Type="LVClass" URL="../Advanced/MeasurementPlugin.lvclass"/>
+	</Item>
+	<Item Name="subVIs" Type="Folder">
+		<Item Name="Convert Location to Address String.vi" Type="VI" URL="../subVIs/Convert Location to Address String.vi"/>
+		<Item Name="DCPower Specific Logic.vi" Type="VI" URL="../subVIs/DCPower Specific Logic.vi"/>
+		<Item Name="Location to Grpc Device.vi" Type="VI" URL="../subVIs/Location to Grpc Device.vi"/>
+	</Item>
+	<Item Name="TestStand Code Modules" Type="Folder">
+		<Item Name="Create and Register Sessions.vi" Type="VI" URL="../TestStand Code Modules/Create and Register Sessions.vi"/>
+		<Item Name="Unregister and Destroy Sessions.vi" Type="VI" URL="../TestStand Code Modules/Unregister and Destroy Sessions.vi"/>
```
bkeryan, line 23: Do not pass timeout=0 to ReserveSessions and ReserveAllRegisteredSessions. Either choose a timeout or pass timeout=-1 for infinite. The Python examples pass timeout=-1.

This also applies to Measurement Logic.vi.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,34 @@
...
+		<Item Name="Update Pin Map.vi" Type="VI" URL="../TestStand Code Modules/Update Pin Map.vi"/>
+	</Item>
+	<Item Name="Get Measurement Details.vi" Type="VI" URL="../Get Measurement Details.vi"/>
+	<Item Name="Get Type Specializations.vi" Type="VI" URL="../Get Type Specializations.vi"/>
+	<Item Name="Get UI Details.vi" Type="VI" URL="../Get UI Details.vi"/>
+	<Item Name="Measurement Configuration.ctl" Type="VI" URL="../Measurement Configuration.ctl"/>
+	<Item Name="Measurement Logic.vi" Type="VI" URL="../Measurement Logic.vi"/>
```
bkeryan, line 30: Each call to the measurement opens three gRPC channels: discovery, session management, and grpc-device. If we don't already have a LV gRPC channel pool in the backlog, we should add it.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,34 @@
...
+
+</Property>
+	<Item Name="Advanced" Type="Folder">
+		<Item Name="Build Assets" Type="Folder">
+			<Item Name="Post-Build Action.vi" Type="VI" URL="../Advanced/Build Assets/Post-Build Action.vi"/>
+		</Item>
+		<Item Name="MeasurementPlugin.lvclass" Type="LVClass" URL="../Advanced/MeasurementPlugin.lvclass"/>
+	</Item>
+	<Item Name="subVIs" Type="Folder">
+		<Item Name="Convert Location to Address String.vi" Type="VI" URL="../subVIs/Convert Location to Address String.vi"/>
```
bkeryan, line 17: Add "insecure" to the VI name to indicate that it uses the insecure port

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,34 @@
...
+		<Item Name="DCPower Specific Logic.vi" Type="VI" URL="../subVIs/DCPower Specific Logic.vi"/>
+		<Item Name="Location to Grpc Device.vi" Type="VI" URL="../subVIs/Location to Grpc Device.vi"/>
```
bkeryan, line 19: Add some form of NI-DCPower to the VI name and update it to use `niDCPower gRPC Device Server.ctl`

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
bkeryan, line 30: FYI: Consider doing the ResolveService calls in parallel, at least for Measurement Logic.vi.

pbirkhol-ni:
> > Have you tested the example by building it into a .exe to flush out any errors that may occur in the runtime but not in the dev environment?
> 
> No, I haven't. I will test that. It gets a little tricky with the code modules. However, I can test that the measurement works. The strategy for deploying the code modules will be decided later.

I built it into an exe, put it in the %Program Data% directory, and everything just worked.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,34 @@
...
+	</Item>
+	<Item Name="TestStand Code Modules" Type="Folder">
+		<Item Name="Create and Register Sessions.vi" Type="VI" URL="../TestStand Code Modules/Create and Register Sessions.vi"/>
```
bkeryan, line 22: The behavior of the Python TestStand fixture is to ignore "session_exists" and always initialize. That way, if the session exists, the grpc-device server will return an error.

bkeryan (changes requested):

pbirkhol-ni:
> > Did you intentionally leave out the .measui and .measproj from the Python example? I don't think it has to be included in this initial checkin, but I think it would be nice to show that a LabVIEW service can use .measui or LV UI just like a Python service.
> 
> It wasn't intentional. I would be surprised if someone writing a LabVIEW measurement uses a .measui UI, but it is definitely possible. I can add one.

I added a .measui. Unfortunately, LabVIEW only looks for UI files in a sub directory so the .measui file and .measproj files are in different directories. I don't think this is a problem, but I can file an issue so that LabVIEW looks in more places for the UI files.

pbirkhol-ni:
> How much failure testing have you done to see how well the example handles and communicates error states? E.g. what if the discovery service is not running when you run the example? What if the pin map service is not installed? etc.
> 
> I don't think this kind of testing needs to hold up the submission of the example, but I think we should try a few of the common errors a user may experience and see how the error reporting is and if it's reasonable to help the user understand what's wrong.

If you specify the wrong resource name, you get a very descriptive error. If the session manager or pin map service aren't installed the error is less clear but it would be possible to figure it out. Overall, all errors get reported, but they may or may not be helpful. I don't think they are any better/worse than errors for other examples.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: I talked to Jonathan Mendez about this. He didn't make a recommendation, but told me that a 0 timeout means that we will try to reserve a session and if it is already reserved it will immediately throw an error. That seems reasonable to me, especially in the code module where we expect the sessions to be available. I would argue that -1 is the worst behavior because it will cause your application to hang and you won't know why. If my understanding is incorrect, let me know.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: Okay. I will make that change.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: I will check and add one if not. I agree that this isn't ideal.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
bkeryan: Ok, how about 60000 (1 min) for the measurement and 0 for the code module? We should do the same in Python.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: Interesting. I didn't know that ctrl existed. Why don't the DCPower MeasurmentLink VIs use it? 

Do you still think this conversion VI should use the typedef ctrl even though the Init and Attach VIs don't use it? It will result in the a coercion dot, which isn't the end of the world.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: Sounds reasonable to me. I while file an issue to python.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
dixonjoel: They should use it.
[AB#2287962](https://dev.azure.com/ni/DevCentral/_workitems/edit/2287962): NI-DCPower gRPC APIs shall use typedef control `niDCPower gRPC Device Server.ctl` in LabVIEW API for initialize and attach.

dixonjoel: I don't know if it can be changed now since NI-DCPower has shipped, but I don't see a reason for us not to use it.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
pbirkhol-ni: I created a new feature in the Feature Debt epic.

pbirkhol-ni: I parallelized it in the measurement logic.

bkeryan:
Let's also add comments explaining the timeouts, if you didn't already (I didn't open the VIs). Something like this:

- teststand_fixture.py: This code module sets up the sessions, so error immediately if they are in use.
- measurement.py: If another measurement is using the session, wait for it to complete. Specify a timeout to aid in debugging missed unreserve calls. Long measurements may require a longer timeout.

pbirkhol-ni:
@bkeryan @dixonjoel I have been discussing this example with Ryan and Jack the last few days. After looking into a few additional options, we decided to leave it as is. I have responded to all of the PR comments and am now just waiting on approval.

dixonjoel (approved):

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/NIDCPowerSourceDCVoltage.lvlib
```diff
@@ -0,0 +1,35 @@
...
+		<Item Name="Unregister and Destroy Sessions.vi" Type="VI" URL="../TestStand Code Modules/Unregister and Destroy Sessions.vi"/>
+		<Item Name="Update Pin Map.vi" Type="VI" URL="../TestStand Code Modules/Update Pin Map.vi"/>
+	</Item>
+	<Item Name="Get Measurement Details.vi" Type="VI" URL="../Get Measurement Details.vi"/>
+	<Item Name="Get Type Specializations.vi" Type="VI" URL="../Get Type Specializations.vi"/>
+	<Item Name="Get UI Details.vi" Type="VI" URL="../Get UI Details.vi"/>
+	<Item Name="Measurement Configuration.ctl" Type="VI" URL="../Measurement Configuration.ctl"/>
+	<Item Name="Measurement Logic.vi" Type="VI" URL="../Measurement Logic.vi"/>
```
bkeryan, line 30: Cosmetic: the cluster input to DiscoveryService ResolveService.vi appears to be running right-to-left under the subVI. 
image

bkeryan (approved):

aepete (approved):

PR #181, opened 2023-02-23
Improve error handling in Run Service.vi by chaining errors

### LINKED ISSUE #99: `Run Service.vi` does not chain registration errors properly
When `Run Service.vi` fails to register with the discovery service, it doesn't display an error until you click the `STOP` button.

[AB#2270486](https://dev.azure.com/ni/DevCentral/_workitems/edit/2270486)
Link added manually by dixonjoel

dixonjoel:

Run Service.vi, the main VI of a measurement service had a number of error handling deficiencies.
* If the discovery service wasn't running, the front panel of the service would stay running, showing a port. It wouldn't show any error until you hit the Stop button. Now, the service stops immediately with this error
image

* Other errors from user VIs such as Get Measurement Details.vi would not show errors. Now you get a dialog such as this immediately and the service stops
image

fixes #99 

Common error conditions should be made clear to the user and fatal ones should immediately stop the service to give feedback.

Manual testing of error conditions after building the measurement service package.


dixonjoel:
Run Service.vi main flow now has mostly a single error rail. The exception is that we still unregister from the discovery service if there was an error elsewhere down the line.
image

dixonjoel:
Register UI Stop Event.vi now stops immediately if there is an upstream error.
image

image

aepete:
> Run Service.vi main flow now has mostly a single error rail. The exception is that we still unregister from the discovery service if there was an error elsewhere down the line.

In `Run Service`, I notice that the error line out of `Stop Server` terminates when it hits the flat sequence.  Even if we deliberately want errors on that wire to be ignored by the discovery unregistration calls, should we best merge it into main error line afterward, so any errors that may occur therein have a means to surface in the simple error hander call?
image

dixonjoel:
> > Run Service.vi main flow now has mostly a single error rail. The exception is that we still unregister from the discovery service if there was an error elsewhere down the line.
> 
> In `Run Service`, I notice that the error line out of `Stop Server` terminates when it hits the flat sequence. Even if we deliberately want errors on that wire to be ignored by the discovery unregistration calls, should we best merge it into main error line afterward, so any errors that may occur therein have a means to surface in the simple error hander call? image

Yes, I think you're right. We should be merging those errors. I'll take a look at the old one and it was probably doing that.

aepete:
.. and since the Unregister call itself is doing cleanup, we should execute it with best-effort, regardless of whether an error had occurred beforehand.  So, I think we should route the error wire _around_ that call (effectively always passing `no error` into the node).  

If I recall, gprc-labview generates the error handling within the Discov. Destroy node so as to execute regardless of whether there was an incoming error.  It'd be nice if we had a way to customize that paradigm of error-handling into the auto-generated Unregister node (... so that callers could wire it up with standard error tracks...), but that customization isn't yet supported in the proto files / generator.

aepete (changes requested):
Recommend routing error wire around the unregister call (prior to merging the error tracks); this ensures the service makes best-effort to unregister itself as it's imminently shutting down.

dixonjoel:
@aepete Not convinced I did this right. How does this look?
image

aepete (approved):

aepete:
That's pretty much what I had in mind, @dixonjoel !

Issue #165, opened 2023-01-25
Improve generator's handling of cases where conflicting measurement / resources exist


aepete:
# Bug Report

We have a bit of logic in `Valid Measurement Service Name.vi` that helps prevent problematic collisions, but it only checks against libraries present in the active project and against a (stale) handful of reserved names (see also #187).  There are several reasonable workflows whereby the generator allows the user to proceed despite an eminent & potentially destructive collision:
- If the user has a colliding library on disk but not in the project, their disk content will be overwritten by the generation process, without ever having provided the user with any notice.  (Due to the destructive behavior, I'd think we should either fail entirely or provide a warning / opt-out.)
- If the user has a colliding build specification in their project (a build spec with the same name as their pending measurement plug-in), the generation process throws an error -4500 dialog.  The generator dialog continues running, too, which I didn't expect.
![image](https://user-images.githubusercontent.com/88401200/214694018-f0bb3316-b120-4355-ba03-e888966a0568.png)

## Repro or Code Sample

Here is the state of the name-checking VI at time of writing, with problematic portions highlighted:
![image](https://user-images.githubusercontent.com/88401200/214695775-34cfe681-6da1-4c70-969c-d6091cecddd3.png)
- Orange circled area only checks for collisions within the project, and it fails to check on disk.  There is also no logic to check against build spec names that might already exist in the project.
  - Note that we'd likely employ better code re-use here by calling `Get Measurement Library.vi`, instead of a series of VI Server nodes dealing with project resources.
- Red circled area contains a stale set of reserved names, tracked by separate bug #187

## Possible Solution

I see generally three options for how to solve this issue:
1. Detect would-be conflicts as user makes value-change events on the `Measurement Plugin Name` control and disallow them to proceed to generate the measurement whenever conflicts are detected.  (This approach would resemble how we handle cases when users enter disallowed characters into the name.)
2. Refrain from checking for on-disk or build spec conflicts until after the user has elected to generate.  Upon detecting conflicts, prompt the user to either overwrite or abort.  Upside: users might find this a natural workflow as a means to regenerate their measurement from scratch.
3. This option appears least attractive, but I've included it for completeness -- to assign uniquified names / disk locations in order to essentially avoid conflicts.  However, this might prove confusing, since we typically want for the disk locations and build spec names to match the assigned Plugin name -- at least when the measurement is freshly-generated.

[...]

PR #186, opened 2023-03-01
Bugfix and new tests pertaining to measurement plugin naming

aepete:

- Updates the list of reserved names which are disallowed as measurement plugin names
  - This constitutes a fix to #187
- Adds new g-test ensuring that attempted usage of a reserved name reports as invalid by `Valid Measurement Service Name.vi`
  - Blank project added as a test asset
- Adds new g-test ensuring that attempted usage of a conflicting name (in case that a library of the requested name already exists in the project) reports as invalid by `Valid Measurement Service Name.vi`
  - Associated project & libraries added as test assets
  
- The bugfix improves product quality, disallowing the user from introducing a namespace conflict via their chosen name
- The tests reduce per-drop & per-release manual test burden, are faster to execute, and cover test cases more exhaustively than we'd typically expect from hand-testing.

- New G tests ran & passed
- Exercised generator dialog, with having attempted conflicting names and reserved names.  Both are rejected, as desired.


Source/Tests/Tests.Generator/Dialog/TestAssets/BlankProject.lvproj
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version='1.0' encoding='UTF-8'?>
```
dixonjoel, line 1: Is another option just to have the test script a blank project in a temp directory? What are the advantages of doing it this way?

dixonjoel (approved):
To make sure I understand, the current expectation is for someone to open each of the test VIs by hand and run them before release, correct?

dixonjoel:
@aepete If you aren't addressing all of the issues in #165 with this PR, I'd suggest splitting that issue into two so we can close out this part of it.

Source/Tests/Tests.Generator/Dialog/TestAssets/BlankProject.lvproj
aepete: You could script it -- and for this particular test that shouldn't be a problem...  But I can anticipate that we might want to execute other tests against a blank project that has disk presence, due to the generator's requirement that it's running in the context of a project that has been saved.  So it just struck me as simplest to put a blank project on-disk as a test asset for all to share.

dixonjoel: Ok. I'm fine with that if we think it has usefulness in the future.

aepete:
> @aepete If you aren't addressing all of the issues in #165 with this PR, I'd suggest splitting that issue into two so we can close out this part of it.

As suggested, forked issue with reserved names into a separate bug (#187), updated #165 and this PR accordingly.

PR #188, opened 2023-03-02
Update generator to reject plugin names containing disallowed chars

aepete:

- Refactors generator's `Valid Measurement Service Name.vi` to explicitly reject names that contain disallowed characters
  - Utility VI `Get Measurement Library.vi` has been adjusted in order to be utilized by (and thus reduce duplicate code in) `Valid Measurement Service Name.vi`
- Updates `Create Measurement Service.vi` to automatically correct user-provided names which might happen to have (disallowed) leading or trailing whitespace.  Also, improves the message shown if a user enters an invalid name.
- Adds new G tests to ensure proper support for both allowed and disallowed measurement plugin names.

- User of generator dialog benefits from better handling of & response to the name they've entered.
- G tests reduce manual test burden are substantially more exhaustive, in terms of coverage, than what we could expect from typical testing by-hand.

- New G tests pass.
- Existing G tests still pass (these tests cover other naming issues: collisions & reserved names).
- Exercised dialog by-hand for several allowed & disallowed requested names; all behaved as expected.

### Additional Notes

- Previously, there was no explicit check for disallowed characters in pending measurement plugin names.  Rather, the generator attempted create a library on disk having the requested file name, and the name would be rejected at that point if it were disallowed by the OS.  To be consistent with what has shipped already, the disallowed character set that's been adopted in this change are consistent with [the set disallowed in Windows filenames](https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names).


aepete:
### Changes to `Create Measurement Plugin.vi`

This is what the generator now shows when the user has entered an invalid name:
image

#### Diagram Changes
- Assigned new text constant.  Changed to a boolean glyph indicator (rather than string with red text) for aesthetics.
image
- Auto-correct for leading & trailing whitespace in the event-handling case upon name value changed.
image

aepete:
### Changes to `Valid Measurement Service Name.vi`

Updated diagram:
image

Cases not visible in the first screenshot:
- Conflicting library in project (outer-most case) -> Rejected
image
- Empty name or name contains whitespace (nested case) -> Rejected
image
  - Note that since the caller trims whitespace, the leading / trailing whitespace cases should never get encountered by users here.  But the utility is written to reject them nonetheless.
- Requested name matches a reserved name (innermost case) -> Rejected
image

aepete:
### Updates to `Get Measurement Library.vi`
- These changes allow callers to refrain from getting an error if the requested library is not found, which is helpful when making a call to check if a prospective library name is unique.
- Also updated matching to be case-insensitive, which is appropriate, since LV uses case-insensitive uniqueness for resources such as libraries.
image

aepete:
New test for checking that allowed names report as valid:

image

New test for checking that disallowed names report as invalid:
image

(unshown cases report a failure while appending the offending information about the offending name into the report.)

dixonjoel (approved):

PR #191, opened 2023-03-17
Add Error Strings global variable to consolidate error messages

### LINKED ISSUE #95: Document or single-source error codes
Are this component's error codes documented anywhere? Some VIs use error codes in the 50xx range, such as 5034 and 5050, but I can't easily tell which error codes in that range are unused. I used -1 in a couple of new VIs (#93).

[AB#2270595](https://dev.azure.com/ni/DevCentral/_workitems/edit/2270595)
Link added manually by dixonjoel

dixonjoel:

Fixes #95 

Adds an Error Strings Global.vi to centralize the error strings in the MeasurementLink Measurement Server library code.
image

Here's a typical usage of the variable. The caller of `Error Cluster from Error Code.vi` ensures the global string and the error code number match.
image

Centralize and document the error codes used throughout the library so it's easy to see all error codes used and their descriptions.

Built the packages locally, installed, and tested typical error conditions. Here's the error dialog for the discovery service not running:
image


aepete (approved):
Seems like a reasonable approach, and there's prior art for having done so (Veristand's error handling also uses a VI global with strings.)

PR #192, opened 2023-03-17
Show the service Display Name in the UI shown when running the service

dixonjoel:

Show the Display Name of the measurement service in the dialog when the service is running. Prior to this, it was impossible to tell which service was which if there were multiple running since it only displayed the Listening Port. Adding the Display Name to this.
Here's an example of two services running at the same time with the update.
image

I added a `Get Measurement Details.vi` to the base class for measurement services. This is used in `Run Service.vi` to set the display name on the front panel when launching. It is also used in a couple other places that had similar code to run the user's `Get Measurement Details.vi` and gather the information.

Improves the LabVIEW UI of a dynamically run service.

Built and installed the LV packages, generated two services and ran them to see the display name showing up.


dixonjoel:
Get Measurement Details.vi block diagram
image

aepete (changes requested):
Two requests:
1. The new `MeasurementPluginService.lvclass:Get Measurement Details.vi` has a coercion dot on its `Measurement Details` indicator terminal, because it's not linked to the respective typedef (`MeasurementLink Measurement Server Internal\RPC Messages\ni_measurementlink_measurement_v1_MeasurementDetails.ctl`).  I'd recommend linking it -- and since this affects the class interface, you'll have to resave the .lvclass file after saving the VI.
image

1. I discovered that the VI `Runtime\MeasurementLink Measurement Server\Helpers\Register to Discovery Service.vi` also makes a call into the customer's `Get Measurement Details.vi`, therefore it'd be best for that code path to exercise the new entrypoint on the class.
image

dixonjoel:
> 

Updated these VIs.

aepete (approved):
Looks great!

dixonjoel:
> One note for potential discussion and/or posterity:
> 
> This modification carries a probably-innocuous change in the interface to the customer's `Get Measurement Details.vi`; we no longer send an `error in` to the customer's VI under any circumstance, and we no longer retrieve an `error out` from the user's VI if they were to throw one. One could argue there was never much value to having `error in` | `error out` in `Get Measurement Details.vi` (aside from dataflow that comes from error tracks in LV, which is no longer applicable, since it's not called as a subVI anymore..).
> 
> Perhaps it's fine to effectively remove them from the interface, which would technically allow us to remove the `error in` | `error out` terminals from the template version of the VI. The option on the other end of the spectrum would be to incorporate both `error in` and `error out` in the read/write by-label VI Server calls. It probably won't hurt to leave it all as-is, though one might reasonably consider an un-caught error coming from the customer's `Get Measurement Details.vi` as unexpected behavior.

IMO, it's very unlikely errors will be a concern with such a simple VI, so I'm ok with this change.

PR #196, opened 2023-04-14
Add new SessionMgmt test verifying error thrown when attempting two consecutive reservations

aepete:

Adds new SessionMgmt test verifying error thrown when attempting two consecutive reservations.

Reservation is a pivotal aspect of the Session Management API, and users are reasonably prone to exercise the API in the same way as is exercised in this new test.  Presence of the test should help ensure that this reservation behavior of the Session Management Service (and its usage from LV clients) is as expected and does not undergo regression.

- Ran test by-hand; passes.
- Ran test by-hand without having started the discovery service.  Test fails as expected with helpful output.
- Ran test by-hand in the context of the full set of Tests.lst.  Test succeeds along with the rest.
- Test runs as part of the action invoked by this PR.


aepete:
Test VI's diagram:
image

- Unseen cases all report `false` for `pass` along with an explanation for the failure into `result`:
  - image
  - image
  - image

Source/Tests/Tests.lst
dixonjoel, line 1: Is it necessary to maintain this list of tests? Could we instead just use something that enumerates all VIs (maybe with a file pattern of, e.g. "Test_*"?

dixonjoel (approved):

Source/Tests/Tests.lst
dixonjoel: Sure, we can defer this question from this PR.

PR #201, opened 2023-05-02
Add V2 measurement service

pbirkhol-ni:

This PR adds the V2 service. This is the output of grpc-labview, unmodified, and is not yet called by anything. 

The existing V1 library is called `MeasurementLink Measurement Server Internal`. The new V2 library is called `MeasurementLink Measurement Service V2`, which is not consistent with the old name. However, I think the new name is better. Ideally we would rename the old library, but I don't know if that is possible without breaking compatibility. 

[AB#2381074](https://dev.azure.com/ni/DevCentral/_workitems/edit/2381074)

This is the first step in supporting the V2 measurement service.

None


dixonjoel:
As far as releasing this in .vips. If we were to build the Runtime .vip, would this get installed already? Or is there additional changes to the .vipb file to ship this? We can do that in a separate PR if it's even necessary.

dixonjoel (approved):

pbirkhol-ni:
> > As far as releasing this in .vips. If we were to build the Runtime .vip, would this get installed already? Or is there additional changes to the .vipb file to ship this? We can do that in a separate PR if it's even necessary.
> 
> I'll try to build this into a VIP. But looking at the VIP file, I see this, so I think it just harvests everything in that directory. <Library_Source_Folder>..\Runtime</Library_Source_Folder>

I just confirmed that it gets automatically added to the VIP.

PR #202, opened 2023-05-02
Add support for V2 measurement services

pbirkhol-ni:

This PR creates a MeasurementServiceV2 class that implements the IMeasurmentService interface. The MeasurementServiceV2 class is very similar to the MeasurementServiceV1 class except that it uses V2 messages/VIs.

Since the V2 proto file adds type annotations to the Output message, this PR also modified the shared Output.ctl to include the type annotation cluster. Because of this, a conversion VI (`Convert Metadata Response to V1`) was added to MeasurementServiceV1 to convert the shared Output.ctl to the V1 specific Output.ctl. 

`Run Service.vi` was modified so that it can service both V1 and V2 requests.

Add V2 support is the next step to supporting continuous measurements.

I ran the shipping examples and made sure they work with both V1 and V2 clients.


dixonjoel:
I suggest casting the V1 and V2 measurement services to IMeasurementService explicitly in Run Service.vi so it's clearer that we're working with that generically.

pbirkhol-ni:
> I suggest casting the V1 and V2 measurement services to IMeasurementService explicitly in Run Service.vi so it's clearer that we're working with that generically.

Already done in my latest push.
image

PR #204, opened 2023-05-04
Add support for continuous measurements

pbirkhol-ni:

[AB#2376465](https://dev.azure.com/ni/DevCentral/_workitems/edit/2376465)

This PR adds support for running continuous measurements.
- Adds a new method to the MeasurementContext class called `Update Results.vi`. The name of this new method mirrors the name of the output cluster, which is called `Measurement Results`. This new method takes a variant for the output cluster to be updated.
- The MeasurementContext class now contains both an IMeasurementService interface and a MeasurementPluginService class. Those two new data members are injected during creation. When a measurement author calls the  `Update Results.vi` on the MeasurementContext class, the call will be forwarded to an `Update Results.vi` method that is on the IMeasurementService interface. This call will be dispatched to either the V1 or V2 implementation, depending on what was passed in during creation.
- As alluded to above, there is now an `Update Results.vi` method on the IMeasurementService interface. The V1 implementation is a no-op and the V2 implementation calls `Pack for gRPC` followed by `Set Output`. The `Streaming?` input is hardcoded to true.
- I missed changing a V1 control in `Get Measurement Details.vi` so I changed it to be the version agnostic control. This should have been done in my previous refactoring PR.
- The changes to `MeasurementServiceBaseV1.lvclass` and `Read Pin Map Context.vi` shouldn't be necessary, but LabVIEW insists on the dirty dot.

With this PR, we officially support continuous measurements in LabVIEW.

I wrote a new example that does continuous updates. It will be submitted in another PR.


Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Update Results.vi
dixonjoel, line 1: Why does this take a variant instead of a 'Measurement Results' cluster like Measurement Logic.vi? How does the user get the variant typically? (since we don't have an example yet). Do they call pack or something?

Or can you wire the measurement results directly into this input?

dixonjoel (approved):

pbirkhol-ni:
> With your example, have you gotten to the point where you've successfully seen continuous updates in InstrumentStudio with locally built packages?

Yep! It all works.

pbirkhol-ni:
FYI @jasonmreding

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Update Results.vi
jasonmreding: The grpc-labview pack/unpack APIs take a variant anyway so we would have to perform the conversion to variant at some point anyway.

jasonmreding:
These changes look good to me.

jasonmreding (approved):

PR #206, opened 2023-05-12
Add UI Progress Updates example

pbirkhol-ni:

[AB#2374853](https://dev.azure.com/ni/DevCentral/_workitems/edit/2374853)

This PR adds an example that shows how to update the UI for a LabVIEW measurement. It also adds a Readme and updates the existing example Readme.

This example is modeled after the equivalent python example.

We want to create examples showing how to use major features.

Manual testing with a prerelease version of InstrumentStudio.


pbirkhol-ni:
Measure Logic

image

Source/Example Measurements/UIProgressUpdates/UIProgressUpdates/Advanced/Build Assets/Post-Build Action.vi
dixonjoel, line 1: image
Seems to only have the V1 measurement service. Did you try building and statically registering to see that we get continuous updates?

dixonjoel (approved):

Source/Example Measurements/UIProgressUpdates/UIProgressUpdates/Advanced/Build Assets/Post-Build Action.vi
pbirkhol-ni: Did you click the arrow to show the next element in the array? It should show v2. However, if your vi.lib VIs are old, it would only show v1. 

To answer your question directly, though, I did test a built exe to make sure it works.

dixonjoel: Yeah, I didn't realize that was in vi.lib. My packages are not up-to-date.

bkeryan:
Outside the loop, update the status to "Total samples: %d" so that we can verify during testing that the final update was sent. Otherwise, there is no visual indication whether you are looking at the final update or the next-to-last update.

Source/Example Measurements/UIProgressUpdates/UIProgressUpdates/Measurement Logic.vi
bkeryan, line 1: Outside the loop, update the status to "Total samples: %d" so that we can verify during testing that the final update was sent. Otherwise, there is no visual indication whether you are looking at the final update or the next-to-last update.

This also makes it unnecessary to skip the next-to-last update, since the final update differs by the status field. :)

bkeryan: Thanks, I didn't download the VIs and I assumed the other case was empty.

Issue #199, opened 2023-04-28
NIDCPowerSourceDCVoltage example silently ignores multiple channel groups


bkeryan:
# Bug Report

If you call NIDCPowerSourceDCVoltage on multiple pin/site combinations that are in different channel groups, it ignores all but the first channel group.

![image](https://user-images.githubusercontent.com/17914399/235226066-bd685627-9a65-4dda-a34b-06d54bc20904.png)

## Repro or Code Sample

Edit the pin map to have multiple pins using multiple channel groups.

## Expected Behavior

Example does one of the following:
- Error when reserving more than one session
- Create a separate session for each channel group

## Current Behavior

Example creates a session for the first channel group and ignores pins that use subsequent channel groups.

[AB#2379261](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2379261)

PR #207, opened 2023-05-16
Moved session management code to helper VIs

Tharun-Sundar:

Move the session management related code to separate helper VIs.
- Create DCPower session.vi - Handles the reservation and creation of DC power sessions.
- Destroy DCPower session.vi - Handles the closure and detaching of DC power sessions.

These changes are intended as a starting point to come up with generic session management helper VIs that can be added to the measurement template.

Manually tested the measurement functionality.


Tharun-Sundar:
@rfriedma Please give your comments on the updated example. The next step is to remove the dependency of the helper VIs on the instrument specific VIs. Once that is accomplished, we will have generic helper VIs that can be added to the measurement template.

pbirkhol-ni:
> @rfriedma Please give your comments on the updated example. The next step is to remove the dependency of the helper VIs on the instrument specific VIs. Once that is accomplished, we will have generic helper VIs that can be added to the measurement template.

While I agree that this has the potential to make writing measurements easier, I also think it propagates (and then promotes in the template) some bad habits. @jasonmreding was doing some benchmarking and I think caching off the Discovery Client and Session Client made some significant performance improvements. I would rather move toward a model where the framework creates these clients and passes them to the `Measurement Logic.vi`, where the measurement author can optionally consume them. That way, we can hide a lot of the complexity and gain performance advantages too. Admittedly, I haven't prototyped this so I don't know the pitfalls or drawbacks.

I did prototype a proper Discovery Client and Session Management Client API a few months ago (see [here ](https://github.com/ni/measurementlink-labview/tree/users/pbirkhol/add-dc-power-example/Source/Session%20Management%20Client_lv/Session%20Manager)and [here](https://github.com/ni/measurementlink-labview/tree/users/pbirkhol/add-dc-power-example/Source/Discovery%20Client_lv/MeasurementLink%20Discovery%20Client%20API)). I think my API is better than a handful of helper VIs, but I also don't know how much time and effort we want to put into this feature.

Tharun-Sundar:
Measurement Logic.vi
image

Create DCPower session.vi
image

Destroy DCPower session.vi
image

DCPower Specific Logic.vi
image

jasonmreding:
For reference, when doing performance analysis, here is the original shipping example I started with:

image

and here is the final version that incorporated caching of clients:

image

In the case of DCPower, what I end up caching is the service address rather than the client (gRPC ID) since the LV driver API wrappers don't allow us to pass the equivalent of a gRPC channel into the create/attach session. I also adopted the convention of referring to the blue gRPC ID wire as the client (e.g. discovery client rather than discovery gRPC ID) since I found that was easier to read and generally made more sense.

The caching VIs (Create/Get/Delete) generally look like:

Get or Create Discovery Client.vi
image

Destroy Discovery Client.vi
image

Discovery Cache.vi
image
image
image

Basically, I cache the client in a functional global and only clear the cache if the measure call produced an error for any reason. This could of course be tuned to only clear the cache for certain types of gRPC errors. While this solution was sufficient for my benchmarking needs, it doesn't properly handle potential problems with clearing the cache while the current client is being used by two or more measure calls being executed in parallel. We would need a more sophisticated cache with reference counting to handle that properly. However, as Paul and I were discussing previously, this might be something we could punt on since I think the current event structure mechanism used to signal the start of the call is effectively serialized anyway which means we can only process one Measure RPC at a time anyway. To add to Ryan's comment earlier, if you had caching of the client, it would trivial to add Get/Create Session Manager Client inside Reserve/Unreserve Session if that cleans things up further.

With all that said, if the client feedback is that they prefer we provide proper APIs we own/maintain to do this sort of thing rather than template/example code the user owns, then I think we could create fairly small client libraries/wrappers for discovery service, session manager service, pin map service, and grpc-device service. The latter would contain as many driver specific wrappers as we think it makes sense to include. The client wrappers I'm envisioning here would either be wrappers around the grpc-labview generated client libraries (I know confusing) or we could merge in hand maintained APIs to that library and deal with the maintenance consequences when the proto file changes and we need to regenerate code. If we don't like the sprawl of all of these smallish client libraries, we could include them all in a MeasurementLink Client library wrapper that provides APIs for all of the various services we think will most commonly be used by the measurement service author. This would effectively be the equivalent of an API gateway but in library form rather than as a service. I forget what deployment size or performance tradeoffs there are to a mega library if any, but I think we have some options here.

jarnoldNI:
This is better, but I would like to see the VIs be instrument agnostic so they could be reused and ideally not 'owned' by the customer like Ryan mentioned earlier in this thread.

Some specific feedback.

- I thought is was strange that there is an input for instrument type id, but then in the block diagram it is hard-coded to find only the DCPower device. Either hard-code it to only work for DCPower or make it generic.

**LabVIEW style**
- Add context help documentation to all of the VIs.
- Use the connector pane to dictate which terminals are required, recommended, optional
- The capitalization of the controls and indicators does not seem to be consistent. My preference is to use all lower case unless it's an acronym.
image
image

jarnoldNI:
Another reusable VI that is very useful is something that takes in a pin and sessions and returns the session associated with that pin. Here's a similar proposal on the Python side of things:

https://github.com/ni/measurementlink-python/pull/263#issuecomment-1555198204

Use case is that you have multiple pin inputs each associated with a different instrument and you need to find the sessions assigned to each one....this is very common.

pbirkhol-ni:
> I thought is was strange that there is an input for instrument type id, but then in the block diagram it is hard-coded to find only the DCPower device. Either hard-code it to only work for DCPower or make it generic.

I think we can resolve a service just by the service class. So when resolving the grpc device server, I would leave the interface blank (since that is going to be device specific) and just use the service class instead, which is device agnostic.

pbirkhol-ni:
> > I think we'll get pushback on this design. Since the helpers strategy is to ship them with the examples they become pieces of code that are owned by the test architects. That would probably be fine to a degree (hardware specific initialization makes some sense to own), it doesn't really address the feedback that there are details of calling sessions that the developer didn't want to see. For instance, they didn't want to mess with discovery service at all and there it is sitting in the helper that they now own. If a subVI was sufficient then the test architects we're talking to would have selected the VIs, clicked "create SubVI" and moved on. Instead they gave us feedback that we should hide more details.
> > Is there a hybrid of this where we add to the Measurement Context (maybe?) that follows the pattern of the UI Updates? That was nice where we added just a single VI call to measurements. For sessions I would imagine that we could reduce it all the way down to "Reserve" and "Unreserve" session that takes in the pin or relay names, an instrument type id and a timeout. We're already partially there with the MeasurementContext object since it already has a Pin Map Context held there and this just builds on top of that information.
> > Also if we focus on just Reserve and Unreserve I feel like we'll be in a good position for the long term. Even in a scenario where sessions came pre-registered due to some sort of session automation, we'd likely still need the ability to signal that the session was in use. We can ignore Register and Unregister for now as that's not currently common in most measurements and ideally the registration that happens currently in TestStand is replaced by a custom step which would move the registration out of all of our LV examples.
> 
> @rfriedma I have now updated the VIs as per suggestions, Currently there are 4 VIs
> 
> 1. Reserve Sessions
>    image
> 2. Unreserve Sessions
>    image
> 3. Create DC Power Session
>    image
> 4. Destroy DC Power Session
>    image
> 
> Due to these changes the measurement logic now looks like, image Essentially doing the following operations,
> 
> * Reserving the sessions
> * Creating sessions
> * Performing measurement
> * Destroying sessions
> * Unregister sessions
> 
> The gRPC ids for the discovery and the service manager client has been cached in the measurement context.

pbirkhol-ni:
Consider removing the accessors for the ID properties. I see no reason to make those writable or even readable outside of the class.

pbirkhol-ni:
@jarnoldNI, are you suggesting that we should have methods on the MeasurementContext class to create and destroy instrument sessions? I think that is code that the measurement author should own, rather than pollute MeasurementContext with a bunch of instrument specific methods for creating sessions.

pbirkhol-ni:
Overall I think this is fine. It's not quite how I would have done it, but after talking to Ryan, we want to do something quick and easy for now. Since this is still an RFC, I'm not going to give it a thorough review. Once it graduates from being an RFC I'll review it further.

jarnoldNI:
> are you suggesting that we should have methods on the MeasurementContext class to create and destroy instrument sessions? I think that is code that the measurement author should own, rather than pollute MeasurementContext with a bunch of instrument specific methods for creating sessions

@pbirkhol-ni No, driver specific stuff should not be in there.

I'm not exactly which comment of mine caused the confusion, but when I mentioned the strange mix of generic and hard-coded, I was referring to this:
image
The only valid option for the _instrument id type_ is dcpower since dcpower is hard-coded as the service to find. I'd prefer this to be instrument agnostic.....if not, then just hard-code the _instrument type id_ to not cause confusion.

pbirkhol-ni:
> > are you suggesting that we should have methods on the MeasurementContext class to create and destroy instrument sessions? I think that is code that the measurement author should own, rather than pollute MeasurementContext with a bunch of instrument specific methods for creating sessions
> 
> @pbirkhol-ni No, driver specific stuff should not be in there.
> 
> I'm not exactly which comment of mine caused the confusion, but when I mentioned the strange mix of generic and hard-coded, I was referring to this: image The only valid option for the _instrument id type_ is dcpower since dcpower is hard-coded as the service to find. I'd prefer this to be instrument agnostic.....if not, then just hard-code the _instrument type id_ to not cause confusion.

Thanks for the clarification. I understand now. I had made a suggestion above as follows:

> I think we can resolve a service just by the service class. So when resolving the grpc device server, I would leave the interface blank (since that is going to be device specific) and just use the service class instead, which is device agnostic.

However, I looked at the code and you do have to provide an interface (this wasn't always true). In theory, though, you could provide _any_ interface that is supported by grpc device, even if it's not the one you want to use. So we could just leave it hardcoded at DCPower since we know that grpc device provides support for DCPower. That is a little dirty, but it would work. The other option is to have a lookup based on instrument type, but that will quickly go stale. We could also require the customer to specify the interface, but how would they know what value to use?

rfriedma:
It would be nice if the instrument_type_id wasn't actually different than the grpc-device interface string

jasonmreding:
The hard coded interface name for DCPower definitely jumps out as weird/wrong. 

image

After this discussion, it's clearer why it's there and why it's not a mistake. It definitely deserves a code comment to avoid confusion. This is probably a reasonable hack and should work unless we decide to break grpc-device up into smaller pieces that aren't deployed as a single monolithic server which contains all of the drivers. However, I think that's unlikely to happen in the near future if at all. However, it is a risk of making this assumption.

We could consider updating the Instrument Type Id control to typedef combo box.

image

This would allow for easy selection for the built in driver types, and users could still manually enter instrument ids for custom instruments and third party instruments. The down side is that LV doesn't support auto update of combo box controls from typedef. It's documented as [expected behavior](https://knowledge.ni.com/KnowledgeArticleDetails?id=kA00Z0000019LGqSAM&l=en-US) which is lame since this is one of the things I helped implement in NXG. If we really wanted to, we could probably make this an X control, and I think we could update the auto update semantics we want.

We could consider caching the session information from Reserve Session in the measurement context as well. This means you wouldn't need to wire it into Unreserve Session which would eliminate a somewhat awkward wire branch/bend. The down side is you wouldn't have as much freedom to mix/match multiple reserve calls together. However, I can't really think of a reason why somebody would want to do that. You might want to make break reserve into multiple calls for each instrument type you're dealing with. However, we could simply just append to the cached array with each reserve call and then unreserve everything in one call. I don't know if that asymmetry would be annoying/confusing or not. Perhaps it's not worth it to save a wire branch?

The Reserve/Unreserve Session VIs are still part of the example library and their icon template has slightly different shade from the other example VIs. Was the intent to make these part of the Measurement Context class? If not, then we have to leave the accessors for the gRPC IDs of the discovery service and session management service public.

We should also consider how the MeasurementContext Update Results.vi would be used with an example that performs incremental updates. My initial thought was we would want to pass the measurement context into the Core measurement VI. However, that becomes awkward when there are multiple sessions that you have to loop over since you have to stitch the results from each session together before writing the result back to the context. However, if that is where the looping for the incremental updates takes place, then maybe the loop to index the sessions would have to get pushed down as well. It might also create a bit of wire leap frogging in regards to what is traditionally wired to the ref in/out position on the con pane. I prefer straight railroad tracks whenever possible, but I'm not sure if that's achievable in this case.

jasonmreding:
> It would be nice if the instrument_type_id wasn't actually different than the grpc-device interface string

I thought the gRPC driver APIs actually provided constants for these interface strings. Assuming my recollection is correct, customers generally wouldn't have to hard code this string. They would just have to know that a constant exists and how to find it. We have to hard code the constant here because we don't want to take on a runtime dependency to one or all of the drivers if we want to make this generic. Changing those interface strings to match the instrument type ids used by the pin map probably isn't feasible now that we've released the drivers. I'm not sure it would be a good idea even if we could. We're kind of penned in by backward compatibility here.

bkeryan:
> I thought the gRPC driver APIs actually provided constants for these interface strings. 

@jasonmreding The Python APIs provide constants: 
- https://github.com/ni/nimi-python/blob/master/generated/nidcpower/nidcpower/grpc_session_options.py#L9
- https://github.com/ni/nidaqmx-python/blob/master/generated/nidaqmx/grpc_session_options.py#L11

I don't think the LV APIs do.

The server provides a JSON file listing all of the interfaces:
```
PS C:\Program Files\National Instruments\Shared\NI gRPC Device Server> type .\server_capabilities.json
{
    "provided_service_interfaces": [
        "nidaqmx_grpc.NiDAQmx",
        "nidcpower_grpc.NiDCPower",
        "nidigitalpattern_grpc.NiDigital",
        "nidmm_grpc.NiDmm",
        "nifgen_grpc.NiFgen",
        "nirfmxbluetooth_grpc.NiRFmxBluetooth",
        "nirfmxgsm_grpc.NiRFmxGSM",
        "nirfmxinstr_grpc.NiRFmxInstr",
        "nirfmxlte_grpc.NiRFmxLTE",
        "nirfmxnr_grpc.NiRFmxNR",
        "nirfmxspecan_grpc.NiRFmxSpecAn",
        "nirfmxwcdma_grpc.NiRFmxWCDMA",
        "nirfmxwlan_grpc.NiRFmxWLAN",
        "nirfsa_grpc.NiRFSA",
        "nirfsg_grpc.NiRFSG",
        "niscope_grpc.NiScope",
        "niswitch_grpc.NiSwitch",
        "nisync_grpc.NiSync",
        "nitclk_grpc.NiTClk",
        "nixnet_grpc.NiXnet",
        "nixnetsocket_grpc.NiXnetSocket"
    ]
}
```

subash-suresh:
I have updated the logic based on the comments given,
The changes are as follows,
- Moved the reservation and unreservation logic to separate VIs under the measurement context class. The gRPC clients are cached in the measurement context, which can be used to reserve and unreserve sessions. Since the measurement logic is part of the measurement server library, this code is completely managed by us. 

  Reserve sessions VI
  image
  
  Unreserve sessions VI
  image
  
  The updated measurement context
  image

- Added VIs in the nidcpower example to create the nidcpower session with the session information returned from the reserve session api call. Similarly an API has been added to destroy the created sessions.

  Create DC Power Session VI
  image
  
  Destroy DC Power Session VI
  image

- The measurement logic has been updated to work in the following order,
  reserve sessions -> create sessions -> measure -> destroy sessions -> unreserve sessions
  image

- A new helper VI has been added to the dc power helpers to allow users to filter sessions for a specific pin as per @jarnoldNI's request.
  image

subash-suresh:
> Does the logic still work if NIDCPower driver isn't installed on the system? Are the provided interfaces for gRPC-device server determined by driver installations, or no?

As long as the device server is installed the machine, it is fine give any supported interface name when trying to resolve the device server. So to answer your question, it does not depend on the driver installation..

subash-suresh:
I am closing this PR, as this prototype uncovered enough details to move forward with the actual implementation.

subash-suresh:
@dixonjoel , @pbirkhol-ni 
Could you please provide your reviews, I have removed the RFC tag.

dixonjoel (approved):

dixonjoel:
> @dixonjoel , @pbirkhol-ni Could you please provide your reviews, I have removed the RFC tag.

@subash-suresh FYI @pbirkhol-ni is out of the office until Monday.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
pbirkhol-ni, line 1: The Discovery Service and Session Manager gRPC IDs should use the `grpcID.ctl` typedef.

pbirkhol-ni (changes requested):
- Make sure all VIs are saved in the upper left of the left most screen (this includes both block diagram and front panel).
- These changes may fix https://github.com/ni/measurementlink-labview/issues/199. Can you confirm? Specifically, I want to make sure that it works as expected when using multiple pins with multiple channel groups. I also wonder if the measurements should run in parallel.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
pbirkhol-ni, line 1: Can you add a comment about why we are using `nidcpower_grpc.NiDCPower` as the provided interface?

subash-suresh:
Thanks for the comments, i will work on it
I would also like to hear your thoughts on this [comment](https://github.com/ni/measurementlink-labview/pull/207#:~:text=Note%3A%0AOne,submodule%20in%20LabVIEW).
@pbirkhol-ni , @dixonjoel.

pbirkhol-ni:
> Thanks for the comments, i will work on it I would also like to hear your thoughts on this [comment](https://github.com/ni/measurementlink-labview/pull/207#:~:text=Note%3A%0AOne,submodule%20in%20LabVIEW). @pbirkhol-ni , @dixonjoel.

I agree that having the reserve/unreserve methods on the MeausurementContext class isn't ideal. My preference (as I stated above) is to make proper APIs for the session manager, pin map client, and discovery client. That way we could reuse the APIs in the measurement and in the TestStand code modules. However, it was deemed too much work. Also, @rfriedma has said that he would like to abstract the session management into a custom TestStand step so that we don't have to write code modules to do the work. 

I think @jasonmreding has some strong opinions here as well so I will give him a chance to weigh in.

rfriedma:
we're trying to embrace "done enough" and allow further changes to come in the future through additional requests. Much of this design is about striking a balance between utility and sustaining activities. For instance, we can change out the implementation of "Reserve Session" in the future if we have a nicer wrapping API around session management (for instance in a class that handles a lot of the caching concerns the context class is currently handling). For better palettes and code around gRPC interfaces we might want to actually improve our generation scripts to create them to begin with. 

I think most of what you noted is good to add over time and not required all up front. In that vein, I think enough uncertainty exists on the sustaining on "what libraries and which palettes" that we can proceed without it for now.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
subash-suresh: fixed

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
subash-suresh: fixed

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
subash-suresh: updated

subash-suresh:
> * Make sure all VIs are saved in the upper left of the left most screen (this includes both block diagram and front panel).
> * These changes may fix [NIDCPowerSourceDCVoltage example silently ignores multiple channel groups #199](https://github.com/ni/measurementlink-labview/issues/199). Can you confirm? Specifically, I want to make sure that it works as expected when using multiple pins with multiple channel groups. I also wonder if the measurements should run in parallel.
1. I have moved the vis to the upper left corner
2. I tried to replicate the bug, but the latest updates seem to handle the issues for the most part. 
- Initially I tried with two pins in two channels which are on different channel groups. Hence, I expected two sessions to be created when I selected both the pins in the UI. This use case worked perfectly, and I was able to get the readings for both the sessions.
- I tried the same with one pin on multiple sites and each site mapped to a different channel group and here I always obtained only one session when reserving for pin. The expectation here is to get two sessions, right?

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
pbirkhol-ni: Yes, sorry, I should be been clearer. Jason mentioned in some feedback above that there should be a VI comment and I agree. If/When customers open that VI, it will be shocking for them to see that string if they aren't using DCPower. Having the comment in the VI will help ease their concerns.

subash-suresh: I have added a comment related to this in the code,
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
bkeryan, line 1: I have not reviewed this in detail, but it sounds like this PR has many of the same problems as https://github.com/ni/measurementlink-python/pull/276

In my opinion, putting a loop around the entire measurement and measuring each pin sequentially is not acceptable for any MeasurementLink example program, except maybe for multiplexer/switching/shared-pin use cases.

NI-DCPower independent channels sessions support accessing multiple instruments in one session and automatically parallelize calls using multiple threads when possible. The only thing they don't support is concurrent calls to the same session from multiple threads.

For single-site measurements, a single channel group is ideal. For multi-site measurements, one channel group per site is better because it enables concurrent calls to the same measurement without blocking.

Does the measurement need to support multiple sessions? If you use one channel group per site, then supporting multiple sessions is necessary to perform a measurement on "all sites". Otherwise, it is not required.

Some measurements should support multiple sessions. The Python examples for NI-FGEN and NI-SWITCH demonstrate how to do this. The session loop should be the innermost loop, so that multiple instruments perform measurements in parallel:

```
for session in sessions:
    open session
for session in sessions:
    configure session
for session in sessions:
    initiate session
for session in sessions:
    fetch results
for session in sessions:
    abort session
for session in sessions:
    close session
```

Cc: @subash-suresh

pbirkhol-ni: I validated that this PR does indeed fix #199, but as Brad pointed out, it serializes the measurements from different sessions. I think there are a few options:

1. In `Measurement Logic.vi`, check to see if the `session information` array is more than 1 element. If so, throw an error.
2. Make the for loop in `Measurement Logic.vi` a parallel for loop. I did this in my local branch and it seems to work. I don't think this is a fool proof solution (what if there are 100s of sessions?) but should be okay for most use cases.
3. Split up the logic as Brad suggested.

I think 3 is too complicated and obscures the purpose of the example. I would lean towards 1 or 2.

jasonmreding: I made a similar comment to this earlier about how this would work for a measurement that wants to provide continuous UI updates at some interval. The update presumably needs to include some data from all sessions. This makes it difficult to execute both loops in their entirety in parallel without some sort of rendezvous/synchronization of the two loops where the data can be joined for sending a response. Maybe this isn't common for DC Power so we ignore it. However, if we want to use the same general pattern in all of our examples, then we need to consider. In this scenario, I think you would probably need an outer loop that has the array of sessions that executes each session in parallel (maybe parallel for loop) and then sends an incremental update at the end of the outer loop once the inner loops have finished executing. Or maybe it would be cleaner to use a queue with each session worth of data and another loop running parallel to them which aggregates data across all sessions and sends incremental updates. This would be more in line with typical producer/consumer patterns in LV.

bkeryan: @jasonmreding Incremental updates are simpler if you structure your measurement VI like this:
```
for session in sessions:
    open session
for session in sessions:
    configure session
for session in sessions:
    initiate session
while not done:
    for session in sessions:
        fetch results
        update UI
for session in sessions:
    abort session
for session in sessions:
    close session
```

With this structure, no additional synchronization/queueing is required and the choice of multi-instrument session vs. for-loop vs. parallel for-loop does not affect the overall program flow.

jasonmreding: That's more or less what I was trying to convey. I was envisioning a for loop in place of your "while not done:", but basically the same thing. I think you could combine the first 3 for loops into a single loop and the last two for loops into a single loop if it simplifies things. However, that's probably a matter of style/preference than anything substantive.

subash-suresh: Thanks for all the comments.
I have updated the logic to tend to some of the comments mentioned in the discussed. 
The updated measurement logic looks like below,
image

There are still some downsides in this logic that needs to improve like the first for loop would fail if any one of the session creations fails and the same applies to the destroy as well. Given this downside, I have tested the example and it seems to function properly.

Note: I have set the default parallel instances to a maximum of 8 as per the LabVIEW default setting.

subash-suresh: @dixonjoel , @pbirkhol-ni , @bkeryan 
Should i update the logic in this PR to just use the first session information from the reserve session call like the logic in #221 ? . Please let me know if that has to be done or if the existing parallel loop logic is fine to have in the dc power example.
One thing to note here is that in python the dc power example just takes the first session information to create a session similar to #221

subash-suresh: Updated the code to use logic similar to #221 , to take the first session information to create the sessions after reservation.
The updated measurement logic looks as follows,
image


I have modified the `get_session_information` vi to accept an array of session information input instead of reserve sessions response object like used in #221 .

sreekantham (changes requested):
I don't approve of adding this to vi.lib with out further discussion as we talked about in the water cooler on June 12th.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
pbirkhol-ni: I think that is reasonable, but based on a long water cooler conversation, Kiran is adamant that we not ship these changes as part of vi.lib. So I think the poly VI solution is less applicable if these are example-only changes.

pbirkhol-ni:
> I don't approve of adding this to vi.lib with out further discussion as we talked about in the water cooler on June 12th.

I discussed with Jack. He is okay with making the two new VIs (Reserve Sessions and Unreserve Sessions) part of the example and not part of the MeasurementContext. The VIs should remain instrument agnostic as much as possible so that they can easily be copied to another measurement.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
bkeryan: ok, so copy "Reserve Session.vi" (singular) to all of the single-session examples

pbirkhol-ni:
> > I don't approve of adding this to vi.lib with out further discussion as we talked about in the water cooler on June 12th.
> 
> I discussed with Jack. He is okay with making the two new VIs (Reserve Sessions and Unreserve Sessions) part of the example and not part of the MeasurementContext. The VIs should remain instrument agnostic as much as possible so that they can easily be copied to another measurement.

Let's actually hold off on taking any more actions. We're discussing this more offline. I'll post back when we come to a conclusion.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
bkeryan: Here is the same idea, implemented in Python: https://github.com/ni/measurementlink-python/pull/287

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Reserve Sessions.vi
bkeryan, line 1: Not all measurement services want grpc-device.

subash-suresh:
@brad, @pbirkhol-ni , @dixonjoel 
Based on the discussions I have created #222, Which try's to maintain the changes with in the example itself.

dixonjoel:
@subash-suresh @Tharun-Sundar does this PR replace this one? Should we just close this PR?
https://github.com/ni/measurementlink-labview/pull/222/files

subash-suresh:
I am closing this PR as there is another PR #222 , which has the implementation discussed in this PR.

PR #209, opened 2023-05-16
Make the Discovery Service helper VIs public

pbirkhol-ni:

There are a handful of Discovery Service helper VIs that we use in examples and that customers probably use as well. I unintentionally made those VIs community scope in a prior PR and now they can't be used outside of the containing library. This PR makes those VIs public again.

The DCPower shipping example is broken without this change.

I opened and ran the DCPower shipping example.

### DIFF
--- a/Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
+++ b/Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
@@ -14,9 +14,11 @@
 		<Item Name="Error Strings Global.vi" Type="VI" URL="../Helpers/Error Strings Global.vi"/>
 		<Item Name="Get Configuration Metadata.vi" Type="VI" URL="../Helpers/Get Configuration Metadata.vi"/>
 		<Item Name="Get Default for Type.vi" Type="VI" URL="../Helpers/Get Default for Type.vi"/>
+		<Item Name="Get Discovery Service Key File Path.vi" Type="VI" URL="../Helpers/Get Discovery Service Key File Path.vi"/>
 		<Item Name="Get MetaData from Configurations.vi" Type="VI" URL="../Helpers/Get MetaData from Configurations.vi"/>
 		<Item Name="Get MetaData from Results.vi" Type="VI" URL="../Helpers/Get MetaData from Results.vi"/>
 		<Item Name="LabVIEW datatype to Measurement datatype.vi" Type="VI" URL="../Helpers/LabVIEW datatype to Measurement datatype.vi"/>
+		<Item Name="Read Discovery Service Key File.vi" Type="VI" URL="../Helpers/Read Discovery Service Key File.vi"/>
 		<Item Name="Register Configuration and Results.vi" Type="VI" URL="../Helpers/Register Configuration and Results.vi"/>
 		<Item Name="Register to Discovery Service.vi" Type="VI" URL="../Helpers/Register to Discovery Service.vi"/>
 		<Item Name="Validate Configuration Metadata.vi" Type="VI" URL="../Helpers/Validate Configuration Metadata.vi"/>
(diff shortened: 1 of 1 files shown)


Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
```diff
@@ -37,6 +34,12 @@
 		<Item Name="PinMapContext.ctl" Type="VI" URL="../Helpers/TypeDefs/PinMapContext.ctl"/>
 		<Item Name="UserInterfaceDetails.ctl" Type="VI" URL="../Helpers/TypeDefs/UserInterfaceDetails.ctl"/>
 	</Item>
+	<Item Name="Public Helpers" Type="Folder">
+		<Property Name="NI.LibItem.Scope" Type="Int">1</Property>
+		<Item Name="Get Discovery Service Address.vi" Type="VI" URL="../Helpers/Get Discovery Service Address.vi"/>
```
jasonmreding, line 39: I think Get Discovery Service Address.vi is the only VI that needs to be public. The other two are implementation details of this VI. Ideally, I think this VI would be packaged as part of the discovery client library. However, I know that breaks backward compatibility. If we did that, we would probably also want to come up with a convention for how to store hand written code in the library with the generated code from grpc-labview. If we had the equivalent client API in LV to automatically start the discovery service, then I don't think we would need this public API at all, although it probably doesn't hurt.

jasonmreding (approved):

Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
pbirkhol-ni: My prototype Discovery Client API (which wrapped the Discovery Client output of grpc-labview) did just this. However, I don't think we're going to productize it any time soon. 

I don't really like that the `MeasurementLink Measurement Server.lvlib` is part of our public API, but there isn't much we can do now. Making most of the helper VIs community scoped will help prevent further usage. Hopefully none of the VIs that we made community scoped aren't being used by customers. If so, though, it's easy enough to release new LabVIEW support.

I will go ahead and only make `Get Discovery Service Address.vi` public.

jasonmreding: Yeah, I think the best we could do is introduce a duplicate VI into the discovery client and update all of our templates and library code to reference this VI rather than the one in `MeasurementLink Measurement Server.lvlib`. We could move the VI into a deprecated folder to make it clear the VI is only being maintained for compatibility purposes. Then after some period of time after we no longer want to support v1 measurements, we could delete the copy in `MeasurementLink Measurement Server.lvlib`. At least this way, all v2 measurements will start out "clean".

pbirkhol-ni: This isn't a v1/v2 issue. When customers grab the latest measurementlink-labview release, they will automatically be promoted to v2 if they are using a v2 client. However, they will still be using `Get Discovery Service Address.vi`. They could even create a new measurement and use `Get Discovery Service Address.vi` if they copy the logic from an existing measurement. 

I do agree with deprecating `Get Discovery Service Address.vi`, but maybe we should first figure out what we're doing as part of [this ](https://github.com/ni/measurementlink-labview/pull/207)RFC PR. If we decide to wrap the grpc-labview Discovery Client then we could also move `Get Discovery Service Address.vi` to the client wrapper. That would also have the benefit of not modifying the output of grpc-labview.

For now I'll create a github issue for this so we don't forget it.

pbirkhol-ni: https://github.com/ni/measurementlink-labview/issues/210

aepete (approved):

Issue #211, opened 2023-05-17
Use Fuse controls (also called NXG controls) for LabVIEW measurement UIs


pbirkhol-ni:
# Bug Report

The LabVIEW PO asked us to use Fuse controls (also called NXG controls) for our LabVIEW UIs. We should update all examples LabVIEW UIs to use fuse controls. We should also make the background color white (RGB 255 255 255).

[AB#2395694](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2395694)

PR #212, opened 2023-05-17
Use NXG/fuse controls for Measurement UIs

pbirkhol-ni:

This PR uses NXG/fuse controls for all Measurement UIs. It modifies the 3 example UIs and the template Measurement UI. The following were done to the VIs:
- The background color was changed to white.
- All controls/indicators were changed to NXG controls.
- The Editor Options were changed to NXG style controls.

This PR also re-saves the .sfp with the new controls (I'm not sure this is necessary. I also saved it in a pre-release version of InstrumentStudio, which may or may not be okay). 

#211 

The LabVIEW PO requested that we use these controls.

I ran all three examples and made sure the bindings still work. I also created a new measurement and verified that the template has the new controls/indicators.

image
image
image
image


aepete (approved):
[Suggestion for consideration] I'm not sure if customers would get hung up on axis labels of `Amplitude` or `Time`...  but being that the associated default logic implementation is just a passthrough that doesn't do any timing-based operations, labeling the graph with `Value` and `Index` might be a better fit out of the box.

jasonmreding:
> This PR also re-saves the .sfp with the new controls (I'm not sure this is necessary. I also saved it in a pre-release version of InstrumentStudio, which may or may not be okay).

This should be harmless but unnecessary. There is no UI information about the service that currently gets persisted in the `ScreenModel`. Saving in an unreleased version or a newer version in general can be dicey since it might prevent the UI from being loadable in older versions of the application. In general, if none of the `Version` and `OldestCompatibleVersion` properties persisted in the elements of the `SourceModelFeatureSet` have changed, then no persistence versioning has changed and everything should have the same compatibility as before. All of the other properties are informational only and designed to help traceability when there are load failures. I'm not sure why the `ClientId` property changed unless you removed the measurement from the sfp and then added it back using the Edit Layout Dialog.

jasonmreding:
Were you going to update the template UI VI used by the generator as well? Or were you planning to do that in a separate PR?

jasonmreding (approved):

pbirkhol-ni:
> Were you going to update the template UI VI used by the generator as well? Or were you planning to do that in a separate PR?

Yes, it is in this PR already. There is a screenshot of it in the PR description.

pbirkhol-ni:
> > This PR also re-saves the .sfp with the new controls (I'm not sure this is necessary. I also saved it in a pre-release version of InstrumentStudio, which may or may not be okay).
> 
> This should be harmless but unnecessary. There is no UI information about the service that currently gets persisted in the `ScreenModel`. Saving in an unreleased version or a newer version in general can be dicey since it might prevent the UI from being loadable in older versions of the application. In general, if none of the `Version` and `OldestCompatibleVersion` properties persisted in the elements of the `SourceModelFeatureSet` have changed, then no persistence versioning has changed and everything should have the same compatibility as before. All of the other properties are informational only and designed to help traceability when there are load failures. I'm not sure why the `ClientId` property changed unless you removed the measurement from the sfp and then added it back using the Edit Layout Dialog.

If it's unnecessary then I'll revert it. For some reason I thought the UI was saved in the sfp. Maybe it is for Measurement UI Editor UIs?

Source/Example Measurements/UIProgressUpdates/UIProgressUpdates/Measurement UI.vi
bkeryan, line 1: The origin is in the wrong place.

image

Source/Generator/MeasurementLink Measurement Template/Measurement UI.vi
bkeryan, line 1: The origin is a little bit off on this one, but not as much as in UIProgressUpdates.

bkeryan (approved):
Approved with suggestions

jasonmreding:
> If it's unnecessary then I'll revert it. For some reason I thought the UI was saved in the sfp. Maybe it is for Measurement UI Editor UIs?

Yes, for .measui UI, a copy of the contents are saved in the .sfp file. While a copy is maintained, resaving isn't strictly necessary because we always update the copy of the contents each time we initialize the measurement. This ensures the sfp always reflects the current contents of the .measui if the service was modified and redeployed after the sfp was saved. If the copy is out of date, there might be some UI glitch/flicker depending on how long the measurement takes to initialize. In essence, we'll display the last known contents for the UI until we can initialize and get the new contents for the UI. If we fail to initialize the measurement, then we'll always just display the last known contents of the UI with an error message saying we couldn't initialize the measurement. For a .vi UI, we just display a blank canvas until the measurement is initialized and we can run/render the VI. If the VI fails to load, then we display a pretty rough looking error text box.

Source/Example Measurements/UIProgressUpdates/UIProgressUpdates/Measurement UI.vi
pbirkhol-ni: I never pay much attention to the origin. Luckily LabVIEW has a button to position the origin in the right place for you.  :)

PR #215, opened 2023-05-25
[DRAFT]

WesleyTangNationalInstruments:
- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-labview/blob/main/CONTRIBUTING.md).

TODO: Include high-level description of the changes in this pull request.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Results Metadata.vi
pbirkhol-ni, line 1: - There are still quite a few reference that don't get closed: The Panel reference and both arrays of Controls references.
- This should be split up into logical subVIs. Those subVIs can probably be used in both this VI and in `Get Configuration Metadata.vi.`
- There is no need to append `output` to the Parameter name (and doing so will cause issues). Because of where we are calling this VI, we already know they are outputs and they are going to get bundled into the `outputs` element of the `measurement_signature` cluster.
- The logic for handling enum arrays is incorrect. We want the refnum of the array, not a refnum of an array element. So the logic should be something like: `if enum or if enum array then cache control refnum`. The cached control will always be either an enum control refnum or an enum array control refnum. This means that you might have to modify your `Create Enum Type Specialization.vi` to work with both an enum refnum and an enum array refnum. 
- You can add a conditional terminal to a for loop by right clicking on it and selecting `Conditional Terminal`. I think that would make your logic simpler when determining if the enum specialization already exists (you can stop the for loop early if/when you find a match). 
- I don't think using `Insert Into Array` is the correct thing to do when adding an enum type specialization. You want `Build Array` instead. Also, anything you insert into the array is going to get lost since you don't save it in a shift register. But saving it in a shift register is weird because the inner for loop iterates over the type specialization array and you wouldn't want it to iterate over new elements (although it probably wouldn't hurt). So I guess you could just use a shift register. You could also make a copy of the array outside of the outer for loop so that one copy stays the same and the other copy is the shift register copy

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Configuration Metadata.vi
pbirkhol-ni, line 1: This VI has similar feedback to `Get Results Metadata.vi`. In addition:
- This could _probably_ be merged with `Get Results Metadata.vi`, but I think I'd rather keep them separate and then refactor the internal logic into subVIs that get reused. That way we're not duplicating a lot of code but we also have more flexibility for the two implementations to diverge in the future (if needed). I don't feel strongly about this, though. My concern is that it will be more difficult to understand the logic if we try to have one VI called `Get Metadata` that handles both inputs and outputs.

pbirkhol-ni (dismissed):

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementPluginService/Get Type Specializations.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Yes, this VI refnum was linking to the old version which did not contain the added enum values for Type Specialization and Type Specialization Key.

image

WesleyTangNationalInstruments:
image

Filter for Enum Controls.vi
image

Get Measurement Configuration Controls.vi
image

Get Configuration MetaData.vi
image

Get Results MetaData.vi
image

Get MetaData from Configurations.vi
image

Get MetaData from Results.vi
image

Get Type Specialization.vi
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Results Metadata.vi
WesleyTangNationalInstruments: I addressed these comments as suggested in my latest update.

WesleyTangNationalInstruments: > This should be split up into logical subVIs. 

I made subVIs for the getting of the controls and the filtering to enum controls, but I left the logic where we translate Type Specializations into Metadata as that seems to be the core function of "getting metadata". If you would prefer I move that logic into a subVI as well, let me know.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Configuration Metadata.vi
WesleyTangNationalInstruments: I left these two VIs (Get Configuration MetaData.vi and Get Results Metadata.vi) as separate VIs as suggested, though they are 100% logically identical after suggested refactors.

jasonmreding:
Create Enum Type Specialization:

- We are always adding enum.values annotation even if a control isn't an enum. If it isn't an enum, the list of enum values is empty so having the annotation at all is confusing. I would prefer not to generate any enum annotations if the control isn't an enum.
- Actually.... it looks like you are only calling this VI after you have already filtered the list of controls to enums. In that case, you should update the control ref in/out to an enum control reference and delete all of the code that validates the control is an enum.
- Rather than require the caller to pass in the name of the parameter by extracting the text for the label of the control, just move it into the implementation of the VI.
- You shouldn't need the upcasts to generic control class. If you want to leave them to get rid of the coercion dot (type widening) due to the implicit upcast by LV for the wire tunnel, I won't complain. However, I don't generally bother with explicit upcasts just to get rid of the coercion dot. This comment likely applies throughout.
- Consider creating IsEnumControl, IsArrayControl, or IsEnumOrEnumArrayControl helper VIs for readability.
- Make this helper private like the rest of them.

Get Results Metadata:

- This is using the control for the configuration rather than the results. If we haven't already, we should test enum indicators work as expected.
- Consolidate logic between this VI and Get Configuration Metadata. Either pass in the control reference for the metadata or create a shared sub VI for the common parts.

WesleyTangNationalInstruments:
> Actually.... it looks like you are only calling this VI after you have already filtered the list of controls to enums. In that case, you should update the control ref in/out to an enum control reference and delete all of the code that validates the control is an enum.

I don't think we can use an enum control reference here as we pass in both controls references of both enum and arrays of enums as valid input.

> This is using the control for the configuration rather than the results. 

Are you referring to `ConfigurationParameter_AnnotationsEntry.ctl` being used in both `Get Configuration Metadata.vi` and `Get Results Metadata.vi`? I think this is really just a naming problem as the AnnotationsEntry gets used by both `ConfigurationParameter.ctl` and `Output.ctl` now and should be generically named. I don't believe we are incorrectly using `ConfigurationParameter.ctl` as that is handled at the `Get MetaData from Results.vi` level rather than here in the `Get Results Metadata.vi` which is agnostic of that type.

> If we haven't already, we should test enum indicators work as expected.

I don't have a fully working example that behaviors correctly in InstrumentStudio yet since there's an issue with serialization/deserialization (which I believe needs to be resolved outside of this LabView code), but I have tested a service with a debugger to verify that an example with an enum will run through this code and produce annotations. 

Screenshots below:

image

Filter for Enum Controls.vi
image

Get Configuration Metadata.vi
image

Get Enum Type Specialization.vi
image

Get Measurement Configuration Controls.vi
image

Get MetaData from Configurations.vi
image

Get MetaData from Results.vi
image

Get Metadata.vi
image

Get Results Metadata.vi
image

IsEnumArrayControl.vi
image

IsEnumControl.vi
image

IsEnumOrEnumArrayControl.vi
image

jasonmreding:
> I don't think we can use an enum control reference here as we pass in both controls references of both enum and arrays of enums as valid input.

The array control also has a control reference for the array element, which we get when generating the enum specialization. What I was suggesting is that the Filter for Enum Controls.vi extract this control reference for array controls (casted as enum control reference) and the original control reference for enum controls (casted as enum reference). I think if the various helpers for Is Enum, Is Enum Array, and Is Enum or Enum Array had an output for the casted reference type this would be a little easier and clean up other parts of the diagram throughout. The control ref out could be changed to the more specific type, but I think this might be more confusing than just having a second ref out for the more specific type of control reference. I still like having the Boolean output since it saves the caller from having to check/call Not A Refnum.

> Are you referring to ConfigurationParameter_AnnotationsEntry.ctl being used in both Get Configuration Metadata.vi and Get Results Metadata.vi? I think this is really just a naming problem as the AnnotationsEntry gets used by both ConfigurationParameter.ctl and Output.ctl now and should be generically named. I don't believe we are incorrectly using ConfigurationParameter.ctl as that is handled at the Get MetaData from Results.vi level rather than here in the Get Results Metadata.vi which is agnostic of that type.

I'm referring to this which is called from both Get Configuration Metadata and Get Results Metadata. If I'm reading this correctly, that means we're going to generate enum annotations based on the configuration controls in both cases.

image

- Get MetaData from Results and Get MetaData from Configuration appear to be entirely duplicated minus the sub VI call to get the metadata that is specific to either the configuration or results. I would recommend refactoring so that either the annotations are passed into both or tunnel through the VI refnum for the appropriate ctl so the sub VI is no longer specific to configuration or results.
- IsEnumArrayControl should use IsEnumControl in its implementation.

WesleyTangNationalInstruments:
> Get MetaData from Results and Get MetaData from Configuration appear to be entirely duplicated minus the sub VI call to get the metadata that is specific to either the configuration or results. I would recommend refactoring so that either the annotations are passed into both or tunnel through the VI refnum for the appropriate ctl so the sub VI is no longer specific to configuration or results.

@jasonmreding I addressed this by refactoring the `Get MetaData from Results.vi` and `Get MetaData from Configuration.vi` so that we call into the non-specific `Get Metadata.vi` and allows us to remove both the specific `Get Configuration Metadata.vi` and `Get Results Metadata.vi` as well as creating a `Get Annotations For Paramater.vi`. 

I addressed the other comments as suggested.

Also note the additional change to `LabVIEW datatype to Measurement datatype.vi` so that instead of coercing LabVIEW enums to UInts, we use the gRPC TYPE_ENUM. This was something Everett and I ran into while trying to get a fully functional example with these changes. 

IsEnumArrayControl.vi
image

IsEnumControl.vi
image

Filter for Enum Controls.vi
image

image

Get Enum Type Specialization.vi
image

Get MetaData from Configuration.vi
image

Get MetaData from Results.vi
image

Get Metadata.vi
image

Get Annotations For Parameter.vi
image

LabVIEW datatype to Measurement datatype.vi
image

dixonjoel:
Are there any examples that are already specifying enum type specializations that we should remove them from?

dixonjoel (approved):

jasonmreding:
IsEnumArrayControl:
- Consider passing out the enum control reference for the array element instead of the array reference itself. This will simplify the caller a bit.

Get Enum Type Specializations:
- Minor: There are a couple of unbundle by name primitives where the Annotations property isn't wired which is kind of confusing. It would be better to just remove that element from the unbundle.

Get Annotations for Parameter:
- Notice the coercion dot on the indicator. That is because we basically have two typedefs for the same thing. One of them comes from ConfigurationParameter_AnnotationsEntry.ctl and the other comes from the grpc-labview generated typedef defined by the v2 measurement service interface. Paul tried to remove any direct dependencies to a version specific interface at this level so here's what I think we should do:
  - Rename ConfigurationParameter_AnnotationsEntry.ctl to Parameter_AnnotationsEntry.ctl.
  - Update Output.ctl to use this typedef rather than the one from the v2 measurement interface.
  - While you're at it, rename Output.ctl to OutputParameter.ctl for consistency.

jasonmreding (approved):

WesleyTangNationalInstruments:
> IsEnumArrayControl: Consider passing out the enum control reference for the array element instead of the array reference itself. This will simplify the caller a bit.

@jasonmreding I've opted not to make any changes for this comment. It seems weird to have `IsEnumControl.vi` return two references to the same control and change `IsEnumArrayControl.vi` to return two references to two different controls. Additionally, we are cleaning up the reference to the array element in `IsEnumArrayControl.vi` and if instead we passed it out as an output, we would move that responsibility into the caller to have to close both the enum and the array reference. From a usage perspective, it seems more intuitive to have them behave the same.

Get Enum Type Specialization.vi
image

Parameter_AnnotationsEntry.ctl
image

OutputParameter.ctl
image

ConfigurationParameter.ctl
image

`GetMetadataResponse.ctl`, `MeasurementSignature.ctl`, `Get Metadata.vi`, `Get MetaData from Results.vi`, and `Get Annotations for Parameter.vi` have no visible changes, but they were just updated to reference the renamed `OutputParameter.ctl` and `Parameter_AnnotationsEntry.ctl`.

PR #219, opened 2023-06-07
[BUG] Fixed an issue with generating enum annotations for arrays of enums.

WesleyTangNationalInstruments:

Fixed an issue with enum arrays where we use the name of the enum inside the array rather than the array itself. This is problematic because we do lookup by parameter name and we fail to find and thus fail to generate enum annotations for enum arrays. I think this got inadvertently broken when we removed the `Parameter Name` input to `Create Enum Type Specialization.vi`.

Arrays of enums do not generate correct annotations otherwise.

Tested by validating the annotations are correct for enum arrays. Note that arrays of enums in `.measui` UIs are [not supported yet](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2372263) and LabVIEW UIs are [currently broken](https://dev.azure.com/ni/DevCentral/_workitems/edit/2411021) so validation was done by examining the annotations.

image


WesleyTangNationalInstruments:
image

Get Enum Type Specialization.vi
image

Filter Enum Controls.vi
image

dixonjoel (approved):

PR #220, opened 2023-06-08
Add a user event that can be used to programmatically stop the server

pbirkhol-ni:

For our internal testing, we have a need to be able to programmatically start and stop a LabVIEW measurement. While starting a VI is easy, it is currently not possible to stop the VI without aborting it, which causes other issues (like not unregistering with the DiscoveryService). This change could also be useful for customer testing as well.

This change does the following:
- Create a new optional input to the `RunService.vi` that accepts a cluster of user events. The only user event at the moment is a stop user event that is wired into the Stop event structure and triggers the measurement to stop when fired. The reason for using a cluster is to allow us to add new user events in the future.
- Modifies the template to add the new user event to the customer's `Run Service.vi`. It is hidden on the front panel since it is not meant for interaction. 

I used this new code from TestStand. I wrote two code modules: one to start the VI and one to stop. TestStand cached the VI refnum and the user event refnum.

Start VI:
image

Stop VI:
image


pbirkhol-ni:
New template VI:
image

pbirkhol-ni:
Run Service.vi. The only change is the addition of the user event as an input and then wiring it to the correct VI:
image

pbirkhol-ni:
The service stop handler VIs. The changes to V1 and V2 are identical. I just added a new input for the refnum and wired it to a subVI.
image

pbirkhol-ni:
Stop Handler Impl.vi. This is where most of the changes occurred. I refactored it quite a bit.

image

pbirkhol-ni:
FYI @jasonmreding , this is what we had discussed yesterday.

jasonmreding (approved):
You might consider creating a typedef for the user event refnum. That way you can add new data to the event in the future without breaking anything. That might make creation of the event more awkward (not sure), but you could always generate a VI that does this as part of the generator.

You might also consider whether you might ever want to add other events for communication in the future. In that case, you might want to convert this to a cluster of user events that you create a typedef out of. That's basically what we did with the GPlugin SDK with the Plugin Events.ctl. Again, a helper VI can help with the ceremony of creating/populating the type. This approach provides room for extending the API without breaking it. The likelihood of you ever needing this is another question.

pbirkhol-ni:
> You might consider creating a typedef for the user event refnum. That way you can add new data to the event in the future without breaking anything. That might make creation of the event more awkward (not sure), but you could always generate a VI that does this as part of the generator.
> 
> You might also consider whether you might ever want to add other events for communication in the future. In that case, you might want to convert this to a cluster of user events that you create a typedef out of. That's basically what we did with the GPlugin SDK with the Plugin Events.ctl. Again, a helper VI can help with the ceremony of creating/populating the type. This approach provides room for extending the API without breaking it. The likelihood of you ever needing this is another question.

I added a cluster with a single user event. Both the cluster and the user event are type defed.

PR #221, opened 2023-06-09
examples: Add HP/Agilent/Keysight 34401A DMM example

bkeryan:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-labview/blob/main/CONTRIBUTING.md).

### What does this Pull Request accomplish?

Add an example that demonstrates how to write a MeasurementLink plug-in using instrument drivers and NI-VISA.

### Why should this Pull Request be merged?

Demonstrate this use case.

Facilitate performance testing/investigation.

### What testing has been done?

Manually tested in InstrumentStudio and TestStand with a 34401A connected via GPIB.

PR #222, opened 2023-06-14
Create helper VIs for session management APIs in DC Power Example

subash-suresh:

Currently the reserve session call returns a list of session info and users have to define their own method to get the exact session's information to use the required session on the pins that is needed to be used. A new helper method has been introduced in this PR to allow the users to get the of session information (single/ multiple) related to a specific pin or a set of pins. Additionally, the logic to reserve and unreserve the session has some repeated logic that could be provided as helpers in the examples and this PR does that as well.

This PR is based on the discussions from #207 

This PR does the following,
1. Added a `Get Session Information.vi` from #221.
2. Added a cluster `gRPC Channel Pool` to maintain the discovery service and session management service gRPC ids.
3. Added polymorphic vi implementation of reserve session(s) which in one version returns a single session information and, in another version, it returns multiple sessions information. This uses `Get Session Information.vi` internally. Similarly, a polymorphic vi implementation is also used for unreserve session(s).
4. Added `Initialize DC Power Session.vi` and `Close DC Power Session` helper vis which are used in 'DC Power Specific Logic.vi` which is the main logic used in the measurement logic.

Validated the dc power example by manually running the example and verifying the results.


subash-suresh:
The updates of the different files can be found below,
- `GRPC Channel Pool.ctl`
image
-  `Get Session Information.vi` from #221 
image
-  `Reserve Session(s)` Polymorphic VI - The logic is similar to #207 , only difference being that grpc ids are taken using a local control instead of measurement context.
  - `Reserve Session- the single session version.
  
image

  - `Reserve Session- the multiple session version.
  
image

-  `Unreserve Session(s)` Polymorphic VI -  The logic is similar to #207, only difference being that grpc ids are taken using a local control instead of measurement context.  
  - `Unreserve Session- the single session version.
  
image

  - `Unreserve Session- the multiple session version.
  
image

- `Create DC Power Session.vi` similar to #207 
image

- `Destroy DC Power Session.vi` similar to #207
image

- `DC Power Specific Logic` similar to #207
image

Finally with all the above changes the measurement logic looks like below,
image

subash-suresh:
Is it better to move the gRPC IDs in the measurement context?. Or is the existing implementation in this PR look good?

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
dixonjoel, line 1: I'm open to discussion on this, but I like showing the instance icon and polymorphic selector on the Reserve Session(s) poly VI to clarify what's happening and distinguishing the icon with, e.g. "Reserve Single Session" and "Reserve Mult Sessions". On the unreserve session, I think we should have a single icon and not show the selector since it will just auto-adapt. Something like this.
image

dixonjoel (approved):
Approved pending discussion on poly VI icons.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
pbirkhol-ni, line 1: I think this should show the poly selector because it won't auto adapt. It is probably okay to not show the selector for Unreserve Session(s).vi because it will auto adapt.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve Session.vi
pbirkhol-ni, line 1: Use Build Array rather than Insert Into Array.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session.vi
pbirkhol-ni, line 1: This should use standard error handling. The logic should be wrapped in a case structure and only execute if there is no error. This is actually true for all subVIs that shouldn't execute on incoming error.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Sessions.vi
pbirkhol-ni, line 1: This should use standard error handling. The logic should be wrapped in a case structure and only execute if there is no error.

pbirkhol-ni (changes requested):

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Get Session Information.vi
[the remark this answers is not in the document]
subash-suresh: The icon has been updated to mention dcpower

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Sessions.vi
subash-suresh: Updated the reserve session and create dcpower VIs to use error case structures as they need to be skipped when there is an error.
- Reserver Session
image
image

- Create DC Power Session
image

- DC Power Specific Logic
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session.vi
subash-suresh: Updated as suggested

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
subash-suresh: I have updated the reserve and unreserve API icons and the poly selector in the measurement logic as shown below,

subash-suresh:
Latest updates on 6/15/23,

@dixonjoel, @pbirkhol-ni , @bkeryan 
I have moved the logic of getting the gRPC device location out of the reserve session api into a seperate helper vi. This is because the reserve sessions API can be used for a device driver that is not part of the grpc device server based on the feedback from Brad in #207 .

This new vi is called `Get gRPC device location.vi`,
This VI receives the gRPC channel pool as input, and if the discovery service id is 0 then it trys to create a discovery service client.
image
image

The `Create DC Power Sessions.vi` uses the `Get gRPC device location.vi` to get the device server location. Hence the VI has been updated to accept gRPC channel pool.
image

Based on these updates the measurement logic looks like below,
image

I am resetting reviewers due to this new update

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Sessions.vi
pbirkhol-ni, line 1: In the error case, the MeasurementContext class should be wired through instead of creating a new MeausurementContext object. As a general rule, any input that has a corresponding output should be wired through the error case. This usually applies to error wires, classes, refnums, etc.

Also, this isn't strictly necessary, but for things like the `session information` array and the `gRPC channel pool out` cluster, you can just right click on the tunnel and select `Use Default if Unwired`. Then you don't have to create a constant in the Error case just to pass out default data. I think this can be clearer to someone reading the code since the tunnel changes appearance and you don't have to look at the Error case to know what is happening.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Get Session Information.vi
pbirkhol-ni, line 1: The `session information` indicator isn't wired.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Sessions.vi
pbirkhol-ni: This same comment applies to other VIs where you added the error case structure.

pbirkhol-ni: Example:
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
pbirkhol-ni, line 1: The Instrument Handle isn't wired in this VI.

Make sure you are testing the example before you submit. Both success and error cases. As is, it couldn't possibly work.

pbirkhol-ni (changes requested):
Make sure you use required inputs where necessary. As a general rule, _all_ inputs should be required unless we can come up with a sane default. So timeouts can be recommended (or even optional), but sessions, class, etc should be required.

jasonmreding:
I have mixed feelings on the channel pool. So it may just be a terminology/naming issue, but I don't really consider this a channel pool as it only exists for the lifetime of a single RPC on the server. Ideally, a channel pool would be utilized so that channels/clients can be reused across multiple RPCs. However, this brings in the complication of knowing when to remove invalid/stale channels from the pool because the service is no longer running on that port. 

If we're trying to avoid taking on the complexity/implementation cost of invalidating the cache for a channel pool, then maybe just tweaking the name is the way to go. However, I'm not really coming up with any good suggestions at the moment. Maybe just calling it something like GRPC Clients.ctl would be better.

If you want to take on creating a proper channel pool (which I think would be valuable), then I think wrapping it in a class or some other API (functional globals) would be better. This would allow you to remove some duplicated for instance to get the discovery client and then update the grpc id in the cluster. Ideally the channel pool would be generic such you could store any client in the pool, although maybe we add specific helpers that make it easy to get "well known" clients like session manager, discovery service, etc. Because the LV driver APIs don't have the concept of a gRPC channel in their create/init APIs, having the channel pool also take on the responsibility of remember past resolve calls to the discovery service for the host:port of the driver API is also useful. The caching of these two things (host:port returned from resolve and clients created from grpc-labview) make a huge difference in run time performance if a service is called multiple times (i.e. sweep loop) in a test sequence.

jasonmreding (changes requested):

bkeryan:
> However, this brings in the complication of knowing when to remove invalid/stale channels from the pool because the service is no longer running on that port.

This is why our other channel pools are indexed by hostname:port, not service interface.

bkeryan:
> I don't think this new VI should be creating a DiscoveryClient. It should require that something external to it create the DiscoveryClient. It should not take a cluster as an input (since it doesn't care about the session manager) and instead just take the DiscoveryClient id. We should consider making a ring control with some common grpc device service classes, but also allow raw text for service classes that we don't know about yet.

This seems like a performance problem. I feel like we are going to have to start caching the discovery client and/or the gRPC channel for the discovery service.

bkeryan:
> I have mixed feelings on the channel pool. So it may just be a terminology/naming issue, but I don't really consider this a channel pool as it only exists for the lifetime of a single RPC on the server. Ideally, a channel pool would be utilized so that channels/clients can be reused across multiple RPCs. However, this brings in the complication of knowing when to remove invalid/stale channels from the pool because the service is no longer running on that port.
> 
> If we're trying to avoid taking on the complexity/implementation cost of invalidating the cache for a channel pool, then maybe just tweaking the name is the way to go. However, I'm not really coming up with any good suggestions at the moment. Maybe just calling it something like GRPC Clients.ctl would be better.
> 
> If you want to take on creating a proper channel pool (which I think would be valuable), then I think wrapping it in a class or some other API (functional globals) would be better. This would allow you to remove some duplicated for instance to get the discovery client and then update the grpc id in the cluster. Ideally the channel pool would be generic such you could store any client in the pool, although maybe we add specific helpers that make it easy to get "well known" clients like session manager, discovery service, etc. Because the LV driver APIs don't have the concept of a gRPC channel in their create/init APIs, having the channel pool also take on the responsibility of remember past resolve calls to the discovery service for the host:port of the driver API is also useful. The caching of these two things (host:port returned from resolve and clients created from grpc-labview) make a huge difference in run time performance if a service is called multiple times (i.e. sweep loop) in a test sequence.

I agree with pretty much all of this. It's misleading to call this cluster a channel pool, but it would be okay to call it `gRPC Clients` or something. And I feel like we will need to write an actual channel pool, sooner or later.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Create DC Power Session.vi
bkeryan, line 1: The driver's proper name is "NI-DCPower". It's okay to omit the dash in PascalCase strings but don't replace it with a space.

`Create NI-DCPower Session.vi`
`NI-DCPower Specific Logic.vi`
`Destroy NI-DCPower Session.vi`

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/DCPower Specific Logic.vi
bkeryan, line 1: Removing init/close from this VI makes it simpler, but now it's no longer a standalone VI that you can run directly.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Create DC Power Session.vi
bkeryan, line 1: I just noticed we're not using the same terminology as the driver APIs. Should we use init/close instead of create/destroy?

bkeryan, line 1: Can we make `Location to DCPower Grpc Device.vi` a malleable VI? I'm not sure if it would require wiring an empty `NI gRPC Device Server` as an input or whether it could adapt to the output connector type.

bkeryan (changes requested):

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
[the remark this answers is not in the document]
pbirkhol-ni: I don't think adapting to output works, but adapting to input does. I think Reserve has to show the selector since it won't adapt properly. Unreserve will adapt, but I'm also not opposed to showing the selector. It's not like showing the selector makes it no longer adapt. In short, I don't care either way. As long as Reserve shows the selector I'm happy.

bkeryan:
> I don't think this new VI should be creating a DiscoveryClient. It should require that something external to it create the DiscoveryClient. It should not take a cluster as an input (since it doesn't care about the session manager) and instead just take the DiscoveryClient id. We should consider making a ring control with some common grpc device service classes, but also allow raw text for service classes that we don't know about yet.

I agree. It feels like this VI has too many responsibilities. 

Consider a measurement that reserves two different instrument types, like niScope and niFgen. If this measurement calls reserve twice with different instrument type ids, then it also creates two connections to the discovery service and session management service.

However, I also have a nagging feeling that performing multiple reservations is not good due to lock ordering. It may be better to omit the instrument type id in this case.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Get Session Information.vi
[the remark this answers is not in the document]
subash-suresh: My bad, i have fixed this

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Create DC Power Session.vi
subash-suresh: updated the names of the VIs as requested

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Get Session Information.vi
subash-suresh: I have wired the session information and made the input as a `required` input to avoid this in future

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
subash-suresh: I have wired this input and made it required to avoid such errors in future. 

I have tested the VIs.

subash-suresh:
> > Latest updates on 6/15/23,
> > @dixonjoel, @pbirkhol-ni , @bkeryan I have moved the logic of getting the gRPC device location out of the reserve session api into a seperate helper vi. This is because the reserve sessions API can be used for a device driver that is not part of the grpc device server based on the feedback from Brad in #207 .
> > This new vi is called `Get gRPC device location.vi`, This VI receives the gRPC channel pool as input, and if the discovery service id is 0 then it trys to create a discovery service client. image image
> > The `Create DC Power Sessions.vi` uses the `Get gRPC device location.vi` to get the device server location. Hence the VI has been updated to accept gRPC channel pool. image
> > Based on these updates the measurement logic looks like below, image
> > I am resetting reviewers due to this new update
> 
> I don't think this new VI should be creating a DiscoveryClient. It should require that something external to it create the DiscoveryClient. It should not take a cluster as an input (since it doesn't care about the session manager) and instead just take the DiscoveryClient id. We should consider making a ring control with some common grpc device service classes, but also allow raw text for service classes that we don't know about yet.

I have updated the new VI to take in `discovery service gRPC ID` 
image

and moved the discovery service creation outside into a separate VI called `Create Discovery Client` 
image

This VI is called in the `Initialize NI DC Power Session.vi` before calling the `Get gRPC Device Location.vi` . 
image

subash-suresh:
> > I have mixed feelings on the channel pool. So it may just be a terminology/naming issue, but I don't really consider this a channel pool as it only exists for the lifetime of a single RPC on the server. Ideally, a channel pool would be utilized so that channels/clients can be reused across multiple RPCs. However, this brings in the complication of knowing when to remove invalid/stale channels from the pool because the service is no longer running on that port.
> > If we're trying to avoid taking on the complexity/implementation cost of invalidating the cache for a channel pool, then maybe just tweaking the name is the way to go. However, I'm not really coming up with any good suggestions at the moment. Maybe just calling it something like GRPC Clients.ctl would be better.
> > If you want to take on creating a proper channel pool (which I think would be valuable), then I think wrapping it in a class or some other API (functional globals) would be better. This would allow you to remove some duplicated for instance to get the discovery client and then update the grpc id in the cluster. Ideally the channel pool would be generic such you could store any client in the pool, although maybe we add specific helpers that make it easy to get "well known" clients like session manager, discovery service, etc. Because the LV driver APIs don't have the concept of a gRPC channel in their create/init APIs, having the channel pool also take on the responsibility of remember past resolve calls to the discovery service for the host:port of the driver API is also useful. The caching of these two things (host:port returned from resolve and clients created from grpc-labview) make a huge difference in run time performance if a service is called multiple times (i.e. sweep loop) in a test sequence.
> 
> I agree with pretty much all of this. It's misleading to call this cluster a channel pool, but it would be okay to call it `gRPC Clients` or something. And I feel like we will need to write an actual channel pool, sooner or later.

I have renamed the control to `GRPC Clients.ctl` for now and we will plan on taking the gRPC channel implementation in a later PR as suggested.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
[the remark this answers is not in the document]
subash-suresh: I have added the poly selector for unreserve as well to avoid confusion.
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
subash-suresh: I have updated the gRPC channel pool output to be a side connector and the updated block diagram looks like below
image

image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Create DC Power Session.vi
subash-suresh: I have updated the terminology to match the DC Power VIs i.e., I used init/close.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/DCPower Specific Logic.vi
subash-suresh: what wpuld be the recommended approach here. Is it better to have the logic separated into different VIs

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Create DC Power Session.vi
subash-suresh: Since the input for the NI DC Power init method's input where this `Location to DCPower Grpc Device.vi` VI's  output is mapped is not a strict type def, we made the output as a simple cluster and let LabVIEW handle the reset.
image
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session.vi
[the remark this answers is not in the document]
subash-suresh: udpated the VIs to be part of the library

subash-suresh:
> > I don't think this new VI should be creating a DiscoveryClient. It should require that something external to it create the DiscoveryClient. It should not take a cluster as an input (since it doesn't care about the session manager) and instead just take the DiscoveryClient id. We should consider making a ring control with some common grpc device service classes, but also allow raw text for service classes that we don't know about yet.
> 
> I agree. It feels like this VI has too many responsibilities.
> 
> Consider a measurement that reserves two different instrument types, like niScope and niFgen. If this measurement calls reserve twice with different instrument type ids, then it also creates two connections to the discovery service and session management service.
> 
> However, I also have a nagging feeling that performing multiple reservations is not good due to lock ordering. It may be better to omit the instrument type id in this case. Otherwise, if some measurements reserve niScope then niFgen and others reserve niFgen then niScope, they may deadlock (until the reservation timeout).



> > I don't think this new VI should be creating a DiscoveryClient. It should require that something external to it create the DiscoveryClient. It should not take a cluster as an input (since it doesn't care about the session manager) and instead just take the DiscoveryClient id. We should consider making a ring control with some common grpc device service classes, but also allow raw text for service classes that we don't know about yet.
> 
> I agree. It feels like this VI has too many responsibilities.
> 
> Consider a measurement that reserves two different instrument types, like niScope and niFgen. If this measurement calls reserve twice with different instrument type ids, then it also creates two connections to the discovery service and session management service.
> 
> However, I also have a nagging feeling that performing multiple reservations is not good due to lock ordering. It may be better to omit the instrument type id in this case. Otherwise, if some measurements reserve niScope then niFgen and others reserve niFgen then niScope, they may deadlock (until the reservation timeout).

I have made the instrument type input as optional input and if the users do not give the instrument type input it automatically reserves all the session without filtering based on instrument type. @bkeryan

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
pbirkhol-ni, line 1: This still isn't right. You should not be creating a DiscoveryClient here. You've already done that and the gRPC ID is part of the `gRPC Clients` input. If this VI were designed to be called in isolation then you'd have to create a DiscoveryClient. But it's designed to be called after reserving sessions, which will already create the DiscoveryClient. 

image

pbirkhol-ni: Also, because it isn't designed to be use in isolation, the `gRPC Clients` input should be required.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
pbirkhol-ni, line 1: Use the default if unwired option in the error case.
image

pbirkhol-ni (changes requested):

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
subash-suresh: updated as per suggestion

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
subash-suresh: updated as suggested.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve (N Sessions).vi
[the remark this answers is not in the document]
subash-suresh: Updated the logic to take the ds client id from resolve session call.
image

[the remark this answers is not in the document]
subash-suresh: removed the unwanted error wire

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve (1 Session).vi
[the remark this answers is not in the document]
subash-suresh: Updated all the places where gRPC channel pool was mentioned

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/GRPC Clients.ctl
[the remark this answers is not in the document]
subash-suresh: updated to `gRPC clients`

pbirkhol-ni (approved):
I think this meets the requirements of the PO. I'm not super thrilled with it, but it doesn't commit us to anything long term, so we can always change it later.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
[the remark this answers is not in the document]
subash-suresh: udpated the logic to remove rebundling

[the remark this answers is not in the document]
subash-suresh: updated the comments

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Get gRPC Device Location.vi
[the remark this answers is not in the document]
subash-suresh: updated to use empty stirngs by using the default cluster value
image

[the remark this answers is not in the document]
subash-suresh: marked the `discovery service gRPC ID` as required.
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve (1 Session).vi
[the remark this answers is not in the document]
subash-suresh: marked the error in as recommended as suggested.
image

jasonmreding (approved):
My only comment is that it might be worth trying to extract driver agnostic code into a shared lvlib at some point that is referenced by all our examples. That doesn't need to be done now though. The library could go at the root of the "Example Measurements" folder. However, I don't know how well LV does with referencing files relatively that aren't under or in the same directory as the project file. Even collecting some driver specific APIs in this library would probably be OK. This could be treated as a preview/alpha version of what might become an official public API in vilib at some point.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Reserve Session(s).vi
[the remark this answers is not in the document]
bkeryan: You capitalized the 'p' in "poly" but I guess that's okay. I checked codesearch and there is no consistent standard for "(poly).vi" vs. "(Poly).vi"

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve (Poly).vi
bkeryan, line 1: Uncheck "Show Selector by Default"
image

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
bkeryan, line 1: Select the non-error case and resave

bkeryan (approved):
Approved with suggestions

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve (Poly).vi
subash-suresh: updated as suggested

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
subash-suresh: updated as suggested

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve (Poly).vi
bkeryan: Why did you uncheck it for Reserve?

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
bkeryan, line 1: > Creates and returns an gRPC based NI DC power instrument handle. 

Only use "an" when the following word starts with a vowel.

bkeryan, line 1: > Creates and returns an gRPC based NI DC power instrument handle. 

branding: "NI-DCPower"

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Close NI-DCPower Session.vi
bkeryan, line 1: > Closes the given NI DC power instrument handle.

branding: "NI-DCPower"

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve (Poly).vi
bkeryan: `Measurement Logic.vi` shows the selector for `Reserve` and hides the selector for `Unreserve`. These should be the defaults in the poly VIs.

subash-suresh:
I will make the suggested documentation changes and poly VI setting changes in a new PR.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
subash-suresh: This has been handled in #230

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Close NI-DCPower Session.vi
subash-suresh: This has been resolved in #230

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Initialize NI-DCPower Session.vi
subash-suresh: This has been updated in #230

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/Unreserve (Poly).vi
subash-suresh: I have updated the reserve sesison to have it checked and unreserved to have this unchecked as that is the expected default value. This update can be observed in #230

PR #226, opened 2023-06-23
Remove enum type specialization from template

pbirkhol-ni:

We added a disabled helper VI in `Get Type Specialization.vi` to help with enum type specializations. This was before we decide to do enum specialization automatically. This PR removes that disabled helper VI. Customers shouldn't have to worry about enum specialization. Perhaps advanced customers will want to do custom enum type specialization, but we shouldn't be promoting that in our template.

This PR makes two changes:
- Removes the disabled structure.
- Removes mention of enums from the comment.

This is less confusing.

None. This is just modifying a template VI.

Before:
image

After:
image


WesleyTangNationalInstruments (approved):

dixonjoel:
The Create Enum Spec VI that it was calling is still used in the process of getting metadata to get automatic type specializations, right?

dixonjoel (approved):

pbirkhol-ni:
> The Create Enum Spec VI that it was calling is still used in the process of getting metadata to get automatic type specializations, right?

Yep! It is called in the VI below. We check for any enum controls and type specialize them (unless they have already been manually type specialized).

image

PR #227, opened 2023-06-26
Fix problems with multiple services in one LV project

### LINKED ISSUE #225: Problems with two LabVIEW services in one project
# Bug Report

Creating two MeasurementLink services in one LabVIEW projects makes the second one hang InstrumentStudio when a panel is created for it.

## Repro or Code Sample

1. Do Tools > MeasurementLink > Create Service twice in the same project.
2. Run the two services.
3. In InstrumentStudio, create a soft front panel for each of them.
4. On the creation of the second .sfp, it seems to hang and doesn't show the UI

## Expected Behavior

Should be able to create an arbitrary number of services in each LV project and have them work correctly in TestStand and InstrumentStudio.

## Current Behavior

On the creation of the second .sfp, it seems to hang and doesn't show the UI

## Possible Solution

During scripting of the second project, it may have logic that finds the first library instead. We have had similar problems in the past.

## Your Environment

Windows 10
Version 1.1.0.2 of the .vip files.
MeasurementLink 23.5d700

* OS & Device: [e.g. macOS, Windows, Linux] on [Mac, PC]
* `ni-measurement-service` version [e.g. 0.12.3]
* MeasurementLink version [e.g. 23.0.0d123]
* Python version [e.g. 3.9]

dixonjoel:

Running two services in a single LabVIEW project would cause the second one to hang when being created in InstrumentStudio.

Update some VIs to shared clone reentrancy since they are long-running and need to run in parallel or they are performance sensitive like Update Results.vi

Fixes #225 

Built the Runtime .vip locally, installed it and re-tested a couple scenarios that were failing: 1. When trying to stop the second running service, it would not stop until the first service stopped. 2. Created two .sfps in InstrumentStudio with each of the services which was hanging before.


dixonjoel:
Service Stop Handler, Start Sync, and Update Results were changed to shared reentrant clone execution setting. The rest of the changes are updates due to type propagation or changed subVI calls.

PR #228, opened 2023-06-26
Add documentation for the public MeasurementContext VIs

pbirkhol-ni:

Adds documentation to the 3 public MeasurementContext VIs.

Customers will appreciate knowing how these VIs are supposed to be used.

None.


Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Is Canceled.vi
pbirkhol-ni, line 1: Returns True if the measurement has been canceled by the client. The Measurement Logic VI should poll this property and immediately return when it is True.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Read Pin Map Context.vi
pbirkhol-ni, line 1: Gets the pin map context for the measurement. The pin map context is used to reserve and unreserve sessions using the session management client.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Update Results.vi
pbirkhol-ni, line 1: Sends measurement results to a measurement client, like InstrumentStudio.

This VI must be called with the Measurement Results cluster as the input. You must update all elements of the cluster, even if they didn't change from the previous update. If you want to display cumulative updates, you must write all of the data to the cluster since InstrumentStudio does not cache data from previous updates.

The final result should be written to the Measurement Results indicator of the Measurement Logic VI. It should not be sent using this VI.

TestStand will ignore all updates from this VI and only use the final measurement result.

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Update Results.vi
nstokes-nati: _Sends measurement results **updates** to a measurement client, like InstrumentStudio._ **Use the Measurement Results indicator of the Measurement Logic VI to write the final measurement result. Do not use this VI to write final results.**

_This VI must be called with the Measurement Results cluster as the input. You must_ **always** _update all elements of the cluster,_ **including unchanged values**. 

**InstrumentStudio does not cache data from previous updates so you must write all of the data to the cluster to display cumulative updates.**

_TestStand ignores updates from this VI and only uses the final measurement result._

nstokes-nati: Original in italic, my suggestions in bold. Mainly I want to avoid the contraction and parallelize the application-specific instructions. If the third-from-last sentence is justified by InstrumentStudio's behavior, consider deleting as it seems redundant against the following sentence.

PR #230, opened 2023-06-27
Move instrument agnostic helpers to a separate helper library.

subash-suresh:

This PR moves the instrument agnostic helper methods into a separate helper library that is placed parallel to the project library.
image

This is the same library that has been added to the template in #229 

This update provides users with useful helper vis that can be used in their measurement logic. Putting it in a separate library also allows the old users to reuse the helper libraries by simply adding the library to their older projects, instead of having to copy one vi at a time.

### What testing has been done

- Manually tested the working of NI-DCPower example.
- Manually tested the  working of the reservation and unreservation of sessions in keysight dmm example. I am yet to test it using a simulated device, will be doing that soon.(No changes were made to the session creation or usage of the session so i hope there are no issues here).
- Other examples are not using the helpers from the newly added library as of now.


subash-suresh:
Additionally i have made some minor documentation updates based on the comments in #222 

 - I have updated the documentation of the following Vis
   - `Initialize NI-DCPower Session`
     - Doc: Creates and returns a gRPC based NI-DCPower instrument handle. 
   - `Close NI-DCPower Session`
     - Doc: Closes the given NI-DCPower instrument handle.

- I have also updated the poly VI settings of reserve and unreserve session poly vis as follows,
  - `Reserve (Poly)` - has the option to "show selector by default" checked
  image
  
  - `Unreserve (Poly`) - has the option to "show selector by default" unchecked
  image

pbirkhol-ni:
Looks good to me!

subash-suresh:
Based on discussions i have also updated the other example projects to include the helper library and as a result
the measurement logic of the Keysight DMM measurement example has been updated to use the helpers. The updates can be visualized below,
Before:
image

After:
image

pbirkhol-ni:
> Based on discussions i have also updated the other example projects to include the helper library and as a result the measurement logic of the Keysight DMM measurement example has been updated to use the helpers. The updates can be visualized below, Before: image
> 
> After: image
> 
> @pbirkhol-ni , i am resetting the review due to these new updates

I'm not sure what discussions are are referring to, but I would prefer not to include the library in examples that don't use it. I was trying to figure out where it was used and it was confusing me. There are a lot of ways to get the library (the template, other examples, etc) so there is no need to put it in an example that doesn't need it. I think this applies to the CancellableMeasurement and UIProgressUpdates.

pbirkhol-ni (commented):
I was talking to @dixonjoel and he brought up some possible namespacing concerns. Should we be putting the helper library inside the Measurement library? Then there is another layer of namespacing to avoid conflicts. This will become especially important if you try to create multiple measurements in the same project.

pbirkhol-ni (changes requested):
I was talking to @dixonjoel and he brought up some possible namespacing concerns. Should we be putting the helper library inside the Measurement library? Then there is another layer of namespacing to avoid conflicts. This will become especially important if you try to create multiple measurements in the same project.

dixonjoel:
> I was talking to @dixonjoel and he brought up some possible namespacing concerns. Should we be putting the helper library inside the Measurement library? Then there is another layer of namespacing to avoid conflicts. This will become especially important if you try to create multiple measurements in the same project.

Have you tried creating multiple measurements in one project? Is there logic to avoid making two copies of the top-level helper library in the generator?

Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Measurement/subVIs/Location to Insecure Address.vi
dixonjoel, line 1: Why are these 3 still here if you're removing them from the library above?

dixonjoel (approved):
Once the helpers are removed from the projects that don't need them, this looks ok to me. Before we add this to the template, I'd like to have more consideration of whether we want the library to be nested, multiple services in a project scenarios, and discuss what happens when we want to change / update the helper library or VIs.

bkeryan:
I tested the 34401A example. I had some issues with the LV UI displaying in InstrumentStudio that were unrelated to this PR. Once I worked around those, the example worked fine.

subash-suresh:
> > I was talking to @dixonjoel and he brought up some possible namespacing concerns. Should we be putting the helper library inside the Measurement library? Then there is another layer of namespacing to avoid conflicts. This will become especially important if you try to create multiple measurements in the same project.
> 
> Have you tried creating multiple measurements in one project? Is there logic to avoid making two copies of the top-level helper library in the generator?

In the generator logic in #229 , I have added logic to prevent the addition of duplicate helper library if the project already contains a helper library.
If all the measurements are in the same project, then yes all the measurement would be using the same helper functions and the changes in the helper libraries would affect all measurements in the project.

Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Measurement/subVIs/Location to Insecure Address.vi
subash-suresh: I have removed the older files now

subash-suresh:
> > Based on discussions i have also updated the other example projects to include the helper library and as a result the measurement logic of the Keysight DMM measurement example has been updated to use the helpers. The updates can be visualized below, Before: image
> > After: image
> > @pbirkhol-ni , i am resetting the review due to these new updates
> 
> I'm not sure what discussions are are referring to, but I would prefer not to include the library in examples that don't use it. I was trying to figure out where it was used and it was confusing me. There are a lot of ways to get the library (the template, other examples, etc) so there is no need to put it in an example that doesn't need it. I think this applies to the CancellableMeasurement and UIProgressUpdates.

I have removed the measurement service helper library from the cancellable and the UI measurement

subash-suresh:
As per the comments we are seeing the following possible solutions for the placement of this measurement helper library.
1. Placing the library parallel to the measurement library.(Currently followed in this PR)
  - Pros
    - This will be a common helper file for all the measurement libraries in the project. 
    - The naming is common, so it is clear that these helpers are measurement agnostic.
  - Cons
    - The helpers cannot be modified to serve the purpose of a single measurement, if there are multiple measurements sharing the same helpers in the project. Modification of the helpers might lead to issues as this is referenced by all the other measurement libraries in the project. This is not an issue if each measurement library is in a separate project file.

2. Placing the helper library as a nested library.
  - Pros
    - This will not have any sort of cross-referencing issues as there will one helper library for each measurement. Which means users are free to add or modify these helpers without worrying about other measurements.
  - Cons
    - The naming of the library should be modified to be unique to the measurement, since there can be multiple measurements in the project, and we cannot have similarly named libraries. This will make the helpers look like it is measurement specific.
3. Placing the library parallel to the measurement library but naming them as per the measurement name.
  - I am not considering this since this might lead to the creation of multiple helper libraries that sits parallelly along with the measurement libraries. 

I went with approach 1, because I wanted to portray this as a measurement agnostic helper library and that it could be used by any measurement in the project. I don't see the need for the users to modify this and if the users wanted to add more helpers that are measurement specific then they can be added in the measurement library itself. 

kindly let me know your preferences on the placement of this helper library based on this.

CC: @dixonjoel , @bkeryan , @pbirkhol-ni ,  @jasonmreding

dixonjoel:
> @bkeryan , @pbirkhol-ni , @jasonmreding

dixonjoel:
> As per the comments we are seeing the following possible solutions for the placement of this measurement helper library.
> 

My opinion - keep the library parallel to the measurement library for these examples so we can get them checked in with helper libraries and get more feedback / have more discussion. We can discuss more in the PR for the template / generator and change the examples to match those decisions later. In other words, keep this PR as is.

subash-suresh:
@sreekantham  and @rfriedma , I have added you to this PR to get your thoughts on this implementation

sreekantham (approved):

pbirkhol-ni (approved):

PR #243, opened 2023-07-21
Generate service annotations into serviceconfig JSON

cumitche:

Added support for service annotations when generating the service config JSON for static registration.

- Modified Post-Build Action.vi
  - Replaced the existing implementation of serviceconfig JSON generation with a new helper VI
- Added Generate Service Config JSON.vi
  - Invokes Encode to JSON String.vi for each value rather than generating the entire JSON with one call to the Flatten To JSON primitive
- Added Encode to JSON String.vim
  - Helper VI to JSON serialize a just a key or value rather than a key/value pair
  - Wraps the object passed in with an array, calls Flatten To JSON on it, and removes the excess brackets from the result

Support for service annotations is necessary for the Measurement Library features.

Built the UIProgressUpdates example in its current state and modified to include annotations.  Inspected generated service config files to ensure they were correct.  Loaded example measurement statically and saw appropriate annotations in InstrumentStudio.


cumitche:
image

cumitche:
image

image

cumitche:
image

image

bkeryan:
Please set the PR title.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Encode to JSON String.vim
cumitche, line 1: I'm unsure of the type of the anything input.  I tried to set it up as a variant, but I would have a broken wire without putting a type in it.  It shows up as a cluster containing a string, but it appears to work as intended with I pass it something other than a string.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Generate Service Config JSON.vi
pbirkhol-ni, line 1: You should make most of these inputs required. Everything that you have wired should be required (except for the error terminals).

Source/Runtime/MeasurementLink Measurement Server/Helpers/Encode to JSON String.vim
dixonjoel: I think the way you have it is OK. I don't fully understand malleable VIs, but the Flatten to JSON says
> This input accepts arrays and clusters of Booleans, floating-point numbers, and strings. anything can also accept an array of clusters, or a cluster of arrays, of these data types. This input does not support other data types, such as enums, refnums, file paths, and fixed-point numbers.

And you're passing an array of strings.

pbirkhol-ni: I don't know either. I would probably just make it a string instead of a cluster of a string, though, since that is going to be the most common use case.

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Generate Service Config JSON.vi
pbirkhol-ni: You could probably create a subVI for creating the key/value pairs in the json. That would avoid a lot of the duplication for displayName, serviceClass, etc. The contents of the subVI would basically be the code you already have for the annotations key/value pairs (see below). The subVI could take a boolean input for whether to include a trailing comma (maybe have it default to True). 

image

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Encode to JSON String.vim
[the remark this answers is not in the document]
cumitche: Made the anything input required.  Made the Name and Value inputs required on Add JSON Name Value Pair.vim.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Generate Service Config JSON.vi
cumitche: Created Add JSON Name Value Pair.vim to reduce duplication.  Made Display Name, Service Class, Provided Interfaces, and Path required inputs.

cumitche:
image

image

PR #244, opened 2023-07-25
Update measurement plugin to launch discovery service just-in-time

Jayaseelan-James:

This PR updates the measurement plugin to launch discovery service just-in-time before registering to it.

Starting a measurement service when discovery service is not running will throw error `5053 - Failed to read the discovery service key file. Ensure that the discovery service is running.` 

With this PR's changes, discovery service will be launched by the measurement if it is not running.

Manually tested the following the changes for the below cases,
1. Ran a measurement service without DiscoveryService.exe running. Discovery service was launched and the measurement was registered to it.
2. Ran a measurement service with DiscoveryService.exe already running. The measurement was registered to the already running discovery service.

Snapshots of File Changes/Additions:

`Read Discovery Service Key File.vi`:
Read Discovery Serv Key

  
`Ensure Discovery Service Started.vi`:
image

Check   Launch Discovery Service 2

  
`Start Discovery Service.vi`:
Start Discovery Serv

image

`Get Discovery Service Address.vi`:
image

`Open Discovery Service Key File.vi`:
image

`Error Strings Global.vi`:
Error Strings Global


Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
dixonjoel, line 1: You'll have to get the Program Files directory from an environment variable and not hard-code it to C:\Program Files. Customers can install NI software to a D: drive for example. See the Python version:
https://github.com/ni/measurementlink-python/blob/942483b17bc3fd718dd16fc94266940ba9dc365e/ni_measurementlink_service/_internal/discovery_client.py#L293

dixonjoel: I'm not sure offhand what the best way to do this in LabVIEW, but here's one possibility (using the registry)
https://forums.ni.com/t5/LabVIEW/How-do-I-programmatically-find-Program-Files-directory/td-p/205804

dixonjoel: This seems to work: Use 'Get System Directory.vi'.
image

bkeryan, line 1: In C# and Python the polling loop opens the key file, rather than merely check that it exists. This may behave differently due to the share mode.

bkeryan (changes requested):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
bkeryan, line 1: When this VI is called, the key file path has already been queried. You could pass it in as a parameter.

pbirkhol-ni, line 1: Perhaps this has already been decided in other implementations, but 30 seconds seems like a long timeout.

pbirkhol-ni (approved):
I don't have much feedback other than what Joel and Brad have already covered.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
Jayaseelan-James: Thanks for the suggestion, @dixonjoel! I have updated the logic to obtain using the `Get System Directory.vi` but I found that this VI would return `<OS Directory root>:\Program Files (x86)` if it is executed in a 32-bit system. So, I have updated the logic to take the root directory from the `Get System Directory.vi` and append it with `..\Program Files\`. 
App Files

Source/Runtime/MeasurementLink Measurement Server/Helpers/Read Discovery Service Key File.vi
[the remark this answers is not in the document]
Jayaseelan-James: Agreed. Updated the logic to invoke the VI in `Get Discovery Service.vi`.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Check and Launch Discovery Service.vi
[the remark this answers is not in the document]
Jayaseelan-James: Updated. Thanks for the suggestion!

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
dixonjoel: No that doesn't seem correct. The whole point is to get the correct program files directory where things are installed independent of drive path or whether it's 64 bit or 32 bit. So I think you should take it as is and append 'National Instruments\Shared\MeasurementLink'. You don't have a 32-bit system where it's installed in c:\program files do you?

Source/Runtime/MeasurementLink Measurement Server/Helpers/Check and Launch Discovery Service.vi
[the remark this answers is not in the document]
Jayaseelan-James: Fixed it.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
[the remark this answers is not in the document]
Jayaseelan-James: Yeah, that's right. Since there was only one path separator, I didn't wire the `replace all?` but yeah considering any future changes, I have now wired it!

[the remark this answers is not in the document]
Jayaseelan-James: Yes! `Strip Path` requires the path separator's to be proper, which is used to provide the working directory to launch the Discovery Service at the right port. Hence, I'm leaving it untouched.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Check and Launch Discovery Service.vi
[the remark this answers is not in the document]
Jayaseelan-James: Updated the File name to `Ensure Discovery Service Started.vi`, but I would like to understand the reason behind naming it as "ensure discovery service started", because I feel that internally what we do reflects the statement, "Check and Launch Discovery Service". ?

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
Jayaseelan-James: Updated the flow to check by opening the key file & close it once it's found.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global.vi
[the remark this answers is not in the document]
Jayaseelan-James: Updated the error message to `Timed out waiting for discovery service to start.`

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
Jayaseelan-James: @bkeryan, @pbirkhol-ni Bit curious to know the reason behind settling with a 30secs timeout?!

Jayaseelan-James: If I understand correct, MeasurementLink runs on `64-bit` machine and so is Discovery Service, right? In that case, if the user for any reason uses LabVIEW 32-bit version on a 64-bit machine, it would give us `<Root>:\Program Files (x86)\` but the discovery service will be present in `<Root>:\Program Files\`. 

Please correct me if I'm wrong.

PS: Sorry that I mentioned it as 32-bit system in my earlier comment instead of 32-bit LabVIEW application.

Jayaseelan-James: Also, now I have added an target type check!
image

dixonjoel: We don't generally support 32 bit LabVIEW versions so I don't think you need the hard-coded Program Files here.

Jayaseelan-James: Ohh, Sure! Removed the hard-coded Program files.
image

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
bkeryan: The timeout was originally added in ASW PR 276020. Maybe @Vasantha-Kumar-J remembers?

Anyway, I think 30 seconds is fine. The discovery service should never take more than 30 seconds to start, even on a heavily loaded machine, except when you have a debugger attached. And it's short enough that users will likely see a timeout error before they can open Task Manager and kill the process.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Check and Launch Discovery Service.vi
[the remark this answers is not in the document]
bkeryan: If the discovery service is still running, opening for write should fail, which prevents deleting the key file. See the discussion in ASW PR 276020.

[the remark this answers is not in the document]
bkeryan: "ensure discovery service started" describes the purpose of the function. The post-condition is that the discovery service has been started. (I guess "running" is even better, since "running" is a state and "start" is a state transition, but whatever.)

"check and launch discovery service" describes the implementation. Also, what does it check? And it doesn't always launch the discovery service, so you end up renaming it to "check whether key file exists and launch the discovery service if needed", which is much more verbose.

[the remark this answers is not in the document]
Jayaseelan-James: Sure, Understood.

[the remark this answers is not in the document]
Jayaseelan-James: Updated the check, @bkeryan. Please review and let me know if any further changes are required.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Discovery Service Address.vi
bkeryan, line 1: This VI sometimes returns error 7, "File not found" without starting the discovery service.

Steps:
- Click the **Run Continuously** button
- Open **Task Manager** and kill the discovery service
- The VI error out will temporarily change to 7

If you add a probe to `Ensure Discovery Service Started.vi` where it clears the error, you will see the probe change from 5 to 7. The error code path does not start the discovery service.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Ensure Discovery Service Started.vi
bkeryan, line 1: [style] FYI, the rest of these VIs use lowercase control/indicator labels, but I guess this component is inconsistent about that

bkeryan (changes requested):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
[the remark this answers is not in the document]
Jayaseelan-James: Thanks for the info, I'll update it!

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Discovery Service Address.vi
Jayaseelan-James: Thanks for point it out. I have updated the logic in `Ensure Discovery Service Started.vi` to handle the `Error 7: File not found.`

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
bkeryan, line 1: FYI: this loop's error wire doesn't have a shift register.

image

However, it's not a problem in this case because the loop breaks when there is an error.

If the loop didn't break when there is an error, then a shift register would be necessary to pass the error to the next loop iteration.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Open Discovery Service Key File.vi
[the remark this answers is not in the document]
bkeryan: You didn't answer my previous question:

> Can you achieve the same goal by calling `Read Discovery Service Key File.vi` instead of splitting it up?

Instead, you changed this code to open the file for write instead of opening it for read+share:

image

Does this introduce any new problems? The most obvious one is a race condition that prevents the discovery service from writing the key file. If this code used `open or create`, it would create a new key file. If the discovery service tried to write the key file while this code had the key file open, then I think it would get an error.

However, since you used `open`. I guess it's okay.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Discovery Service Address.vi
bkeryan: Ok, I can't reproduce this problem anymore. Good.

bkeryan (approved):

Issue #245, opened 2023-07-25
Support PPL LabVIEW UIs


pbirkhol-ni:
## Problem to Solve

Some LabVIEW UIs will need to use a source distribution or PPL in order to be loaded by InstrumentStudio. While this use case is supported in development, it doesn't work well for building an exe or deploying an application. We should add support for building and deploying PPL UIs.

## Proposed Solution
The Post-Build Action.vi handles copying measurement UI VIs to the build output directory. We should modify it to handle PPLs. 

[AB#2466421](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2466421)

bkeryan:
> The Post-Build Action.vi handles copying measurement UI VIs to the build output directory. We should modify it to handle PPLs.

Or allow including dependencies. See https://github.com/ni/measurementlink-labview/issues/240#issuecomment-1646085903

Issue #240, opened 2023-07-18
`Keysight 34401A DMM Measurement` UI doesn't work when statically registered


bkeryan:
# Bug Report

`Keysight 34401A DMM Measurement` UI doesn't work when statically registered because the built EXE doesn't include `Keysight 34401A Function.ctl`.

## Repro or Code Sample

Build an EXE for `Keysight 34401A DMM Measurement`.
Statically register it.
Open its `.instudioproj`. 

## Expected Behavior

Displays UI.

## Current Behavior

Displays blank panel. (Also see #239 )
![image](https://github.com/ni/measurementlink-labview/assets/17914399/985e0deb-2159-4918-a9bd-8fb7c553182c)

## Workaround

Manually copy `Keysight 34401A Function.ctl` to the build directory.

## Possible Solution

Build a source distribution or packed project library for the UI.

## Context

Trying to test `Keysight 34401A DMM Measurement` with static registration.

## Your Environment

* OS & Device: Windows
* `ni-measurement-service` version: 1.1.0.3
* MeasurementLink version: 23.5
* LabVIEW version: 2020 SP1

[AB#2459128](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2459128)

bkeryan:
This post-build action VI is responsible for copying `Measurement UI.vi` and the owning `.lvlib` into the built EXE's directory:
![image](https://github.com/ni/measurementlink-labview/assets/17914399/9fc44125-77f1-4296-be19-f24d48412b43)

PR #255, opened 2023-08-08
Add support for LabVIEW PPL LabVIEW UIs in built apps

pbirkhol-ni:

This PR fixes the issue in #240 and also partially fixes the issue in #245. It is not necessarily the ideal solution, but it will get customers further than they are today.

- Modify the `Post-Build Action.vi` to support copying LabVIEW UIs that are in PPLs. Prior to this change, trying to copy a LabVIEW UI in a PPL would error, which would break your build spec. Note that the PPL must be in the same directory as the `Measurement UI.vi`.
- Update the `Keysight 34401A DMM Measurement` example to use a PPL for its UI. This fixes the issue in #240 and also provides an example for how to use PPL-based UIs.
- Update documentation.

Any reasonably complex UI will require external dependencies. I anticipate this to be a common workflow and we should have a solution, even if it isn't the ideal solution.

I tested using standalone Measurement UIs and PPL UIs, both with dynamic registration and static registration.


Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Measurement/Get UI Details.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Build Assets/Post-Build Action.vi
pbirkhol-ni, line 1: image

I kept as much existing logic as I could. The only change is that if the LabVIEW UI is in a PPL, we only copy the PPL and don't attempt to copy the VI itself (since it is already in the PPL).

dixonjoel (approved):

PR #257, opened 2023-08-10
For built exes, copy the entire containing directory for PPL UIs

pbirkhol-ni:

I received some offline feedback for #255. It didn't handle the case where the UI PPL had external dll dependencies. This change will copy the entire containing PPL directory during the build so that any necessary dlls also get copied. It also updates the Keysight example and documentation to mention that PPLs should be placed in a subdirectory next to the `Measurement Logic.vi.`

We don't expect UIs with dll dependencies will be common, but it is certainly possible and we should support it.

I created a measurement with a UI that uses the LabVIEW keyboard VIs, which has a dependency on `lvinput.dll`. I verified that before my change, the UI wouldn't load in InstrumentStudio. After the change, the dll gets copied and the UI loads.


Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Measurement/Get UI Details.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Build Assets/Post-Build Action.vi
pbirkhol-ni, line 1: image

dixonjoel: The comment on the 'Loose VI' case looks like it needs to be updated.

dixonjoel (approved):

dixonjoel:
Please update the PR title before submitting for accuracy when looking back at PRs.

pbirkhol-ni:
> Please update the PR title before submitting for accuracy when looking back at PRs.

Ugh. Why is github like this?

Source/Runtime/MeasurementLink Measurement Server/Build Assets/Post-Build Action.vi
pbirkhol-ni: That case is only copying the VI so I think the comment is fine. Do you have a suggestion? Looking at that diagram, the wiring is overlapping which I will probably fix anyway.

pbirkhol-ni: Okay, I updated that comment to be more generic.

Source/Example Measurements/Keysight 34401A DMM Measurement/README.md
jasonmreding, line 1: As structured, this example effectively builds the same library twice, once as a packed library for the UI and once as an exe for the service. Since the code for both the service backend and the UI front end is included in the same library, you are effectively building and distributing it twice. This might be OK for small services and UIs that don't have a lot of code or dependencies. However, it will become more problematic/undesirable as the dependency tree for each portion grows. For the exe build spec, you can actually exclude the Measurement UI.vi even though we explicitly include it currently when it's not needed. For packed library build specs, you don't really have an option. Anything that is public/protected in the top level library is automatically included and you can't override the behavior.

From a build / code organization standpoint, it would be better to package the UI VI in its own library that is distinct from the library used to implement the backend service. The exe build spec would then reference one library and the ppl build spec would reference the other library. This organization of the code would better scale to projects of any size even though it probably seems like overkill for small/simple projects. However, I don't think the initial example is that much more complicated, and I would rather have a starting example that scales rather than have users start down the simple path and then have to refactor later if their project gets big enough that they start caring.

The comment about the exe build spec including the Measurement UI.vi probably pertains to all of our examples. We can probably clean that up as separate bug/tech debt work item if needed.

jasonmreding (changes requested):

bkeryan:
> > Please update the PR title before submitting for accuracy when looking back at PRs.
> 
> Ugh. Why is github like this?

The default behavior probably makes more sense for forks that don't have `users/foo/` boilerplate.

FYI, there is an action to run checks on the PR title: https://github.com/marketplace/actions/pr-title-checker

Source/Example Measurements/Keysight 34401A DMM Measurement/README.md
pbirkhol-ni: I agree with everything you're saying. And I think it would be reasonable to set up an example this way. However, then the question is whether we do that for the template too and have all customers inherit that extra complexity. Would you be okay just doing it for the example?

jasonmreding: By template I assume you mean the code generator to generate a new measurement service and not the Measurement UI.vit that was [recently added](https://github.com/ni/measurementlink-labview/pull/251)? My personal opinion is that yes we should change the code generator so the default project organization sets you up for the best chance of long term success without a lot of rework. In my opinion, that includes clearly organizing the code/build to indicate the front end code is distinct and separate from the back end code. They are two separate run times and execute in different processes. I realize that is a concept that will not be obvious to most new users, but hopefully organizing the project in this manner will help them come to understand that over time. The build/debug work flow for both sets of code is significantly different. However, I haven't taken the time to fully think through the ramifications to the post-build action and relative pathing to the UI VI for simple standalone VI use case. If it significantly complicates the standalone VI case, and we think that is the predominant use case, then we may want to just live with updating the examples. You may want to touch base with a product owner on this. I don't view updating the generator in this PR as a requirement. I'll take whatever incremental progress I can get.

pbirkhol-ni: By template I mean the Measurement Generator. Should it be generating a UI library and a measurement library (and the helpers library)? We discussed this earlier in the week and we weren't sure the added complexity was necessary. I have a meeting to discuss this with Cristina today, but in the meantime I updated the PR so that the Keysight example uses a separate library for the UI. I think this should satisfy your concerns and also give customers a better starting point for PPL-based UIs.

With my latest change, I think this PR is okay to submit and we can revisit updating the template/generator later.

pbirkhol-ni:
> The exe buildspec is still pulling in Measurement UI.vi which isn't necessary. This is something that we should fix across all of our examples and the generator. If you want to address this in a follow on PR with a more comprehensive set of changes, I'm fine with that.

Sure. I'll do a quick follow-on PR.

PR #258, opened 2023-08-11
Remove the Measurment UI from the Always Included section of the build spec

pbirkhol-ni:

The `Measurement UI.vi` doesn't have to be included in the measurement build spec. This PR removes it from the examples and also prevents it from getting added to generated projects.

This will avoid adding unnecessary VIs to the exe, potentially making it smaller.

I generated a new measurement and verified it no longer included the `Measurement UI.vi` in the build spec. I also build and ran one example.


Source/Generator/_Editor/Create Build Specification.vi
pbirkhol-ni, line 1: I added a new case to this inner case structure that is a no-op for the `Measurement UI.vi`. The other cases add the VIs to the build spec.
image

jasonmreding (approved):

dixonjoel (approved):

PR #259, opened 2023-08-11
Support absolute paths when launching the DiscoveryService

pbirkhol-ni:

On my dev machine, `MeasurementLinkServices.json` has an absolute path to `NationalInstruments.MeasurementLink.DiscoveryService.exe`. The absolute path was causing the new DiscoveryService launching logic to error. This PR fixes the issue so that both absolute and relative paths to the DiscoveryService are supported.

I don't know that customers need this since we always install a relative path, but I want it on my machine. Also, if we ever change to use an absolute path to the DiscoveryService, this change will make it work without issues.

Tested both absolute and relative paths to make sure both work.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
pbirkhol-ni, line 1: I added logic that detects whether a path is absolute or relative. If it is absolute, nothing is appended to the path. If it is relative, the existing logic is maintained where the relative path is converted to an absolute path.

image
image

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
dixonjoel, line 1: nit: you could do a . reference of the discovery path to avoid two unbundle nodes.
image

Issue #235, opened 2023-07-13
Upgrading to grpc-labview 1.0.0.2 or later breaks measurement service


cumitche:
# Bug Report

Upgrading grpc-labview to a newer version that what is provided in the release assets causes InstrumentStudio to be unable to connect to running LabVIEW measurement services.

## Repro or Code Sample

Upgrade grpc-labview from 1.0.0.1 to 1.0.0.2 or later.  Run a LabVIEW measurement service.  Add the service to a large panel in InstrumentStudio.

## Expected Behavior

Measurement service should load UI and run in InstrumentStudio.

## Current Behavior

![image](https://github.com/ni/measurementlink-labview/assets/77293599/9c9ebf6a-f0b1-4bae-958e-5ccaac8546ee)
Measurement service fails to load UI and displays an error stating that it cannot connect to the measurement.

## Context

Attempted to use the latest release of grpc-labview (1.0.0.5) to regenerate a client and discovered the breakage.  Traced it back to 1.0.0.2.  This issue is not blocking me, I can continue to use 1.0.0.1 for now.

## Your Environment

* OS & Device: Windows 10 on PC
* `ni-measurement-service` version 1.0.1.1
* MeasurementLink version 23.3.0f793
* Python version N/A

[AB#2455408](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2455408)

pbirkhol-ni:
I looked into this today. The issue is caused by `grpc-labview` adding enum support. Before `grpc-labview` had enums, it would use integers as enum values. The generated messages would use integers in place of enums and that is what the generated v1 and v2 measurements use in `measurmentlink-labview` (the v1 and v2 measurements don't use enums directly, but they do use google types that are enums). When upgrading to `grpc-labview v1.0.0.2` or later, the expectation is that any enum messages will use LabVIEW enums and not integers. So if you don't regenerate the gRPC messages, setting response data will throw an unhandled exception. See [here](https://github.com/ni/grpc-labview/blob/cab29387ed6d9e0889626b3233e7ce74afac3c2b/src/grpc_interop.cc#L401C42-L401C57). This happens in `measurementlink-labview` during calls to `GetMetadata.`

[...]

PR #264, opened 2023-08-31
Update clients and servers to use grpc-labview 1.0.0.7

pbirkhol-ni:

#235

- Regenerate all of our clients using grpc-labview 1.0.0.7. 
- Create documentation for how to update the clients and servers.
- Update the .vipb files to require grpc-labview 1.0.0.7 or later.
- Create two new projects: `Measurement Service V1.lvproj` and `Measurement Service V2.lvproj`. Prior to creating these new projects, the V1 library, V2 library, and Measurement Server libraries were all in the same project. This was causing linking issues during generation since the Measurement Server library depends on the V1 library and V2 library. Splitting them all up into separate projects resolves those issues and makes regeneration much easier (and hopefully automatable one day).

Customers will want to use the latest version of grpc-labview

Ran some shipping examples.


Source/Runtime/MeasurementLink Discovery Client/RPC Messages/FlatToRichni_measurementlink_discovery_v1_EnumerateServicesRequest.vi
dixonjoel, line 1: Did you get a chance to ask about all this FlatToRich and RichToFlat API, e.g. what it's used for and if we can avoid it?

docs/Regenerating Server and Client.md
dixonjoel, line 1: Great! Thanks for writing this up.

dixonjoel (approved):

Source/Runtime/MeasurementLink Discovery Client/RPC Messages/FlatToRichni_measurementlink_discovery_v1_EnumerateServicesRequest.vi
pbirkhol-ni: I talked to the product owner a few weeks ago. Here is what he said:

> With regards to the intermediate files (we call serialization typedef) are introduced as part of the support for Oneof type. I have previously pointed this out to my team that these files only be needed for the clusters with oneof type and add unnecessary overhead for all other cases.
Team is looking at improving this behavior. Having these files will have any functional impact. But please let me know otherwise

I mentioned to him that we were less concerned about the performance than we are about having to ship hundreds of extra files. But for now, there is nothing we can do other than not upgrade, which I don't think is a realistic option.

PR #266, opened 2023-09-08
Fix observed hangs with LV measurements for certain situations of service launches

jonathanmendez:

Fixes [AB#2508868](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2508868) and [AB#2464807](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2464807)

In the existing implementation of the service startup code, there is a race condition caused by a delay between registering the measurement service with Discovery Service (enabling clients to try to call the measurement service) and registering the user events for handling calls to the measurement service (enabling the measurement service to receive and react to calls). If a client makes a call to the measurement service between those two pieces of code, the measurement service never handles the call and the client waits indefinitely.

The fix moves the event registration before the measurement is registered with Discovery Service. The events are registered and the registration refnum is stored in the service private data to later be wired to the event handler. Through testing, the registration call is sufficient to be ready to handle calls, even if the event handler structure has not been reached yet.

Tested the implementation using a reproducing TestStand sequence that repeatedly calls and kills a measurement. Before the fix, a hang would consistently occur within a few minutes. After the fix, I have not observed a hang over many minutes and repeated runs.


dixonjoel:
Would you be able to post some screenshots for the relevant changes? I'm not sure if I need to look at all these files or if most are just type prop that had to be saved.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementServiceV1/Start Sync.vi
pbirkhol-ni, line 1: Rather than unregistering for user events here, it seems more natural to do it in Destroy now. Same comment applies to V2.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementServiceV1/Register gRPC Methods and Messages.vi
pbirkhol-ni, line 1: This is pretty minor, but I wonder if it would make sense to just rename this to `Register` since it is registering a lot of things now. You would have to make the same change to the V2 and to the interface.

pbirkhol-ni (approved):

pbirkhol-ni:
> Would you be able to post some screenshots for the relevant changes? I'm not sure if I need to look at all these files or if most are just type prop that had to be saved.

I found that the only VIs with real changes are:
- `Register gRPC Methods and Messages.vi`
- `Start Sync.vi`

jonathanmendez:
> Would you be able to post some screenshots for the relevant changes? I'm not sure if I need to look at all these files or if most are just type prop that had to be saved.

Yes, I'm working on that now. Unfortunately I ran out of time at my computer as I was posting the PR earlier.

jonathanmendez:
Class private data:
image

Register gRPC Methods and Messages:
image

Start Sync (before/after):
image
image

The rest are just recompiles due to adding data to the class.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementServiceV1/Register gRPC Methods and Messages.vi
jonathanmendez: Done.
Renamed VI and updated icons (before/after of V1 shown, same done for V2):
image
image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementServiceV1/Start Sync.vi
jonathanmendez: Good idea. Done. Screenshots below of V1, though both V1 and V2 have same changes.

Start Sync (before/after):
image
image

Destroy:
image

PR #267, opened 2023-09-08
Minimize presence of measurement services when running as a built application

jasonmreding:

When running a measurement service as a built application:
- Avoid popping modal dialogs and showing front panels. Even if the front panel is updated to hidden after running, the window still flickers into view briefly. This is distracting so avoid it when possible for the built app case. Sadly, I haven't been able to figure out a way to eliminate the flicker for the startup VI of the built app.
- Exit when an error is encountered and the service is no longer responding to rpc requests.

Behavior when running a measurement service interactively through the IDE remains unchanged.

It makes the service behave more as a console application than a GUI application which is what one would expect for a background service.

Manually tested built apps after the changes and running measurement services interactively through the IDE.


Source/Generator/MeasurementLink Measurement Template/Run Service.vi
jasonmreding, line 1: image

The SubVI Node Setup for Run Service was also modified to no longer show the front panel when called (it took me forever to figure out this was there). This isn't needed since we're explicitly showing the front panel within Run Service.vi when not executing in the runtime engine.

image

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
jasonmreding, line 1: The only changes were to the beginning and end of the VI. We no longer open the front panel when executing in the runtime engine. This prevents a brief flicker of the front panel window before it is hidden.

image
image

And we no longer launch modal dialogs for errors when executing in the runtime engine.

image
image

pbirkhol-ni: I would be in favor of not having the simple error handler at all and just having an error indicator on the front panel. If a customer ever wants to automate testing measurements in the dev environment (like we do in our TestStand test suite), that simple error handler could cause issues.

pbirkhol-ni: Whether you agree or not with removing the Simple Error Handler, there is also a One Button Dialog in the False case of the inner case structure that should be in a conditional case structure (or removed entirely and made an error).

Source/Generator/MeasurementLink Measurement Template/Run Service.vi
pbirkhol-ni: It seems like the `FP.Open` Invoke node could just be called unconditionally because of the new `FP.Open` in the other `Run Service.vi`. This would mean that in the IDE, the front panel would be hidden but then would immediately be shown when `Run Service.vi` is invoked.

pbirkhol-ni (approved):

pbirkhol-ni:
> At one time there was a long forum post about how to avoid the flicker of a startup app, but I can't find it anymore. I don't recall if there was a good solution, though, or just workarounds to minimize the flicker.

[Here](https://forums.ni.com/t5/LabVIEW/Is-it-possible-to-write-an-application-in-LabVIEW-2020-that-has/td-p/4292930) is the thread. There are some suggestions on how you _might_ achieve it, but also comments that it might not be possible.

Source/Generator/MeasurementLink Measurement Template/Run Service.vi
jasonmreding: I'm not sure I understand what you are proposing. You want the FP to be hidden even when running in the IDE and then become visible again after execution?

pbirkhol-ni: There are two `FP.Open` Invoke nodes with your proposed change. One in the `Run Service.vi` from the generator and one in the `Run Service.vi` from vi.lib. I don't think the `FP.Open` in the generator `Run Service.vi` needs to be in a conditional disable. It can just always be called. That will result in the following:
- When using the RTE, the front panel gets hidden and stays hidden.
- When using the IDE, the front panel gets hidden (this happens even before the subVI is called). Then in the vi.lib `Run Service.vi`, it will become visible.

pbirkhol-ni: This is pretty minor feedback. You could argue that the code is clearer with both conditional disables.

jasonmreding: > When using the IDE, the front panel gets hidden (this happens even before the subVI is called). Then in the vi.lib Run Service.vi, it will become visible.

If you do this, the generator Run Service.vi remains invisible after running unless you double click to open again from the project window. That seems like pretty bad UX. If you want it to remain invisible only while running, you would need to make the FP visible again before exiting. You can accomplish similar behavior by updating the front panel transparency to 100% while running. That's relatively easy to do if that's the behavior you are after.

pbirkhol-ni: I see now. For some reason I thought that both invoke nodes were for the same VI (the vi.lib `Run Service.vi`). I am dropping this issue.

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
jasonmreding: Diagram after latest changes:
image
image

The modal dialog in the error case is now gone. I updated Validate Configuration Metadata.vi so that it generates an error if validation fails rather than have a separate Boolean output. Because of that, the call moved outside the case structure, and I removed the nested case structure. I moved the call ahead of Get Discovery Service Address.vi to clean up the wiring a bit since order really shouldn't matter. To ensure there are no unintended consequences of this, I added an error case structure to Get Discovery Service Address.vi so it does no work if called with an error. This ensures error in is preserved on error out.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global.vi
jasonmreding, line 1: Added 5057 error for metadata validation. I updated the error explanation slightly.

image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Discovery Service Address.vi
jasonmreding, line 1: Added error case structure.

image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Validate Configuration Metadata.vi
jasonmreding, line 1: IsValid Boolean indicator and string error details indicator are now gone. Instead, that information is conveyed via error out.

image

Source/Generator/MeasurementLink Measurement Template/Run Service.vi
jasonmreding: Latest version. When running in the IDE, we now hide the front panel when running and make it visible again when we stop running.

image

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
pbirkhol-ni: The error wire is not wired through Validate Configuration Metadata.vi. We really should get some VI analyzer tests running on our code base. All talk to Adam about that.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
pbirkhol-ni, line 1: It doesn't seem like this VI meaningfully changed. Let me know if I'm missing something.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
jasonmreding: Doh, thanks for catching that. It's fixed now.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Start Discovery Service.vi
jasonmreding: Nothing changed. If memory serves, it got saved because one of its sub VIs changed and needed to recompile (I think the global error VI). Technically, if we ensure all VIs are checked in as source only, it would avoid these sorts of spurious changes. Perhaps that's another VI Analyzer / PR gate style check we could add at some point.

pbirkhol-ni: > Nothing changed. If memory serves, it got saved because one of its sub VIs changed and needed to recompile (I think the global error VI). Technically, if we ensure all VIs are checked in as source only, it would avoid these sorts of spurious changes. Perhaps that's another VI Analyzer / PR gate style check we could add at some point.

I was talking to Ryan yesterday about writing a script to go through and change that setting on all of our VIs. If we do that and we also set the default on all of our projects to "Separate compiled code from new project items", then we would be in a lot better shape. It's not a substitute for VI analyzer, but it is an easier short term solution. I'd have to lock down the measurementlink-labview repo before doing a change like that, though.

PR #268, opened 2023-09-13
Update scripted build spec that is generated for measurement services

jasonmreding:

Update the scripted build spec for a measurement so that:
- Fast File Format is enabled. This roughly halved the size of the exe for the continuous update example service. It also helps eliminate some issues with the startup VI front panel window flickering into view briefly. 
- Override the startup VI so that the runtime position of the front panel window is minimized. This is the second piece that helps eliminate window flickering when the built app is started.

This should eliminate any remaining artifacts of the front panel window flickering briefly into view when starting the built application for the measurement service. Note: this requires use of the LV 2023 RTE or newer to eliminate the flicker. There is a bug present in previous LV versions where the flicker still occurs otherwise.

Instead of enabling fast file format, we could instead eliminate the flicker by moving the startup VI outside of a library. There seems to be a LV bug where the flicker occurs when the startup VI is owned by a library and doesn't occur if it isn't owned by a library. However, after some internal discussions, the size reduction to the built exe fast file format also seemed appealing so we decided to go that route instead.

- Manually created a new measurement service with the updated generator code and validated the buildspec is scripted with the desired configuration.
- Manually tested starting a measurement built into an exe with these buildspec settings and verified the startup flicker no longer occurs.


Source/Generator/_Editor/Create Build Specification.vi
jasonmreding, line 1: Changes to this VI are the additions of the 3 sub VI calls with the yellow banner.

image

Source/Generator/_Editor/NI_AB_API_EXE_Extensions/Create New EXE Extensions.vi
jasonmreding, line 1: This logic was more or less stolen from the existing vilib VI for creating a new exe buildspec. It has been modified to return the extensions class I introduced which is what allows to access data that is otherwise protected. This then allows to me expose public APIs off of this class for functionality that isn't exposed publicly in the app builder APIs installed to vilib.

image

Source/Generator/_Editor/NI_AB_API_EXE_Extensions/Set Fast File Format.vi
jasonmreding, line 1: image

Source/Generator/_Editor/NI_AB_API_EXE_Extensions/Set Window Run-Time Position.vi
jasonmreding, line 1: image

jasonmreding:
Other design/implementation notes:

Here is what the project looks like with these additions:
image
image

I placed the new class files under the "_Editor" folder rather than a sibling directory so they wouldn't show up as commands in the Tools menu. Since this is a sub-directory of an existing directory, I assume were good to go from a packaging perspective, but I haven't verified this yet.

In order to programmatically script these build spec changes, I was forced to create a new class that derives from the existing classes defined by the app builder vilib VIs. The naming of the class adopts the naming convention used for the app builder types. So `NI_AB_API_EXE_Extensions.lvclass` extends the existing `NI_AB_API_EXE.lvclass` defined by app builder.

This class is nothing more than a wrapper around the existing class that allows me to access protected data needed to configure the build spec as desired. App builder contains APIs to do everything I want, they just aren't exposed publicly for some reason. By creating a new "dummy" build type, I can then access these protected APIs and add public APIs on the "dummy" type to do what I want. 

I feel slightly bad about this, but only slightly since I couldn't find any other path forward. I also researched the possibility of cloning/copying a template buildspec from one project to another. However, everything I found also indicates this isn't possible. The only work around I found mentioned for this was to copy project XML content from one project to another. Creating a programmatic solution around that seemed significantly worse to me than using a dummy extension class.

Source/Generator/_Editor/NI_AB_API_EXE_Extensions/Set Window Run-Time Position.vi
pbirkhol-ni: A few things:
- The error wire for ItemID isn't connected.
- The False case has a TODO about throwing an error. I think what you have is fine. For one, I don't know how this could happen if the VI got added to the build spec. So in theory, you would be getting an upstream error and it wouldn't hit the False case anyway. Furthermore, even if the VI got added to the build spec and we failed to set the run time position, so what? You might have a flicker when opening the app and we can help you work through that when/if it happens. I don't know that it should be a fatal error. All that being said, if you want to add an error I wouldn't oppose it.

pbirkhol-ni:
Thanks for figuring this stuff out.

pbirkhol-ni (approved):

Source/Generator/_Editor/NI_AB_API_EXE_Extensions/Set Window Run-Time Position.vi
jasonmreding: - Fixed the error wire chaining for the ItemID property.
- I updated the error case to merge in a custom error code/message indicating the source item doesn't exist. This should only be an internal error and not something that should make it into the wild. It would only happen if you call this VI before making the calls to add the source item to build spec to begin with.

PR #269, opened 2023-09-15
Wrap the MeasurementContext private data in a DVR

pbirkhol-ni:

[AB#2519684](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2519684)

For some upcoming features, it will be possible for measurements to indirectly modify the internal data of the MeasurementContext class. As a by-value class, this would cause inconsistent state when the wire is branched. Since the class will be storing data that might have to be be disposed, it makes sense that the class should be by-reference so that the state stays consistent.

- Replace the private data of the MeasurementContext with a single DVR reference to `Data.ctl`. `Data.ctl` is a typedefed cluster that stores the actual private data for MeasurementContext.
- Modify `Create.vi` to create the DVR.
- Add a `Destroy.vi` to destroy the DVR.
- Modify all of the class methods to get the internal data via the DVR. In other words, fix the Run arrow that broke as a result of adding the DVR.
- Call `Destroy.vi` in `Invoke Measure Logic.vi`. It is called after the measurement has completed execution. I also made a change to retrieve the MeasurementContext output of the measurement, which we were previously ignoring.
- Refactor `Invoke Measure Logic.vi` so that it isn't so messy. The primary change was to create a subVI called `Run Measure Logic and Get Results.vi` that sets the inputs to the measurement, runs the measurements, and retrieves the results. The rest of the cleanup was straightening wires, etc.
- Change the MeasurementContext class wire to be a teal refnum wire. This is recommended in many style guides for by-ref classes (including many NI internal style guides).

Prep work for future changes.

This change does not change any behavior. I ran a handful of examples that exercise all entry points of the MeasurementContext class. Note that I did not write unit tests because I didn't find them to be valuable. As is, all public methods of the MeasurementContext class are either data accessors or wrap a single VI call.


Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Invoke Measure Logic.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Run Measure Logic and Get Results.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Is Canceled.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Read Pin Map Context.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Destroy.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Update Results.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Run Measure Logic and Get Results.vi
pbirkhol-ni: This is just a refactor. I pulled out all of the logic for setting inputs, running the VI, and getting the outputs into a subVI.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Is Canceled.vi
pbirkhol-ni: No functional changes. Just accessing the class data using a DVR instead of accessing it directly.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Accessor/Read Pin Map Context.vi
pbirkhol-ni: No functional changes. Just accessing the class data using a DVR instead of accessing it directly.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Update Results.vi
pbirkhol-ni: No functional changes. Just accessing the class data using a DVR instead of accessing it directly.

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Invoke Measure Logic.vi
jonathanmendez, line 1: There's one small subtle change here and I don't know if it's important.

Previously, an error from the measurement logic VI would not get passed to the Results "Pack for GRPC" VI (the wire from Variant to Data passed around it). Now that error wire is getting combined with the standard flow error wire inside the new "Run Measure Logic and Get Results" VI and thus is getting passed to the Results "Pack for GRPC" call.

pbirkhol-ni: Good catch. And yes, it does matter. We could make it not matter by modifying the Pack for GRPC VI, but that is a VI that we generate and that the customer owns, so we can't really assume anything about it.

pbirkhol-ni: So the error was reported correctly with my change, but the measurement output never got sent to the client in the case of an error. Prior to my change, the output was sent to the client, even in an error case. I restored the original behavior and will update the screenshots.

jasonmreding (approved):

PR #271, opened 2023-09-15
Create service location pool

jonathanmendez:

[AB#2499304](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2499304)

This adds a service location pool/cache so that discovery service does not have to get repeatedly queried for every measurement call's connection to grpc-device. The upcoming session management features can utilize the pool for simplicity and performance.

This PR adds a LV2-style global (FGV) that holds a map from `Resolve Service Request` to `Service Location`. On a cache miss, it uses Discovery Service to resolve the service. On a cache hit, it returns the cached location. The cache can be cleared, so that if an error occurs we can avoid stale data.

Prep work for future changes.

Added unit tests for the cache behavior.
Also tested corner cases with manual inspection (since testing the internals of a cache is questionable with black-box testing).


Source/Tests/Tests.Runtime/Measurement Server/MeasurementServerTests.lvlib
```diff
@@ -1,17 +1,22 @@
...
+</Property>
+	<Item Name="Helpers" Type="Folder">
+		<Item Name="Check If Discovery Service is Running.vi" Type="VI" URL="../Helpers/Check If Discovery Service is Running.vi"/>
+		<Item Name="Get Enum Getters.vi" Type="VI" URL="../Helpers/Get Enum Getters.vi"/>
+		<Item Name="Kill Discovery Service.vi" Type="VI" URL="../Helpers/Kill Discovery Service.vi"/>
+	</Item>
+	<Item Name="ClientPool_CreateDestroyAndRecreateClients_ClientsCreatedWithNonZeroGrpcId.vi" Type="VI" URL="../ClientPool_CreateDestroyAndRecreateClients_ClientsCreatedWithNonZeroGrpcId.vi"/>
+	<Item Name="ClientPool_GetClients_ClientsCreatedWithNonZeroGrpcId.vi" Type="VI" URL="../ClientPool_GetClients_ClientsCreatedWithNonZeroGrpcId.vi"/>
+	<Item Name="DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi" Type="VI" URL="../DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi"/>
+	<Item Name="ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi" Type="VI" URL="../ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi"/>
```
jonathanmendez, line 18: I need to add these to Tests.lst

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
jonathanmendez, line 1: image
image
image

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi
jonathanmendez, line 1: image

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_GetInvalidServiceTwice_DoesNotCacheInvalidData.vi
jonathanmendez, line 1: image

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_GetLocationTwice_ReturnsCachedLocation.vi
jonathanmendez, line 1: image

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_GetLocation_ReturnsValidLocation.vi
jonathanmendez, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
pbirkhol-ni, line 1: This is not where I would have expected this VI to be on disc. Given our current organization where the project hierarchy roughly matches the disc hierarchy, I'd put it here: C:\dev\measurementlink-labview\Source\Runtime\MeasurementLink Measurement Server\Helpers

pbirkhol-ni: Remove the Default case for the Enum case structure. That could cause subtle bugs if/when we add a new element to the enum. Default cases are required for some data types, but not for enums.

pbirkhol-ni: I find it a little confusing that the case structure for Clear Cache doesn't actually clear the cache. Instead, the next time the FGV gets called, the cache gets cleared. My preference would be to Remove the boolean shift register completely and just have the Clear Cache case create an empty map constant. You would still need the first call primitive to create and map initially. 

This looks like it was modeled after the client pool FGV, which is different enough from the Client Pool FGV that I don't think we don't have to match the architecture exactly.

pbirkhol-ni: There is a Map Get / Replace Value inplace elment structure node that can do the checking and adding/replacing with more concise syntax. However, I don't think it would help much in our case since we have to perform an expensive operation to generate a location and don't want to do that if there is already an entry. It would be more beneficial when it is free to create a new element for the map and simply discard it if the key already exists.

All that being said, I still like the inplace element structure syntax and it can give you some assurances that you're not making copies. For example, you branch the map wire before calling `Look in Map`. Does that create a copy? Probably not, but the inplace element structure would offer a stronger guarantee. 

Ultimately, I'll leave this up to you. What you have is fine, I'm mostly just rambling.  :D

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi
pbirkhol-ni, line 1: On the block diagram, can you put all of the controls on the left (which is already true, I think) and all the indicators on the far right? It's too easy to miss indicator/controls in the middle of the diagram. This comment hold true for all VIs.

pbirkhol-ni: Should we be restarting the DiscoveryService in tests that kill it? That probably depends on what we consider the default state of the test system. Should tests assume the DiscoveryService is running?

pbirkhol-ni:
This is just me thinking out loud here. Right now, the service location pool has a dependency on the client pool. But that dependency could be reversed where the client pool calls the service location pool when it is resolving the discovery, session manager, or pinmap services. If we did that, the service location pool would want to cache its own copy of the discovery client. I don't know if any of this matters as long as the FGV's are community scoped. If we opened them up to customers, reversing the dependency could eliminate a few extra grpc calls in some cases, but is it worth it?

Anyway, no action item. Just talking.

pbirkhol-ni (changes requested):

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
jonathanmendez: > This is not where I would have expected this VI to be on disc. Given our current organization where the project hierarchy roughly matches the disc hierarchy, I'd put it here: C:\dev\measurementlink-labview\Source\Runtime\MeasurementLink Measurement Server\Helpers

Look at me thinking that moving things in the default view of a project would have some effect on disk 🤦‍♂️
Fixed.

jonathanmendez: Implemented the remaining suggestions here as well. I'll update the screenshots.

jasonmreding:
> This is just me thinking out loud here. Right now, the service location pool has a dependency on the client pool. But that dependency could be reversed where the client pool calls the service location pool when it is resolving the discovery, session manager, or pinmap services. If we did that, the service location pool would want to cache its own copy of the discovery client. I don't know if any of this matters as long as the FGV's are community scoped. If we opened them up to customers, reversing the dependency could eliminate a few extra grpc calls in some cases, but is it worth it?

I have similar questions about packaging/componentization. I think it's likely we'll want to extract the channel pool and similar functionality into its own top level library that can be referenced by multiple dependent libraries. Otherwise, I think we're going to run into issues with circular dependencies. We can leave all of this community scope for now if we're not comfortable making things public now.

Anyway, I won't have time to look at this in detail until after the weekend. I'll try to reach out after that and share some other thoughts on packaging/componentization I think will be important for this feature as well.

bkeryan:
> This adds a service location pool/cache so that discovery service does not have to get repeatedly queried for every measurement call's connection to grpc-device. The upcoming session management features can utilize the pool for simplicity and performance.

Should we do something like this for Python?

I'm currently planning to cache the SessionManagementClient for the lifetime of the service, which will have the same benefit in this specific case.

jasonmreding:
> > This adds a service location pool/cache so that discovery service does not have to get repeatedly queried for every measurement call's connection to grpc-device. The upcoming session management features can utilize the pool for simplicity and performance.
> 
> Should we do something like this for Python?
> 
> I'm currently planning to cache the SessionManagementClient for the lifetime of the service, which will have the same benefit in this specific case.

I wouldn't expect this to have as much or any value for Python as long as we already have a channel pool. I believe the Python APIs require that the gRPC channel be passed as a parameter so caching the channel should be good enough. Unfortunately, the LV driver APIs don't have the same flexibility and always require the address be passed in. Presumably this means they are always creating a new gRPC channel internally. So the best we can do is cache the address lookup so we don't have to incur the lookup cost every time.

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
jasonmreding: I would create a typedef for the Action control. I think it's unlikely to change, but if we ever want to rename or add things in the typedef, we'll be covered.

jasonmreding:
> I have similar questions about packaging/componentization. I think it's likely we'll want to extract the channel pool and similar functionality into its own top level library that can be referenced by multiple dependent libraries. Otherwise, I think we're going to run into issues with circular dependencies. We can leave all of this community scope for now if we're not comfortable making things public now.

I posted a document on packaging / componentization of the LV code you can review. Ultimately, I think this code will either need to move to our session management library at community scope or to a top level gRPC utilities library at public scope (although it could be community scope initially until we're comfortable enough with the API to make it public). I don't feel strongly about whether that refactor happens later after submitting this change, or whether that initial work should be done as part of this change. As long as we don't release with anything that is public that we don't want to be public long term, I'm happy to let you decide how you want to stage the work.

jasonmreding (approved):

dixonjoel (approved):

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi
jonathanmendez: > On the block diagram, can you put all of the controls on the left (which is already true, I think) and all the indicators on the far right? It's too easy to miss indicator/controls in the middle of the diagram. This comment hold true for all VIs.

Done

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
pbirkhol-ni: When making the typedef, make sure to replace all instances of the enum with the typedef. It's already propagated many times, mostly in the tests.

Source/Tests/Tests.Runtime/Measurement Server/ServiceLocationPool_DiscoveryServiceUnavailable_GetUncachedLocation_ReturnsError.vi
jonathanmendez: > Should we be restarting the DiscoveryService in tests that kill it? That probably depends on what we consider the default state of the test system. Should tests assume the DiscoveryService is running?

I think this is a good general question. I don't know if there's any concept of test setup/teardown that makes sense for us, but for the moment I added code to the two tests that kill discovery service to restart it at the end.

pbirkhol-ni: I was just about to mention starting the DiscoveryService even in an error case, but you beat me to it.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Classes/GrpcClientPool/Service Location FGV.vi
jonathanmendez: Thanks for the heads-up, @pbirkhol-ni .
Done.

jonathanmendez:
> > I have similar questions about packaging/componentization. I think it's likely we'll want to extract the channel pool and similar functionality into its own top level library that can be referenced by multiple dependent libraries. Otherwise, I think we're going to run into issues with circular dependencies. We can leave all of this community scope for now if we're not comfortable making things public now.
> 
> I posted a document on packaging / componentization of the LV code you can review. Ultimately, I think this code will either need to move to our session management library at community scope or to a top level gRPC utilities library at public scope (although it could be community scope initially until we're comfortable enough with the API to make it public). I don't feel strongly about whether that refactor happens later after submitting this change, or whether that initial work should be done as part of this change. As long as we don't release with anything that is public that we don't want to be public long term, I'm happy to let you decide how you want to stage the work.

Given that I think a lot of the shape of this will change in our upcoming work, I'm going to package it as-is for now for staging. I'll go take a look at the document.

PR #273, opened 2023-09-19
Create the MeasurementLink Instrument:Instrument interface

pbirkhol-ni:

[AB#2519927](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2519927)

- Add a `MeasurementLink Instrument.lvlib` library that contains the `Instrument.lvclass` interface.
- Add the 4 interface methods: `Initialize MeasurementLink Session.vi`, `Close MeasurementLink Session.vi`, `Read Instrument Type ID.vi`, and `Read Provided Interface.vi`. All of these are required overrides except for `Read Provided Interface.vi`, which will return a blank string by default.
- Add a `Remote Connection Options.ctl`. This typedef matches the gRPC device options cluster that the MI drivers use, but is named such that it could be generically used to connect to any remote server.

The `Instrument` interface is a key building block for supporting other hardware.

None.


Source/Runtime/MeasurementLinkInstrument/Instrument/Initialize MeasurementLink Session.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLinkInstrument/Instrument/Get gRPC Device Interface.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLinkInstrument/Instrument/gRPC Device Options.ctl
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
dixonjoel: How do we know if it will close or detach? Do we need to explain it more?

dixonjoel, line 1: The other library has a space on disk. 'Source\Runtime\MeasurementLink Measurement Server'. Should this be 'Source\Runtime\MeasurementLink Instrument'?

Source/Runtime/MeasurementLinkInstrument/MeasurementLink Instrument.lvlib
jasonmreding, line 1: For consistency, the containing folder should have a space in it. i.e. "MeasurementLink Instrument" rather than "MeasurementLinkInstrument".

dixonjoel (approved):

Source/Runtime/MeasurementLinkInstrument/MeasurementLink Instrument.lvlib
dixonjoel: Beat you to it :)

Source/Runtime/MeasurementLinkInstrument/Instrument/gRPC Device Options.ctl
dixonjoel: The types match the typedef in the drivers? Especially long vs ulong. I know we had some differences in different drivers on that front at one point early in the MI drivers development.

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
jasonmreding: Should we include `instrument out` on the connector pane? I guess it probably doesn't matter that much since we don't expect users to call this interface method directly.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get Instrument Type ID.vi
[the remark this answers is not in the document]
jasonmreding: If these are intended to represent simple "getter" / "setter" methods on a class, the LV style guide says we should use the terms "Read" / "Write" instead.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get gRPC Device Interface.vi
jasonmreding: If these are intended to represent simple "getter" / "setter" methods on a class, the LV style guide says we should use the terms "Read" / "Write" instead.

Source/Runtime/MeasurementLinkInstrument/Instrument/gRPC Device Options.ctl
jasonmreding: I wonder if we should call this `Session Options.ctl` instead if we want to try and minimize direct coupling to `grpc-device`? This might be a little friendlier if we ever want to use this interface as part of a HAL that isn't based on `grpc-device` or as a way to add some measure of pin mapping support on top of an existing native driver. The latter is effectively what we are doing with our drivers rather then building native pin map support into every driver.

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
pbirkhol-ni: > How do we know if it will close or detach? Do we need to explain it more?

Internally, we will know whether to close ou detach because the session manager will give us that information. We will then just pass that information to `Close Measurement Session.vi`, which is free to do something with that information or not. I assume not every plugin will have to handle close vs detach. Would you rather this match the session manager verbiage and call it `session exists`?

pbirkhol-ni: > Should we include instrument out on the connector pane?

No, we shouldn't. This is an oversight. I will fix it.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get Instrument Type ID.vi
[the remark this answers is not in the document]
pbirkhol-ni: > I think "When you add an instrument to the pin map file that MeasurementLink does not natively support..." reads better.

I stole this from TSM. Are you saying you know better than a tech writer?  :D   But I agree with you.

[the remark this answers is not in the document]
pbirkhol-ni: > If these are intended to represent simple "getter" / "setter" methods on a class, the LV style guide says we should use the terms "Read" / "Write" instead.

This is not designed to be a setter or a getter of internal class data, although it could be. I can't control how it is implemented by a client. Below are two images: the first is how I would implement it, which would not be a traditional getter. The second is how it _could_ be implemented, which would be a traditional getter. 

If you think it should be called `Read Instrument Type Id` then I can make that change. It seems within the spirit of the style guide to call it `Read`.

image
image

Source/Runtime/MeasurementLinkInstrument/Instrument/gRPC Device Options.ctl
pbirkhol-ni: > The types match the typedef in the drivers? Especially long vs ulong. I know we had some differences in different drivers on that front at one point early in the MI drivers development.

Nope. I was using an I32. I will push a change.

pbirkhol-ni: > I wonder if we should call this Session Options.ctl instead if we want to try and minimize direct coupling to grpc-device?

I am skeptical of doing this. Clients (including us) will be wiring this directly to driver APIs, so we can't just add things to this typedef in a way that is backward compatible. We could consider adding a different input that is reserved for future use. It could be a typedef cluster, a string, or something else.

It'd be nice to make this interface generic, but it's really not generic. It is highly coupled to information from session manager and/or grpc-device. We already have a mechanism for you to pass arbitrary data to Initialize (via your plugin's internal data). What other data would we have in our framework that Initialize might need?

Source/Runtime/MeasurementLinkInstrument/MeasurementLink Instrument.lvlib
pbirkhol-ni: Closing this issue. I'll fix the one Joel brought up.

Source/Runtime/MeasurementLinkInstrument/Instrument/gRPC Device Options.ctl
jasonmreding: > The types match the typedef in the drivers? Especially long vs ulong. I know we had some differences in different drivers on that front at one point early in the MI drivers development.

If different drivers aren't consistent with their use of unsigned vs signed integer for the port, then LV will allow coercion of the type even when contained in a cluster. We're going to end up with a coercion dot anyway since the typedef here won't be the same type defined by individual drivers. 

The DCPower API (the only one I have installed with grpc-device support) uses an I32 which then gets coerced to a U32 when fed to the Call Library Node on the block diagram. The TCP primitives use a U16 data type when specifying port information. Given these inconsistencies, I would probably lean towards using a U16, but I don't feel strongly about it.

pbirkhol-ni: > The DCPower API (the only one I have installed with grpc-device support) uses an I32 which then gets coerced to a U32 when fed to the Call Library Node on the block diagram.

You must have an old version of the DCPower API installed. I believe this was remedied in future versions. All of the MI drivers that I have installed use a U32 so I changed our typedef to also be a U32.

jasonmreding: > I am skeptical of doing this. Clients (including us) will be wiring this directly to driver APIs, so we can't just add things to this typedef in a way that is backward compatible. We could consider adding a different input that is reserved for future use. It could be a typedef cluster, a string, or something else.

I wasn't suggesting that we would ever add to or remove contents of the typedef over time as it would break the primary use case of interfacing with grpc-device. I was simply suggesting we change the name of the ctl so it's not strictly tied to grpc-device. In the end, the cluster is just capturing information about how to connect to a remote driver. That could apply to multiple driver APIs and not just grpc-device. We could also consider calling it something like `Connection Options.ctl` or `Remote Connection Options.ctl`. If the instrument doesn't support remoting, it would simply ignore these inputs. If the driver APIs ever support SSL connections, we'll likely end up modifying this typedef or creating a new typedef cluster input to Create to accommodate it.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get gRPC Device Interface.vi
jasonmreding: I think we should try to use discovery service terminology here and call this "provided interface" or "grpc provided interface" instead. Again, that makes this a little more general purpose and not as tightly coupled to `grpc-device`.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get Instrument Type ID.vi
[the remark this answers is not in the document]
jasonmreding: > If you think it should be called Read Instrument Type Id then I can make that change. It seems within the spirit of the style guide to call it Read.

If there were no style guide, my preference would be to use Get/Set terminology. I'm just trying to follow the style guide or the spirit of the style guide since it's not always 100% clear. If implementers of the interface follow the style guide, then they would end up with a mix of Read/Write and Get/Set methods. Maybe that's fine, but I suspect it might be somewhat annoying. 

However, you're right that for interfaces you have no control over how it's implemented, and the style guide doesn't specifically address this point. It's also not clear if this naming style was only intended to convey support of the property node for reading/writing the class data and shouldn't be used otherwise even if the getter/setter VIs simply provide access to the underlying data.

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
jasonmreding: > Would you rather this match the session manager verbiage and call it session exists?

I don't think we should do that. If we do anything, I would suggest trying to tie it more closely to the verbiage used during creation of the session. Maybe something like `created session?`. Having an input called `close session` to a `Close` method does look a little ridiculous at first. If you didn't want to close the session then why are you calling close?

I thought I remembered seeing something about using question marks in parameter names in the past (as in don't do it), but I can't find any reference to the usage of this style now. I think this is a somewhat contentious style argument that is similar to naming Boolean parameters that start with the word "Is". I would lean towards not using "?" with Boolean inputs, but I don't feel strongly about it. I know there is precedence in other APIs where it is used.

jasonmreding: Table 3 of this [document](https://www.ni.com/docs/en-US/bundle/labview-nxg-g-programming-api-overview/page/icons-connector-panes-api.html) indicates you should use "?" over "Is" for Boolean parameters that ask a question, and you should omit the "?" if the parameter is a command. I'm really not sure how to interpret that in this case. I guess I would still lean towards a command without a "?" since the user is telling us what to do.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get Instrument Type ID.vi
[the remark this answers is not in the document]
jasonmreding: > I think "When you add an instrument to the pin map file that MeasurementLink does not natively support..." reads better.

I find this to be a bit misleading or inaccurate. MeasurementLink doesn't own support for anything in the pin map. It's really the pin map editor which natively supports certain instrument types and not others. I think we could just say something like:

"Returns the type ID for the instrument in the pin map file. When using third party instruments not natively supported by the pin map editor, you must create a custom instrument in the pin map file and assign it a unique instrument type ID. Return the same instrument type ID from this method to access the third party instrument from within MeasurementLink."

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
pbirkhol-ni: What if we made an enum input instead? It could be called "close action" (and there would be an equivalent "initialize action" for Initialize). The options would be "close session" and "detach from session". I feel like that would help with naming ambiguity and make it clearer what options are available. It would also allow us to add actions in the future without breaking the interface (or at least I don't think it would break the interface).

Source/Runtime/MeasurementLinkInstrument/Instrument/Get gRPC Device Interface.vi
pbirkhol-ni: Updated the name to `Read Provided Interface.vi`. Also updated the indicator names and help to no longer mention grpc-device.

Source/Runtime/MeasurementLinkInstrument/Instrument/Get Instrument Type ID.vi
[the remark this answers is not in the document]
pbirkhol-ni: Updated the name to be `Read Instrument Type ID.vi` and also updated the help as suggested.

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
jasonmreding: I was thinking something similar. This would match [SessionInitializationBehavior](https://github.com/ni/grpc-device/blob/main/imports/protobuf/session.proto#L10) from grpc-device.

Source/Runtime/MeasurementLinkInstrument/Instrument/Initialize MeasurementLink Session.vi
jasonmreding: Updated documentation:

image

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
jasonmreding: Here is the updated API / documentation:

image

image

image

image

I defined the enums so that the default value is attach/detach rather than create/close since I think that is the primary use case from a measurement service perspective. However, I can also see the argument that it should be a required input so it doesn't matter. Let me know if you have a preference for this.

pbirkhol-ni: In my prototype I had every input to Initialize and Close as required. The reason I did this is because we should be the only callers of these VIs and we will always want to wire them. I eventually decided against it for the final implementation, but I still think my preference is to have everything required.

pbirkhol-ni:
I have been cleaning up the front panel by using "ctrl u" before submitting any VIs. That will lay out the front panel in a reasonable way and also save it in the upper left corner. I noticed in the VIs you touched that you didn't do that.

pbirkhol-ni:
I can't approve my own PR, so consider this comment my approval.

Source/Runtime/MeasurementLinkInstrument/Instrument/Close MeasurementLink Session.vi
jasonmreding: OK, made both inputs required and updated the enum items so they are shorter and more symmetrical. I'm resolving this thread.

image

image

image

image

jasonmreding (approved):

PR #287, opened 2023-09-22
Add gRPC Utilities library

jonathanmendez:

Move the Client Pool and Service Location FGV to a new "gRPC Utilities" library.

We have realized via other work that creating a utility library will be necessary to avoid circular dependencies among libraries.

Opened affected caller VIs to ensure they are not broken.
PR tests.


Source/Runtime/gRPC Utilities/Service Location FGV.vi
jonathanmendez, line 1: New icon header:
image

Source/Runtime/gRPC Utilities/TypeDefs/Service Location FGV Action.ctl
jonathanmendez, line 1: New icon header:
image

Source/Runtime/gRPC Utilities/gRPC Utilities.lvlib
jonathanmendez, line 1: To match the current behavior (and avoid presenting a new public API until we're ready), I've kept the VIs in this library as Community scope for now. The Service Location FGV doesn't have usages yet.

jonathanmendez:
All other file changes are the result of recompiling after the moves.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Location to Insecure Address.vi
pbirkhol-ni, line 1: As far as I can tell, this is only used by the GrpcClientPool class. I think it would make sense to move it to that class until something else requires it. You can also make it private if you move it to that class.

Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
pbirkhol-ni, line 1: What do you think of the library being called `MeasurementLink gRPC Utilities.lvlib`? That would make it consistent with our other libraries and less likely to clash with a customer's library. If you make this change, make sure to also rename the directory.

pbirkhol-ni (approved):

Source/Runtime/gRPC Utilities/GrpcClientPool/GrpcClientPool.lvclass
jasonmreding, line 1: Let's try to stick with book title casing here for consistency. Also, use gRPC vs. Grpc for consistency. So this name would become gRPC Client Pool.lvclass. The containing folder name would be updated to match.

jasonmreding: Friend management in LV is weird. I would have expected you could mark the entire library/class as community scope instead of individual items within the class/library. Kind of like internal in C#. However, that seems to push the friend specification up to the owning library of the library/class rather than applying it to this library. Given that, I think what you have is probably about as good as we can make it, but I haven't experimented that much with it.

jasonmreding (approved):

Source/Runtime/MeasurementLink Measurement Server/MeasurementLink Measurement Server.lvlib
jasonmreding: I think that's reasonable given the pool is currently specific to MeasurementLink services. If it were more generic, then I might argue for `NI gRPC Utilities` instead.

pbirkhol-ni: I just realized that I put this comment on the wrong library. But I think you understand what I meant.

PR #288, opened 2023-09-25
Implement IInstrument for NI-DCPower

pbirkhol-ni:

[AB#2519941](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2519941)

- Creates a implementation of `IInstrument` for NI-DCPower.
- Tests for the 4 combinations of create/attach/detach/close

Automated tests


Source/Runtime/MeasurementLink niDCPower/Instrument/Close MeasurementLink Session.vi
jonathanmendez, line 1: nit: (for this and all added VIs) Style guide says to use simply "error in" and to not specify the default value " (no error)" for the terminals.

jonathanmendez (approved):

Source/Runtime/MeasurementLink niDCPower/Instrument/Close MeasurementLink Session.vi
pbirkhol-ni: Fixed in latest push.

synesthetize (approved):

dixonjoel:
> Since this PR has a lot of VIs and taking screenshots would be tedious, I am going to try and upload an html report that shows the changes. I don't know how well this will work, but it's worth a shot. [DCPower PR.zip](https://github.com/ni/measurementlink-labview/files/12718914/DCPower.PR.zip)

Cool. How did you generate this?

pbirkhol-ni:
> > Since this PR has a lot of VIs and taking screenshots would be tedious, I am going to try and upload an html report that shows the changes. I don't know how well this will work, but it's worth a shot. [DCPower PR.zip](https://github.com/ni/measurementlink-labview/files/12718914/DCPower.PR.zip)
> 
> Cool. How did you generate this?

If you have a VI open, you can go to `File >> Print`. Then you have to go through a bunch of dialogs, select your VIs, etc. It's not easy, but it's better than taking a bunch of screenshots.

dixonjoel (approved):
Approved with suggestions.

Source/Runtime/Measurement Server.lvproj
jasonmreding, line 1: One thing I've done in my prototype branch as I've been experimenting with driver specific API changes is to add an "Instruments" folder under the "Libraries" folder in the project. I then placed the IInstrument library and all of the driver specific libraries under it. This is just a virtual folder in the project and not something for which I changed the disk layout, although I'm not necessarily opposed to that. I've found this change to be helpful in segregating instrument specific code from the rest of the service related code and navigating the project in general. If you agree, then I recommend putting this change in place before work on all of the other instrument specific libraries begins.

Source/Runtime/MeasurementLink niDCPower/Instrument/Close MeasurementLink Session.vi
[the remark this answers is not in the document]
jasonmreding: I'm in favor of stealing the banner for the library and perhaps the icons as well. However, I would expect a second round of changes polishing the icons once we have final iconography decided upon. Either way, this class isn't ultimately what would get this icon treatment. It's the owning library of this class which will be presented as the public API which should get this treatment. This class is just an implementation detail. As such, I think text icons are fine.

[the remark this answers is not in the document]
jasonmreding: The way interfaces work in LV is they are basically classes without a private data control. As such, there is no way to provide a common implementation in the interface that requires the use of private data. This is similar to default interfaces in C#. So we can either by convention require all implementations to store their session in the private data of their IInstrument implementation and not include the session on the conn pane, or we can track the session for them in the `SessionReservation` class as a variant and then pass it to them on the conn pane. I don't feel that strongly about this one way or the other, but this is where we landed initially. I don't know that one approach vs. the other is substantially more intuitive for an implementer than the other.

Source/Tests/Tests.Runtime/Measurement Server/niDCPower/RemoteDCPowerSession_CallCloseMeasurementLinkSessionWithDetachBehavior_RemoteSessionDetached.vi
jasonmreding, line 1: The error chaining at the end when closing sessions was bit confusing for me at first. It appears that you can close an already closed session which I would expect to generate an error. This error then gets chained through a couple of other VIs which basically ignore error in so it should be harmless. However, the standard error data flow makes it easy to jump to the opposite conclusion. I don't necessarily have a great suggestion for improving this. I'm mostly mentioning it to see if it sparks any ideas in you. You could technically delete the error wire for data flow control and execute all three cleanup VI in parallel. However, I don't know if that really helps readability any.

Source/Tests/Tests.Runtime/Measurement Server/niDCPower/DCPowerInstrument_CallInitializeMeasurementLinkSessionWithCreateBehavior_RemoteSessionCreated.vi
jasonmreding, line 1: You may want to eventually replace some of this code with the public APIs that will be created in MeasurementLink niDCPower.lvlib and go on the palette. Maybe that's a reason to include those changes as part of the IInstrument implementation in the future? Or maybe just add the tests with that change instead of this change.

jasonmreding, line 1: There appears to be some common code patterns duplicated in the tests for validating the error code/cluster. This might be a good opportunity to create some Assert style helper VIs since I would expect they would get a lot of reuse. I'm OK doing that sort of thing in a follow on change as well.

jasonmreding (approved):

Source/Runtime/Measurement Server.lvproj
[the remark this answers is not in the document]
jasonmreding: Paul had discussed this issue with me briefly before leaving for the day. If the only issue was receiving a warning dialog like this once each time you opened the project when you don't have all of the drivers installed, I would be inclined not to care about this. However, Paul was more worried about the LabVIEW behavior regarding interfaces that he was seeing when any of the interface implementations were broken. From my experiments, here is what I determined:
- For classes implementing the interface, any VI part of the class is broken if any interface implementation loaded into memory doesn't override all of the required methods.
- For classes implementing the interface, a broken interface method for one class implementation doesn't impact run status of other classes implementing the same interface (assuming all implementation override the required methods as described above).
- For VIs calling an interface method, the caller is broken if any interface implementations of that method loaded into memory are broken. I verified this by both adding a missing sub VI dependency and just breaking the diagram logic.

This behavior is definitely not what I would have expected and I would argue that it is a bug. However, given the current behavior, any customer wanting to write their own plugin would likely need to have all of the drivers installed in order for their own plugin implementation to be runnable as they are developing it. This definitely isn't ideal. 

As a work around, Paul mentioned possibly making everything a malleable VI so the VI wouldn't break at edit time if the driver wasn't installed and instead generate a runtime error. I think this is worth considering. I also think we might be able to use a conditional compilation structure and then "document" this conditional symbol so that plugin authors can use it as a work around. The project we use for development would also define this symbol so that we can work around this behavior at edit time but still get edit time error checking when running from vilib. We would need to test this though as I know LV has sometimes had unexpected behavior in regards to dependency management for things in disabled cases of a conditional compilation structure.

Source/Runtime/MeasurementLink niDCPower/Instrument/Close MeasurementLink Session.vi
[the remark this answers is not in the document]
pbirkhol-ni: I agree that icons need work. All public VIs should get proper icon treatment later. For private VIs, we've just been using the color and style that LabVIEW gives us by default.

Source/Runtime/Measurement Server.lvproj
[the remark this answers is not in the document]
pbirkhol-ni: > However, given the current behavior, any customer wanting to write their own plugin would likely need to have all of the drivers installed in order for their own plugin implementation to be runnable as they are developing it

I don't think this is true. There would be no reason for our `IInstrument` implementations to be in memory when a customer is developing a plugin. And if some were in memory, they would likely be in memory because the customer has the driver installed and is using them. So I don't see this as a problem for plugin development, even if we don't use the Type Specialization structure. It's a problem for our internal development because we have all of the `IInstrument` implementations in the same project and therefore they are all in memory when you open that project.

> As a work around, Paul mentioned possibly making everything a malleable VI so the VI wouldn't break at edit time if the driver wasn't installed and instead generate a runtime error.

You can use the Type Specialization structure in a normal VI. I am not recommending using a malleable VI because that comes with its own drawbacks (inlined, reentrant, no debugging).

>  I also think we might be able to use a conditional compilation structure and then "document" this conditional symbol so that plugin authors can use it as a work around.

This is something I have also considered. It's not automatic like the Type Specialization structure, but it could give you compile time errors which are better. In this case, I'm not convinced that compile time errors are _that_ much better. This isn't a corner case that you might not hit in normal testing. This would be a fundamental problem and your measurement wouldn't work at all. You would know the first time you ran your measurement. So I'm inclined to use a Type Specialization structure for now. We can revisit that decision later.

Source/Runtime/MeasurementLink niDCPower/Instrument/Close MeasurementLink Session.vi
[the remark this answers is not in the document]
pbirkhol-ni: I can see the appeal for storing it in class data. However, unless there is a compelling advantage, I'm going to leave it as is.

Source/Runtime/Measurement Server.lvproj
pbirkhol-ni: I'll make the change in the project for now. Making a change on disk requires us to change the installer, which is not a change I want to do at the moment.

dixonjoel: I like this idea - I think it's a good compromise between having a separate project and having all the instrument code and generic code mashed together.

Source/Tests/Tests.Runtime/Measurement Server/niDCPower/RemoteDCPowerSession_CallCloseMeasurementLinkSessionWithDetachBehavior_RemoteSessionDetached.vi
pbirkhol-ni: You're absolutely right. That error chaining depends on all downstream VIs to ignore incoming errors, which they do. I don't have a better suggestion, which is why I did it like that. 

> You could technically delete the error wire for data flow control and execute all three cleanup VI in parallel.

You can't execute all 3 VIs in parallel. You can probably execute the closes in parallel, but the `Restart Discovery Service` has to happen after close or else the local close could fail. I know this because I tried to do them in parallel and it led to partial cleanups. This may be a bug in the drivers.

I know it's a cop out, but I'll just comment this section of code and explain what is happening. I could also do something like put a clear errors between each call, but I don't think that is any better.

Source/Tests/Tests.Runtime/Measurement Server/niDCPower/DCPowerInstrument_CallInitializeMeasurementLinkSessionWithCreateBehavior_RemoteSessionCreated.vi
jonathanmendez: In my opinion, while there will be overlap, I prefer to keep these tests (and continue the pattern for other instruments). However, I recognize it's a bit of a gray area/philosophical question.

One angle is "unit vs integration" - to me, these tests are closer to unit tests in that they rely on straightforwardly using the `IInstrument` implementation VIs with simple inputs. A test for `Create Sessions` brings in the `Session Reservation.lvclass` as a separate entity, and in particular requires either setting up a pin map, registering it, and reserving sessions; or more likely requires setting up a testing seam to provide hardcoded Session Infos for an instance of the class. Either way, while such a test will definitely be valuable, to me it is the kind of test where "more things could go wrong" and therefore a failure would take longer to diagnose. The current tests for the `IInstrument` give a much clearer picture of what went wrong, should they fail.

Another angle is "public vs internal API" - there's an argument to be made that `Create Sessions` is the public API and the appropriate level for testing. However, the `IInstrument` interface is also going to be public enough (in that we expect customers to write plugins for 3rd party instruments) that I don't worry about the usual concerns of testing something internal (i.e. that the implementation is subject to change frequently and the tests will get in the way of change).

There's another angle of "total cost of tests" where, if each test has significant overhead, then the marginal cost of the overlap may override the other benefits. I don't have a good feel for the marginal cost of each individual test VI and whether we need to worry about this.

pbirkhol-ni: Thanks for weighing in. I will keep the tests. There is minimal runtime overhead, so I'm not concerned about that. Across all drivers, I expect these tests to add a minute or less to the PR time. There will be some overhead in implementing them for other drivers, but that is also minimal now that there is a pattern to follow.

Source/Runtime/Measurement Server.lvproj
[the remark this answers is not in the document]
pbirkhol-ni: Here is what the Type Specialization looks like:

image

Source/Tests/Tests.Runtime/Measurement Server/niDCPower/DCPowerInstrument_CallInitializeMeasurementLinkSessionWithCreateBehavior_RemoteSessionCreated.vi
pbirkhol-ni: I created two new subVIs that these tests can call as well as the other driver tests.

Source/Runtime/Measurement Server.lvproj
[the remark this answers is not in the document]
jasonmreding: The error string corresponding to error 5052 doesn't include a format specifier for the driver name in the version I pulled down. Did you forget to save your edit as the new default value for the control?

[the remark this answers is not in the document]
pbirkhol-ni: > Did you forget to save your edit as the new default value for the control?

That's exactly what happened. Fixed now.

PR #289, opened 2023-09-25
Separate compiled code from source for all VIs

pbirkhol-ni:

- Check the `Separate compiled code from source file` property for all VIs. The only exceptions are: 
    - The `Measurement UI.vi` in the examples and the template. That VI needs the compiled code so that it can be loaded by InstrumentStudio.
    - The VIs that are generated by grpc-labview. 
- Check the `Separate compiled code from new project items` for all projects. 

Ran examples. Created a new measurement. Created a built exe.


Source/Runtime/Measurement Service V1.lvproj
```diff
@@ -1,5 +1,7 @@
 ﻿<?xml version='1.0' encoding='UTF-8'?>
 <Project Type="Project" LVVersion="20008000">
+	<Property Name="NI.LV.All.SourceOnly" Type="Bool">true</Property>
```
jonathanmendez, line 3: Given this project represents generated VIs, is there a reason to check the box here?

jonathanmendez:
For other reviewers:
I loaded up each of the projects in LabVIEW, and verified:
- The "Separate compiled code from new project items" box was checked, but more importantly
- Clicking "Mark Existing Items" showed the correct VIs to be marked (all but the Measurement UI VIs)

jonathanmendez (approved):

pbirkhol-ni:
> For other reviewers: I loaded up each of the projects in LabVIEW, and verified:
> 
> * The "Separate compiled code from new project items" box was checked, but more importantly
> * Clicking "Mark Existing Items" showed the correct VIs to be marked (all but the Measurement UI VIs)

That is a lot more thorough review than I expected.  :)

Source/Runtime/Measurement Service V1.lvproj
pbirkhol-ni: The only reason I did it is to be consistent with the other projects (both the generated projects and non-generated projects). Also, if we ever add tests to these projects then they will be marked source-only be default, which is what we would want.

dixonjoel:
In general, what is the purpose of this? Is it to limit the size of our packages? And is it safe to do this since we're shipping the source code into vi.lib and LabVIEW will compile them as needed?

pbirkhol-ni:
> In general, what is the purpose of this? Is it to limit the size of our packages? And is it safe to do this since we're shipping the source code into vi.lib and LabVIEW will compile them as needed?

The purpose of this is to help minimize the amount of "phantom" changes to VIs. Prior to this PR, if you made a change to a subVI, it was highly likely that all its callers would also have a dirty dot. With this change, that will no longer be true. 

It is not only safe to do it, but it is recommended by NI. I believe everything in vi.lib that the LabVIEW team ships is source only. LabVIEW will compile the code and put it into the compiled object cache when necessary.

dixonjoel (approved):

PR #291, opened 2023-09-26
Move the Error Strings Global.vi to the Utilities library

pbirkhol-ni:

Moves the `Error Strings Global.vi` to a common library so that it can be used in other libraries without causing circular dependencies. 

None.


Source/Runtime/MeasurementLink gRPC Utilities/MeasurementLink gRPC Utilities.lvlib
pbirkhol-ni, line 1: Should we rename this to be `MeasurementLink Utilities.lvlib` since the `Error Strings Global.vi` is not a gRPC utility.

dixonjoel:
Are all of the other VI changes just because of updating based on the location change?

pbirkhol-ni:
> Are all of the other VI changes just because of updating based on the location change?

Yep. Changing libraries will update all of the callers.

Source/Runtime/MeasurementLink gRPC Utilities/MeasurementLink gRPC Utilities.lvlib
dixonjoel: I could go either way on this. Not a big deal, but slightly preferable to have a less specific name.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementPluginService/Get Plugin Paths.vi
jasonmreding, line 1: You should be able to delete the three usages from within this class and probably free up the error code for something else. The class is already configured such that these methods must be overridden, and I can't think of any reason why this could would otherwise be needed or be reachable.

jasonmreding:
I have mixed feelings about this change. Can you provide a little more detail of how you plan to take advantage of this in the future?

My concerns  are:
1. Having a global errors file that is referenced from multiple libraries seems like a bad idea from an encapsulation standpoint. Why not just put everything in or under a single top level library at that point? This isn't something we would think about in other programming languages so what's different about LV that would make this OK or necessary?
1. By Making this public within the utilities library, it seems like it would be easy to add this as a dependency to the measurement service template code at some point without realizing it. If that happens, we're locked into maintaining this global errors VI in this library forever (or until we're comfortable breaking customer code). We can make the VI community scoped to help prevent that, but then we're promoting even more things to friendship status.  Maybe that's OK, but I fear our rampant use of friends will come back to bite us at some point.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementPluginService/Get Plugin Paths.vi
pbirkhol-ni: Those errors have been helpful to me, but now I don't remember the context. I think maybe I was trying to use the `MeasurementPluginService` class directly. However, I agree that it isn't valuable for customers so I can just hardcode an error.

Source/Runtime/MeasurementLink gRPC Utilities/MeasurementLink gRPC Utilities.lvlib
pbirkhol-ni: I am now making this a loose VI, so it is no longer part of this library.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementPluginService/Get Plugin Paths.vi
pbirkhol-ni: I removed the official errors and added a simple "-1" error for internal benefit.

image

pbirkhol-ni:
> > I have mixed feelings about this change. Can you provide a little more detail of how you plan to take advantage of this in the future?
> > My concerns are:
> > 
> > 1. Having a global errors file that is referenced from multiple libraries seems like a bad idea from an encapsulation standpoint. Why not just put everything in or under a single top level library at that point? This isn't something we would think about in other programming languages so what's different about LV that would make this OK or necessary?
> > 2. By Making this public within the utilities library, it seems like it would be easy to add this as a dependency to the measurement service template code at some point without realizing it. If that happens, we're locked into maintaining this global errors VI in this library forever (or until we're comfortable breaking customer code). We can make the VI community scoped to help prevent that, but then we're promoting even more things to friendship status.  Maybe that's OK, but I fear our rampant use of friends will come back to bite us at some point.
> 
> For the Get Connections user story, there are a few cases where I will need to throw errors. I would like to use the `Error Strings Global.vi` to do that so it is easier to verify that I am not reusing an error code. I view `Error Strings Global.vi` as measurementlink-labview's error database. Yesterday we discussed other ways we can single source our error codes (like the Error Ring), and we should probably do that some day, but I don't want it to distract from our features.
> 
> Would it be better if I just made it a loose VI that wasn't part of any library? Or put it in its own library? Both of those are possible and might alleviate your concern about the template.

Jason and I discussed this offline. We are going to make the error VI a loose VI for now and create a user story to handle errors better in the future.

PR #293, opened 2023-09-28
Implement IInstrument for NI-Digital

WesleyTangNationalInstruments:

[AB#2531939](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2531939)

- Creates a implementation of IInstrument for NI-Digital.
- Tests for the 4 combinations of create/attach/detach/close

These changes are based on Paul's PR, #288 .

Close MeasurementLink Session.vi
image

Create.vi
image

Get Instrument Type ID.vi
image

Get Provided Interface.vi
image

Initialize MeasurementLink Session.vi
image

- Automated tests pass.


jasonmreding (commented):
Most of the VI descriptions for VIs in the Instrument.lvclass contain stale information from the interface VIs. I would recommend just deleting the descriptions since these VIs are really considered implementation details at this level and not public APIs. The same issue exists for the DCPower instrument class.

Source/Runtime/MeasurementLink niDigital/Instrument/Close MeasurementLink Session.vi
jasonmreding, line 1: The close wires error in to the `Variant to Data` node. I haven't tested this, but I think this will fail if error in contains an error which will then cause `Close` to fail since the refnum is invalid. This breaks typical close semantics for drivers where they always execute even if error in contains an error. If this is correct, I would recommend forking the error wire and adding a join errors node at the end. The same problem exists for the DCPower instrument, which I assume is what you modeled this after.

jasonmreding (approved):

Source/Runtime/MeasurementLink niDigital/Instrument/Close MeasurementLink Session.vi
pbirkhol-ni: We could also consider creating a test around this case.

WesleyTangNationalInstruments: I've fixed this for both Digital and DCPower.

image

WesleyTangNationalInstruments:
> Most of the VI descriptions for VIs in the Instrument.lvclass contain stale information from the interface VIs. I would recommend just deleting the descriptions since these VIs are really considered implementation details at this level and not public APIs. The same issue exists for the DCPower instrument class.

I've deleted the VI descriptions (via "Properties -> Documentation -> VI Description" for each DCPower and Digital VI) as requested.

Source/Runtime/MeasurementLink niDigital/Instrument/Close MeasurementLink Session.vi
jasonmreding: Unfortunately, the fork isn't what we want. The error in should just be wired to the merge errors at the end directly. It should also be the first input that is wired so that error in is preserved if this VI generates an error for some reason.

WesleyTangNationalInstruments: Hmm, so we don't want the `error in` to pipe into the `Variant to Data` VI at all then, right? This would ensure that `Variant to Data` does not fail due to an error from the caller. 

Here's an update. Let me know if my understanding is still incorrect.
image

jasonmreding: Yes, that is correct. Ideally the behavior when calling this VI wrapper should be indistinguishable from calling the driver Close/Detach VI directly. So if calling the driver Close succeeds in closing the session even if error in has an error, then so should this VI.

Source/Runtime/MeasurementLink niDCPower/Instrument/Get Instrument Type ID.vi
synesthetize, line 1: I'm not 100% sure but I think this is defined as "niDigitalPattern" in the PinMapFramework.

WesleyTangNationalInstruments: Good catch, I've made another PR to address this since this is already merged.

https://github.com/ni/measurementlink-labview/pull/295

PR #294, opened 2023-09-28
Implement IInstrument for NI-FGEN

synesthetize:

[AB#2533773](https://ni.visualstudio.com/DevCentral/_workitems/edit/2533773)
- Creates an implementation of `IInstrument` for NI-FGEN
- Adds tests for Create, Attach, Detach, Close

Based on #288 

zipped html of changes: [fgen impl (2).zip](https://github.com/ni/measurementlink-labview/files/12753520/fgen.imple.2.zip)

Auto tests pass


Source/Runtime/MeasurementLink niFGEN/Instrument/Close MeasurementLink Session.vi
dixonjoel, line 1: Icons. For FGen, you have a light-blue banner. I don't know if that matches the driver API icons, but for Digital, we used the same green banner as DCPower. We should decide if we want a unique color for each instrument implementation, to copy the color of the driver, or to use the same color indicating 'Instrument' but only knowing which is which by reading the banner text.

dixonjoel (changes requested):

Source/Runtime/MeasurementLink niFGEN/Instrument/Create.vi
pbirkhol-ni, line 1: image

The channel name is getting passed to Create here, which means that the customer is going to be required to provide the channel name on their diagram. However, I think this should come from the session information that we get from the session management service. We are trying to shield the customers from that raw information, so we will have to modify the IInstrument interface so that it takes a channel input so that our framework can provide the channel for the customer.

FYI @bkeryan @jonathanmendez

Source/Runtime/MeasurementLink niFGEN/Instrument/Close MeasurementLink Session.vi
dixonjoel: On second thought, this is something we can easily change later. So it shouldn't hold up your PR.

dixonjoel (approved):

Source/Runtime/MeasurementLink niFGEN/Instrument/Create.vi
jasonmreding: I've always been a little confused as to how this works in general with pin mapping. All drivers effectively have the notion of device names and channel names. However, some drivers munge this data together into a single resource string input and others have separate string inputs for the device name vs. the channel name(s)? Also, in some instances, the driver requires fully qualified channel names (because of channel expansion), and sometimes it doesn't? I think some drivers even error if you try to use a fully qualified channel name?

Given these discrepancies, where is the data normalized and whose responsibility is it to convert the normalized data to driver specific input strings? Or am I wrong and most of the drivers behave the same way so we don't have to worry about normalization? Could the MeasurementLink session creation API we present normalize everything into a single resource string that includes both the device and channel names? Driver specific implementations would then have to transform this string into driver specific string inputs if the driver doesn't support this syntax directly. Would that be cleaner than adding a channel(s) input to `IInstrument` even though most instruments will never use it? If there are multiple pins in the `SessionReservation` for that instrument session, what meaning does the channel string have?

jonathanmendez: This is an area where the pin map framework makes certain assumptions and encodes a specific behavior for each instrument type.

For example:
- A DMM only has one channel (you cannot specify otherwise in the pin map)
- A VST similarly has two hardcoded channels ("In" and "Out").
- A DCPower lets you specify groupings of channels and encodes the instrument and channel names that are grouped into a resource string

As far as I know, most of the assumptions encoded into the PMF are matched well to driver usage patterns.

However for FGEN, the PMF assumes that the by-instrument sessions will be used rather than by-channel. The fact that apparently FGEN's gRPC Initialize specifies a channel could cause us problems.

Scenario: You have one FGEN instrument "FGEN1" with channels 0 and 1 connected to pins A and B respectively
Workflow:
- Reserve, Create, and Register a session for Pin A. The resource name is "FGEN1", the initialize specifies channel 0, the session name registered with grpc-device and with session manager is "FGEN1".
- Now reserve a session for Pin B. The session manager will return a session info with `session_exists = true` so your code attaches to the session instead of initializing it. At this point you think you have a session for pin B/channel 1, but you actually have a session for pin A/channel 0. You presumably either get an error (if the driver code has you specify a channel for driver operations) or, more worrisome, you do channel-less operations thinking that you're using one channel but you're actually using another.

synesthetize: > some drivers munge this data together into a single resource string input and others have separate string inputs for the device name vs. the channel name(s)? Also, in some instances, the driver requires fully qualified channel names (because of channel expansion), and sometimes it doesn't? I think some drivers even error if you try to use a fully qualified channel name?

I'm familiar enough with DCPower, FGEN, and SCOPE driver history to provide at least some context for the discrepancies. I think initially they all probably just accepted `resourceName`. DCPower was probably the first one to support channel based sessions and that was initially implemented with a `channelName` input. SCOPE never supported channel based sessions because of hardware limitations with independently configuring the acquisition engine. FGEN eventually added devices that could support independent channel based sessions and followed DCPowers lead by using a `channelName` input (note: you can also pass in what channels you want to use via the option string instead so maybe that's a potential solution?).

When SCOPE supported channel expansion the solution for having multiple devices was to allow a comma delimited list of resouceNames. When DCPower added channel expansion you now had an issue where "what device is `channelName` actually referring to now if there can be multiple devices". The solution there is to allow a comma delimited list of fully qualified channel names in the `resourceName` arg. The original entrypoint that takes a `channelName` is now deprecated in favor of the independent channels one which accepts the fully qualified names. FGEN never supported channel expansion so it doesn't accept fully qualified names at all and will error.

I don't know enough about DMM or Digital to comment.

jasonmreding: If the pin map doesn't support the concept of an FGEN device with channel based sessions, then it seems like we shouldn't expose MeasurementLink APIs that would potentially require the use of channel based sessions. I don't know if this is problematic or not. From the LV FGEN examples, it appears the channel string can be empty, but the documentation doesn't explain what that means. I assume it means create a session for all channels on the device.

synesthetize: > From the LV FGEN examples, it appears the channel string can be empty, but the documentation doesn't explain what that means. I assume it means create a session for all channels on the device.

That is correct.

jasonmreding (commented):
All of the comments I made in #293 about VI descriptions and the Close VI apply to this review as well.

Source/Runtime/MeasurementLink niFGEN/Instrument/Create.vi
pbirkhol-ni: This wouldn't be exposed to the customer directly. I think the question is more about if should add a channels input, both for FGEN and for other drivers that might support per channel initialization. If we add a channels input, we could ignore it for FGEN. Then later if we decide to support reserving independent channels on FGEN we could start using it without breaking compatibility.

bkeryan: In Python, the session constructor callback will take a SessionInformation so that the driver-specific implementation can use whichever fields it needs.

bkeryan: Yeah, you're painting yourselves into a corner by using a resource name in the `Initialize MeasurementLink Session` conpane:
image

Replace it with a `SessionInformation` cluster so you can also access the channel string.

dixonjoel: If we're doing that, IMO that update plus retrofits to DCPower and NI-Digital instruments should go in before this PR.

pbirkhol-ni: If we used a `SessionInformation` cluster, we would no longer need to have the `resource name` or `initialize session behavior` inputs. In my first prototype many weeks ago, I was using the `SessionInformation` cluster as the only input. However, since it has a lot of extraneous parameters, I decided to only pass what (I thought) was needed. As it turns out, maybe we shouldn't assume what instruments will need.

WesleyTangNationalInstruments (changes requested):

WesleyTangNationalInstruments (approved):

Source/Runtime/MeasurementLink niFGEN/Instrument/Close MeasurementLink Session.vi
synesthetize: I'm went ahead and changed this to match the other impls just to be consistent for now. As you said it should be trivial to change later if we want.

Source/Runtime/MeasurementLink niFGEN/Instrument/Create.vi
synesthetize: Went ahead and removed `channel name` from `Create.vi` and private class data and no longer wire up a channel string to the driver init call since PMF doesn't support it. Whatever changes we need for retrofitting DCPower and Digital should line up well for FGEN now as well.

jasonmreding: > If we're doing that, IMO that update plus retrofits to DCPower and NI-Digital instruments should go in before this PR.

Or have someone do it for all drivers once the initial submit for each driver has been completed.

> As it turns out, maybe we shouldn't assume what instruments will need.

It's an interface so we have to make assumptions about what any instrument might need. Even if we move everything into a cluster, the cluster still contains the explicit fields that we define, and there is always some risk this won't be sufficient for some future driver use case. The typedef cluster just allows us to add things in the future without breaking the VI signature. 

There is always some amount of subjectivity to this sort of API design in regards to usability vs. versatility. Having to always bundle/unbundle data is kind of a pain so there is always some pressure to define the most common inputs as independent input terminals and push everything else into an "advanced" or "not commonly used" input terminal. I feel like having a single input cluster is generally going a little too far. However, since we don't expect customers to call these interface APIs directly, maybe it's OK in this case. 

> In Python, the session constructor callback will take a SessionInformation so that the driver-specific implementation can use whichever fields it needs.

I think that would be the equivalent of adding the `SessionInformation` to the `Initialize MeasurementLink Session` interface VI. Alternatively, the driver already has access to the `SessionReservation` within their `Create Sessions` wrapper VI. We could just provide APIs on `SessionReservation` that gives access to the underlying `SessionInformation`. The only difference at that point is the driver would have to filter the session info again to that which only pertains to its driver.

jasonmreding: My recommendation for now would be to proceed without the channel information. I would create an item in the backlog to figure out the appropriate interface signature before releasing and retrofitting the existing implementations as necessary.

jasonmreding (approved):

Source/Runtime/MeasurementLink niFGEN/Instrument/Create.vi
pbirkhol-ni: > We could just provide APIs on SessionReservation that gives access to the underlying SessionInformation

We already plan to do this as part of one of the features. That being said, I don't like this solution. Part of the benefit of having the interface is that plugin authors don't have to worry about parsing the array of SessionInformations. We do that for them and they only have to worry about initializing/closing their driver.

I am in favor of either passing SessionInformation directly into Initialize or adding an `additional initialization settings` input. All things being equal, I would suggest we match python.

PR #296, opened 2023-09-28
Update .proto files in LabVIEW repo

dixonjoel:

@aepete pointed out that our measurement_service.proto file is behind the original one. In addition, xydata.proto has been added. I see no reason not to update these now, even if they won't take effect until we regenerate our libraries with grpc-labview.

Another option would be to defer adding these until the same PR where we're updating the generated libraries. I could see how someone would get the impression that our libraries would match the submitted protos.

Keep our .proto definitions up-to-date.

None. I expect no effects of this for now.

### DIFF
--- a/Source/Protos/ni/measurementlink/measurement/v2/measurement_service.proto
+++ b/Source/Protos/ni/measurementlink/measurement/v2/measurement_service.proto
@@ -109,6 +109,8 @@ message ConfigurationParameter {
   int32 field_number = 1;
 
   // Required. The data type for the configuration parameter.
+  // NOTE: If this is Kind.TypeMessage, you *must* also fill in
+  // the 'message_type' field below with the gRPC full name of the message.
   google.protobuf.Field.Kind type = 2;
 
   // Required. The name of the configuration parameter. When defining a user interface for the measurement, a control
@@ -133,6 +135,10 @@ message ConfigurationParameter {
   //   - Expected format: JSON dictionary string"
   //   - Example: "{\"RED\":0, \"GREEN\":25, \"BLUE\":5}"
   map<string, string> annotations = 5;
+  
+  // Required when 'type' is Kind.TypeMessage. Ignored for any other 'type'.
+  // This is the gRPC full name of the message type.
+  string message_type = 6;
 }
 
 // Message that defines an output of the measurement.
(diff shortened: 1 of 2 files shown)


pbirkhol-ni (approved):

dixonjoel:
@jasonmreding I'd like to get your approval as well before submitting.

aepete (approved):

aepete:
>> Another option would be to defer adding these until the same PR where we're updating the generated libraries. I could see how someone would get the impression that our libraries would match the submitted protos.

Yeah... I guess I don't feel too strongly about this, but it is a risk.  Might not hurt to just re-gen the libraries at our first upcoming convenience, whenever that might be.

dixonjoel:
> > > Another option would be to defer adding these until the same PR where we're updating the generated libraries. I could see how someone would get the impression that our libraries would match the submitted protos.
> 
> Yeah... I guess I don't feel too strongly about this, but it is a risk. Might not hurt to just re-gen the libraries at our first upcoming convenience, whenever that might be.

@pbirkhol-ni has tried about 5 times, but we keep finding bugs in grpc-labview. If at first you don't succeed...

jasonmreding:
> > > Another option would be to defer adding these until the same PR where we're updating the generated libraries. I could see how someone would get the impression that our libraries would match the submitted protos.
> 
> Yeah... I guess I don't feel too strongly about this, but it is a risk. Might not hurt to just re-gen the libraries at our first upcoming convenience, whenever that might be.

Until the VIPB files are updated, it doesn't really matter whether the generated libraries using grpc-labview have been added to the repo or not. At least not unless you are wanting to build/contribute to this repo.

dixonjoel:
> > > > Another option would be to defer adding these until the same PR where we're updating the generated libraries. I could see how someone would get the impression that our libraries would match the submitted protos.
> > 
> > 
> > Yeah... I guess I don't feel too strongly about this, but it is a risk. Might not hurt to just re-gen the libraries at our first upcoming convenience, whenever that might be.
> 
> Until the VIPB files are updated, it doesn't really matter whether the generated libraries using grpc-labview have been added to the repo or not. At least not unless you are wanting to build/contribute to this repo. This is in regards to `xydata.proto`.
> 
> On the other hand, it is a little odd that `measurement_service.proto` doesn't reflect the generated code from the proto that has been added to the repo. In general, my preference would be to keep things synchronized. I'm not adamant about this if there are valid reasons to submit things incrementally. However, I would hope the amount of time where the two things don't agree in the repo is kept to a minimum and not days or weeks. Long term, ideally all of this will be automated as part of the build (with the exception perhaps of the VIPB file updates) so things should never get out of sync. Until things are automated, it would be ideal if our manual update process modeled the desired automated process.

Sure, I have no problem with that. I'll close this PR because given the problems with grpc-labview it could well be weeks or more until we get the generated files updated.

dixonjoel:
Abandoning. Proto files should be updated along with a new set of generated libraries. Or shortly before if you want a separate PR and you're pretty sure the generated libraries will closely follow.

PR #298, opened 2023-09-29
Add Create Sessions implementation with NI-DCPower vertical slice

jonathanmendez:

[AB#2520982](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2520982)

- Add the Create Sessions implementation for the Session Reservation class.
- Add the NI-DCPower implementation for the Create Sessions poly VI (1 Session and N Sessions)

[CreateSessionsPR.zip](https://github.com/ni/measurementlink-labview/files/12785771/CreateSessionsPR.zip)

With the underlying implementation in place for the Session Reservation class, this unblocks the other drivers to create their Create Sessions VIs.

TODO: Need to add unit tests.


Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
pbirkhol-ni, line 1: For Get Connections, I have a separate VIM per poly instance. In your case, that would mean you would have a `Create Session.vim` and `Create Sessions.vim`. The reason I do that is so customers who want to create their own API to match ours only have to create their plugin object and then call a VIM. The way you did it, for the 1 session case, the customer would have to duplicate our error handling as well. 

I am not necessarily saying my way is better, but I think we should discuss how we want to present the VIMs, since I believe they will be made public.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get Connection Information.vi
pbirkhol-ni, line 1: This VI name is too close to `Get Connections`. Can you instead call it something like `Get gRPC Device Server Location`?

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
pbirkhol-ni, line 1: We should try to come up with a better name than `session created`. I was reading the code thinking that `session created` was the same as `session exists` and was wondering what you were doing.  :D   I wonder if we can tell if a session has been created or not by comparing the session variant to an empty variant? Then we wouldn't need that boolean at all.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
pbirkhol-ni: Another reason in favor of multiple VIMs is that we can do almost all of the tests on the VIMs instead of the driver specific APIs. That should cut down on a lot of test duplication.

jonathanmendez: I think the design with multiple VIMs makes sense for the reasons you gave. I'll update for that.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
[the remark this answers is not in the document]
jonathanmendez: Done.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
jonathanmendez: Updated.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get Connection Information.vi
jonathanmendez: Done.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Session and Driver Info.ctl
jasonmreding, line 1: In some of my prototyping, I split this into two typedefs. I called one `Connection.ctl` and the other `Connection with Generic Session.ctl`. The first typedef just contains the connection information from the session manager, and the second bundles that information with a variant for the driver session data type. Having this split out will be useful later when adding the `Get Connections` VIs for each driver. Inside the driver specific library, I then had a `Connection with Session.ctl`. 

If you like this idea and want to get ahead of it, I would make the change now. Otherwise, we can defer naming and refactoring around this until you add the initial `Get Connections` API support.

jasonmreding: There is also been an ad hoc convention of creating Typedef virtual folders in the project to contain all type definitions. If that's something we want to employ consistently, you should do that here as well.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get gRPC Device Server Location.vi
jasonmreding, line 1: Might want to update the icon as well with the name change. Maybe something like GET GRPC LOC or GET GRPC SERV.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Session and Driver Info.ctl
pbirkhol-ni: > In some of my prototyping, I split this into two typedefs. I called one Connection.ctl and the other Connection with Generic Session.ctl

This seems like feedback for Get Connections. I have already implemented something very similar to what you mentioned (except with different names). I have `Hardware Independent Connection.ctl`, which contains the pin, site, hardware type, and channel, and then `Connection.ctl`, which is a cluster of a `Hardware Independent Connection.ctl` and a variant session.

jonathanmendez:
Added code to `Create Sessions Impl.vi` to handle an error occurring during the creation of a session. In case of an error, it now Closes/Detaches sessions that were initialized in the loop, and does not update the class data.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get gRPC Device Server Location.vi
jonathanmendez: Good catch, done.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Session and Driver Info.ctl
jonathanmendez: I created a virtual folder for the typedefs.

As for the structure of the typdefs, I'd like to understand your ideas better in some conversation, but in any case I would prefer to defer that kind of change until a later refactor. Paul is building his work on top of this structure (it may not be the most disruptive but I'd rather not get in his way for the unblocking work). Plus we just had a discussion in which we thought through other potential changes to the typedef to help optimize our code (for at least cleanliness, but perhaps some performance), but we want to see the whole picture first.

I think we'll do a better job with a holistic look after Create Sessions and Get Connections have their initial implementations.

jonathanmendez:
image

I wanted to make a note and raise a question about the `IInstrument` class implementations, with regard to potential internal data. I updated this code to use a shift register with the `IInstrument` wire, since a third party instrument implementation could possibly want to store internal data that gets updated as each session is initialized (for example, to maintain a session monitor, or to update some HAL structure). In that case, it also seems plausible that the `Close` may want to do the same thing, and should have an `instrument out` wire.

While that looks unusual at first glance for a "Close" VI, the argument is that what's being closed is not the Instrument but the Session (the second parameter).

It may be very unlikely that someone would implement internal data that needs to be updated with Initialize and Close, but we probably can't/shouldn't assume anything about customer implementations.

@jasonmreding @pbirkhol-ni

pbirkhol-ni:
> image
> 
> I wanted to make a note and raise a question about the `IInstrument` class implementations, with regard to potential internal data. I updated this code to use a shift register with the `IInstrument` wire, since a third party instrument implementation could possibly want to store internal data that gets updated as each session is initialized (for example, to maintain a session monitor, or to update some HAL structure). In that case, it also seems plausible that the `Close` may want to do the same thing, and should have an `instrument out` wire.
> 
> While that looks unusual at first glance for a "Close" VI, the argument is that what's being closed is not the Instrument but the Session (the second parameter).
> 
> It may be very unlikely that someone would implement internal data that needs to be updated with Initialize and Close, but we probably can't/shouldn't assume anything about customer implementations.
> 
> @jasonmreding @pbirkhol-ni

Jonathan convinced me that we should have a class wire out. We can't presume anything about the implementing class, and our `Close MeasurementLink Session.vi` may not be the end of the implementing class' lifetime. If that is the case, though, we need to pass the interface wire out of our VIMs so that a 3rd party instrument implementer could get it back with the updated state. I created #AB2537821 to track this work as well as the work about updating the interface to handle a channels input for the Initialize method.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get gRPC Device Server Location.vi
jasonmreding: - This should check if the `IInstrument` returns an empty string for the provided interface and skip the discovery service lookup if it does. This should allow someone to add pin mapping support on top of a local session.
- The fact that we have hard coded "ni.measurementlink.v1.grpcdeviceserver" as the service class seems bad. That prevents a third party driver implementation that registers itself with the discovery service. I wonder if we should update the `IInstrument` implementation to provide both pieces of information, the service_class and provided_interface. We would want to change the name of the interface method to something like Get Discovery Service Resolve Info.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
jasonmreding, line 1: Should creation of zero sessions be an error similar to `Create Session.vim`?

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
jasonmreding, line 1: There isn't a VI description for this vim. We should probably copy the description from `Create Sessions.vim`.

jasonmreding: I'm not sure how you created the icons for the vims so that they have the orangish background. However, as soon as I open the icon editor, all of the text shifts and things look wonky. It's fine as long as you cancel out, but if you ever need to edit the icon then I can't figure out how to get LV to apply a sane layout to the text. Maybe there is a better way to manage the layers here with text to get the desired outcome? I've tried looking at other vims to see what they do, but none of them use text (only glyphs). Anyway, I'm not sure there is an action to take at this time. It's just something I've noticed that has been driving me crazy.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
jonathanmendez: Good question. Based on some other conversations we may need a more general discussion about expected behavior/errors/etc. surrounding some of these VIs.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
jonathanmendez, line 1: Need to close the sessions in the error case.

jonathanmendez: Updated.

The orangish background is just using the fill tool in the icon editor with the same color of existing VIMs from the LV palette. Paul and I have noticed the text wonkiness and don't understand what's going on. I don't know why it didn't occur when I created this particular icon initially but I saw it when I re-opened the icon editor. My workaround for it is to add a space (" ") into one of the lines of the icon text.

Source/Runtime/MeasurementLink niDCPower/Create Sessions.vi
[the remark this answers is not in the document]
jasonmreding: For the menu names, I prefer using the terms "Single", "Multiple", or "All" as the first work in each menu item. So in this case, it would either be:
- Single Session, Multiple Sessions
- Single Session, All Sessions

Multiple Sessions still works if we add filter criteria in the future. All Sessions more accurately conveys the current behavior. We should be able to change these menu names without breaking compatibility.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
jasonmreding: It's not strictly necessary for correctness, but you might consider adding a stop condition to the for loop and wiring the error cluster to it. That would prevent "empty" iterations where the iteration is starting with the shifted error cluster set to an error state.

jasonmreding:
> > image
> > I wanted to make a note and raise a question about the `IInstrument` class implementations, with regard to potential internal data. I updated this code to use a shift register with the `IInstrument` wire, since a third party instrument implementation could possibly want to store internal data that gets updated as each session is initialized (for example, to maintain a session monitor, or to update some HAL structure). In that case, it also seems plausible that the `Close` may want to do the same thing, and should have an `instrument out` wire.
> > While that looks unusual at first glance for a "Close" VI, the argument is that what's being closed is not the Instrument but the Session (the second parameter).
> > It may be very unlikely that someone would implement internal data that needs to be updated with Initialize and Close, but we probably can't/shouldn't assume anything about customer implementations.
> > @jasonmreding @pbirkhol-ni
> 
> Jonathan convinced me that we should have a class wire out. We can't presume anything about the implementing class, and our `Close MeasurementLink Session.vi` may not be the end of the implementing class' lifetime. If that is the case, though, we need to pass the interface wire out of our VIMs so that a 3rd party instrument implementer could get it back with the updated state. I created #AB2537821 to track this work as well as the work about updating the interface to handle a channels input for the Initialize method.

I guess I'm still not convinced. It seems like it would only be useful if the `IInstrument` itself is calling Close, and I don't think that will ever be the case (or at least not always the case). I guess maybe the argument is that since we're calling Close multiple times on the same `IInstrument`, then we should be chaining the `IInstrument` state through?

Also, at some point, we are going to implement `Close Sessions` on the reservation class and add auto cleanup from the `MeasurementContext` class. These current changes don't account for that since there is no state tracking that captures the `IInstrument` instance needed to close the session(s) that were created. This can be added easily enough, but if it is by value class data, we more or less have the same problem. What if the `IInstrument` implementation has APIs that can be called between Initialize/Close that modify its state? Those state changes won't be reflected in our "snapshot" of the `IInstrument` that was captured during creations of the session(s). That means Close won't be called with the desired state anyway. For that to work, the `IInstrument` implementation will need to implement reference semantics using a DVR. If the class has by ref semantics, then we don't strictly need the ref out wire as any changes will also be reflected in ref in.

So it seems like we can either:
1. Add the ref out terminal which will only be useful if the class only mutates internal data when closing or initializing. Actually, this isn't true either unless we maintain a mapping of the 1 IInstrument instance to N sessions that were created for it in the private data for the reservation.
2. Don't add the ref out terminal and recommend/document that the IInstrument implementation needs to implement reference semantics if it maintains mutable data.
3. Some combination of the above.

Up to this point, I have always been mentally associating one IInstrument instance with one session instance so it took me back a bit when I realized this wasn't the case. I don't know there is anything that can or should be done about this right now, but it left me a little uncomfortable since it mandates all sessions will have the same configuration upon initialization. That's probably fine, but the instancing struck me as odd for some reason. I guess if we're going to allow Close to be called multiple times on `IInstrument` in a valid way, then we should probably have a ref out terminal even if it doesn't address some of the problems I mentioned above.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
pbirkhol-ni, line 1: I am making the following a threaded conversation.

> I wanted to make a note and raise a question about the IInstrument class implementations, with regard to potential internal data. I updated this code to use a shift register with the IInstrument wire, since a third party instrument implementation could possibly want to store internal data that gets updated as each session is initialized (for example, to maintain a session monitor, or to update some HAL structure). In that case, it also seems plausible that the Close may want to do the same thing, and should have an instrument out wire.
While that looks unusual at first glance for a "Close" VI, the argument is that what's being closed is not the Instrument but the Session (the second parameter).
It may be very unlikely that someone would implement internal data that needs to be updated with Initialize and Close, but we probably can't/shouldn't assume anything about customer implementations.
@jasonmreding @pbirkhol-ni

>> Jonathan convinced me that we should have a class wire out. We can't presume anything about the implementing class, and our Close MeasurementLink Session.vi may not be the end of the implementing class' lifetime. If that is the case, though, we need to pass the interface wire out of our VIMs so that a 3rd party instrument implementer could get it back with the updated state. I created #AB2537821 to track this work as well as the work about updating the interface to handle a channels input for the Initialize method.

>>> I guess I'm still not convinced. It seems like it would only be useful if the IInstrument itself is calling Close, and I don't think that will ever be the case (or at least not always the case). I guess maybe the argument is that since we're calling Close multiple times on the same IInstrument, then we should be chaining the IInstrument state through?

Also, at some point, we are going to implement Close Sessions on the reservation class and add auto cleanup from the MeasurementContext class. These current changes don't account for that since there is no state tracking that captures the IInstrument instance needed to close the session(s) that were created. This can be added easily enough, but if it is by value class data, we more or less have the same problem. What if the IInstrument implementation has APIs that can be called between Initialize/Close that modify its state? Those state changes won't be reflected in our "snapshot" of the IInstrument that was captured during creations of the session(s). That means Close won't be called with the desired state anyway. For that to work, the IInstrument implementation will need to implement reference semantics using a DVR. If the class has by ref semantics, then we don't strictly need the ref out wire as any changes will also be reflected in ref in.

So it seems like we can either:

Add the ref out terminal which will only be useful if the class only mutates internal data when closing or initializing. Actually, this isn't true either unless we maintain a mapping of the 1 IInstrument instance to N sessions that were created for it in the private data for the reservation.
Don't add the ref out terminal and recommend/document that the IInstrument implementation needs to implement reference semantics if it maintains mutable data.
Some combination of the above.
Up to this point, I have always been mentally associating one IInstrument instance with one session instance so it took me back a bit when I realized this wasn't the case. I don't know there is anything that can or should be done about this right now, but it left me a little uncomfortable since it mandates all sessions will have the same configuration upon initialization. That's probably fine, but the instancing struck me as odd for some reason. I guess if we're going to allow Close to be called multiple times on IInstrument in a valid way, then we should probably have a ref out terminal even if it doesn't address some of the problems I mentioned above.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Get gRPC Device Server Location.vi
jonathanmendez: I've updated the helper VI to return the whole Remote Connection Options cluster so that it can return a whole default value in the case of an empty provided interface (or service class, once we add that to the query). It avoids calling into discovery service if either is empty. This also changes where I call the helper, putting it into the loop, but since the service location is cached, we still only call into discovery service once.

jonathanmendez: I've added details to the description of AB#2537821 to implement having the `IInstrument` return both service class and provided interface.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
pbirkhol-ni, line 1: Jason has been using <b></b> in the documentation to bold any inputs and outputs. I like that convention and have also been using it.

Source/Runtime/MeasurementLink niDCPower/Create Sessions.vi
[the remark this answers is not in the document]
jonathanmendez: I went with Multiple Sessions
image

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
pbirkhol-ni: The DCPower specific Create Session VIs should also use bolded text for inputs and outputs.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
jonathanmendez: > It's not strictly necessary for correctness, but you might consider adding a stop condition to the for loop and wiring the error cluster to it. That would prevent "empty" iterations where the iteration is starting with the shifted error cluster set to an error state.

I'm assuming this is referring to the Impl VI. I went ahead and put the stop condition on the loop with the error wire.

jonathanmendez: > Need to close the sessions in the error case.

I moved the error check into the Impl VI so that I could utilize the same cleanup code there. The Impl has a boolean input to determine whether to enforce a single-session check.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
jonathanmendez: I think it makes sense to consider whether to do anything about this in the Harden IInstrument user story. I left things as they are for now, i.e. a shift register in the creation loop and no IInstrument output.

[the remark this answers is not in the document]
jonathanmendez: Good catch, fixed.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Helpers/Check for Single Session.vi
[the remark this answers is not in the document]
jonathanmendez: Done now.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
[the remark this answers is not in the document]
jonathanmendez: Added a check and error for an empty instrument type id.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session.vim
jonathanmendez: Done.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions.vim
jonathanmendez: While we may still want to have a conversation about it, it seems reasonable to generate an error if zero sessions are created, and the python implementation is doing the same. I added code to generate an error for 0 sessions (in the implementation VI).

PR #308, opened 2023-10-05
Add Get Connections VIs for NI-FGEN

synesthetize:

Adds poly-VI implementation of `Get Connections` for NI-FGEN

[AB#2533777](https://ni.visualstudio.com/DevCentral/_workitems/edit/2533777): Create NI-Fgen-Specific Implementation of "Get Connections.vi"

[fgen get connections v2.zip](https://github.com/ni/measurementlink-labview/files/12822856/fgen.get.connections.v2.zip)

Added unit tests


Source/Tests/Tests.Runtime/Measurement Server/Get Connections/Test Inputs/Two Sessions Two Pins Two Instruments Two Sites - FGEN.vi
WesleyTangNationalInstruments, line 1: Minor: Consider zero-indexing the sites. Sites are sequential and the use of site2 implies the existence of site0 and site1 which is slightly misaligned with the file name of "Two Sites". The DCPower tests also does this and I guess it isn't technically incorrect to be using site1 and site2 and ignoring site0, but I think it's more clear starting from site0.

WesleyTangNationalInstruments (approved):

pbirkhol-ni (approved):

Source/Tests/Tests.Runtime/Measurement Server/Get Connections/Test Inputs/Two Sessions Two Pins Two Instruments Two Sites - FGEN.vi
pbirkhol-ni: FYI, the reason many of the Get Connection tests don't use a site 0 is because 0 is the default value for an integer in LabVIEW. That makes it more difficult to tell whether a test passed because it did the right thing or whether it passed because it simply had the default value.

That being said, these per instrument tests are not meant to be exhaustive, so I am less concerned about using 0 as a site.

PR #314, opened 2023-10-05
Optimize connection retrieval and sort connections

pbirkhol-ni:

This PR does two main things:
- Filters the connections first by site and then by pin. Prior to this change, connections were sorted by pin and then by site.
- Changes the internal implementation of `Get Connections Impl.vi`. Before it had a triple nested for loop that searched by pins, then by sessions, then by channel mappings. After getting the connections, it used more loops to sort the connections. The new implementation stores a map of connections. It is a triple nested map by instrument type id, then site, then pin. This map is created when reserving sessions. During Create Sessions, the session in the connection will get updated with the newly created session. During Get Connections, we simply query the map for the desired connections and return them. The order is automatically baked into the implementation, so no post sorting is necessary.

This PR also updates tests. For the most part, the tests didn't have to be updated. The only updates were to use the new error code and to fix tests that assumed a different connection ordering.

Auto tests still pass. I should write some new tests for Create Sessions. I will wait for Jonathon to finish his Create Sessions tests before adding my own.


Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections Impl.vi
pbirkhol-ni, line 1: @jasonmreding reminded me offline about system pins, which I think the old implementation handled but this one doesn't.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
pbirkhol-ni, line 1: @jasonmreding @jonathanmendez We branch `connection map` and `session and driver infos` before modifying them. In one branch, we populate them with the newly created sessions. In the other branch, we don't do anything. In the event of an error, we will pass through the unmodified branch. If there is no error, we pass through the modified branch.

The reason I am bringing this up is because it almost certainly makes a data copy. I don't think we care too much about real time performance, so maybe this isn't a big deal. However, it could slow down execution. We don't _have_ to branch the wires, but it makes the logic a lot easier.

Anyway, I don't know if there is any action to take, especially without having any performance benchmarks for comparison. I just thought I'd draw attention to it in case anyone is horrified by it.  :)


image

Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections Impl.vi
pbirkhol-ni: FYI @bkeryan, Jason and I have been discussing offline about how to handle system pins. Right now, the default value for the site input to Get Connections is -1, which means get all pins. However, -1 is also the site value for a system pin. I am considering changing the default value to be -32768 so that customers can use -1 to query only for system pins.

bkeryan: Ok, that sounds reasonable since that's the same sentinel value we use elsewhere.

pbirkhol-ni: Since this will mean modifying the default values for all Get Connections VIs, many of which are still in progress, I created a work item for now. AB#2543570

pbirkhol-ni: System pins are now supported with tests.

Source/Runtime/MeasurementLink Session Management/Connection Map/Get Connections.vi
jasonmreding, line 1: You can probably simplify implementation a bit by just not wiring through the connections in the error case below. That would allow you to remove the `Select` primitive that outputs an empty array in the case of an error.

image

pbirkhol-ni: Done. I also did some general diagram cleanup. For example, the error wire was being passed into loops/cases where no error could be generated, which made it hard to reason about what cases can throw errors.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
jasonmreding: LV doesn't seem to show buffer allocations for maps using the Show Buffer Allocations profiling tool. Without that, it's difficult to say, but I suspect you're right. 

If you wanted to avoid the copy, I guess you would have to clear out the sessions in the various data structures/wires so you never need to fork the wire. If we move forward with the close sessions and auto cleanup functionality, we'll need to implement something like this anyway. Maybe that's motivation enough to try and fix this now. Or you could wait and revisit this once we get to that work since I suspect there will be some overlap in functionality that could be shared here.

Source/Runtime/MeasurementLink Session Management/Connection Map/Get Connections.vi
jasonmreding, line 1: There is some work being done here is that is only valid if you want to generate an error. Maybe it would be better to wrap that section in another case structure so you can optimize out that work.

image

Similarly, there is some effort spent building up an array of pin names that were matched which is only used in the case where errors are reported if the pin is not found. I think you could just extract that information from the list of matching connections in that case. You could also just shift through a pin not found Boolean if you don't want to incur the expense of calculating by building sets / arrays and then comparing the collections. There are some trades offs between readability and performance here so I'll let you use your own discretion on which changes you think are worth while or not.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections Impl.vi
jasonmreding, line 1: Consider creating a constant for -1 value meaning system pins.

jasonmreding, line 1: I think we should be adding -1 for system pins to the list of sites all of the time. I'm pretty sure if you ask the pin map service for all pin connections on site 0, it will also return system pins as part of the returned data. If there isn't a test for this, we should probably add one.

jasonmreding (commented):
We should probably add some tests for shared pins as well as I think we have some gaps in regards to them. In particular, if you have a shared pin on site 0 and site 1 and then reserve sessions for site 0 and 1 and include the system pin name, then I believe you will only get back one channel mapping in the session to that pin, and it will only list one of the sites. If you then called Get Connections on site 0 followed by Get Connections on site 1, I think one of the calls will return the connection and the other one won't with the current code.

I'm not really sure what to do about this. On one hand, I suspect it might not matter since I don't think it makes sense in general to perform measurements on a shared pin. However, generating a value on shared pin is probably more realistic (e.g. a power supply voltage). In this type of mulit-site scenario, I'm not sure how customers generally write their measurement logic. I can see how it might work if you just accept you might duplicate configuration/programming on the shared pin. However, I'm not sure how you would structure the code so that you only do the work once unless you explicitly write the code knowing the pin is shared ahead of time.

If we wanted to fix this, it seems like we would either need to:
1. Update the pin map service to return an array of sites as part of the channel mappings.
2. Update the pin map service to duplicate information in the channel mappings for the shared pin.

The first option doesn't sound all that practical as it would break backward compatibility with the service API. The second option seems more reasonable, but I would need more time to think through the ramifications of that as I think it would change behavior for the current service API as well as the underlying pin map framework library. I imagine @jonathanmendez might have some thoughts on this as well. If there is no obvious straight forward fix for this, it might be better to create a work item and revisit this in a follow up PR.

jasonmreding (approved):

Source/Runtime/MeasurementLink Session Management/Connection Map/Get Connections.vi
pbirkhol-ni: > There is some work being done here is that is only valid if you want to generate an error. Maybe it would be better to wrap that section in another case structure so you can optimize out that work.

Done.

> Similarly, there is some effort spent building up an array of pin names that were matched which is only used in the case where errors are reported if the pin is not found.

You can get this information from the list of connections, but it requires an extra loop to extract the pins from the map. Since we are already iterating over the pins anyway, I decided to cache them off to avoid the extra loop later. The "pin not found" boolean in a shift register isn't as simple as it seems since a key could be missing for one site but not another. So "key not found?" being true for a pin doesn't mean that the pin doesn't exist on another site. I'm sure it would be possible to get this to work, but the logic wouldn't be trivial.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections Impl.vi
pbirkhol-ni: I have a user story for this. AB#2543570. Right now, -1 doesn't mean system pins, so it doesn't make sense to create the constant yet.

jasonmreding: Currently, there are two -1 constants on the diagram. One of them means system pins / sites. The other currently means all sites, although I think you are planning to change that.

Source/Runtime/MeasurementLink Session Management/Connection Map/Get Connections.vi
jasonmreding: You're actually iterating over an array of equivalent length anyway so I wouldn't expect the iterating overhead to be much of a factor. I was mostly speaking to the cost of building up an array that might not be needed. I don't know how efficient the LV compiler is at this when using for loop tunnels to build the array. Ideally, the compiler would be smart enough to allocate enough space up front so it doesn't have to constantly grow the array and copy the old contents into the new array. However, there are some conditional tunnels at play here so I'm not sure how much the compiler could realistically infer. 

On the other hand, if you had a single for loop at the end (no nesting) that built up an array, I think the chances are much higher that the compiler would be able to allocate an array of the correct size up front without performing multiple allocations. To some degree this is irrelevant since the array is just an intermediate data structure so that you can build up a set in a convenient way. You could just as easily do this from a single array of connections vs. creating a connections array and a pin array, although it means you wouldn't be able to use the same sub VI to build up both sets.

pbirkhol-ni:
> We should probably add some tests for shared pins as well as I think we have some gaps in regards to them. In particular, if you have a shared pin on site 0 and site 1 and then reserve sessions for site 0 and 1 and include the system pin name, then I believe you will only get back one channel mapping in the session to that pin, and it will only list one of the sites. If you then called Get Connections on site 0 followed by Get Connections on site 1, I think one of the calls will return the connection and the other one won't with the current code.
> 
> I'm not really sure what to do about this. On one hand, I suspect it might not matter since I don't think it makes sense in general to perform measurements on a shared pin. However, generating a value on shared pin is probably more realistic (e.g. a power supply voltage). In this type of mulit-site scenario, I'm not sure how customers generally write their measurement logic. I can see how it might work if you just accept you might duplicate configuration/programming on the shared pin. However, I'm not sure how you would structure the code so that you only do the work once unless you explicitly write the code knowing the pin is shared ahead of time.
> 
> If we wanted to fix this, it seems like we would either need to:
> 
> 1. Update the pin map service to return an array of sites as part of the channel mappings.
> 2. Update the pin map service to duplicate information in the channel mappings for the shared pin.
> 
> The first option doesn't sound all that practical as it would break backward compatibility with the service API. The second option seems more reasonable, but I would need more time to think through the ramifications of that as I think it would change behavior for the current service API as well as the underlying pin map framework library. I imagine @jonathanmendez might have some thoughts on this as well. If there is no obvious straight forward fix for this, it might be better to create a work item and revisit this in a follow up PR.

I tested this and you're right on all accounts. If I am using All Sites and have a shared pin, the channel mapping only lists the shared pin for a single site. I am going to defer this for now since I don't know how to fix it given the information that we get from the session manager. As far as I can tell, there is no indication that a pin is shared. Perhaps our guidance should be to call Get Connections on All Sites when getting connections for shared pins. This would require the measurement author know that a pin is shared, but they should know that, I think. I have captured this in AB#2543570.

Source/Runtime/MeasurementLink Session Management/Connection Map/Get Connections.vi
pbirkhol-ni: >  Ideally, the compiler would be smart enough to allocate enough space up front so it doesn't have to constantly grow the array and copy the old contents into the new array.

I am fairly certain this is true. I'm less sure for conditional tunnels, though. In any case, I think I'm going to leave this as is unless you have strong objections.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
pbirkhol-ni: I added this to AB#2543570, which represents non-showstopper follow up work for Get Connections.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections Impl.vi
pbirkhol-ni: I confirmed the behavior you mentioned and made the change. I also added two tests.

PR #317, opened 2023-10-06
Automatically find tests instead of using a test list

pbirkhol-ni:

Currently, we keep a list of tests to run during a PR build. If you forget to add a test to that list, it won't run. This is annoying and error prone. 

This PR automatically finds all tests that begin with "test_". It also changes the name of all tests to start with "test_".

I added as skip tests file for tests that shouldn't be run.

PR build


Source/Tests/skip_tests.lst
```diff
@@ -0,0 +1 @@
+test_DefaultTemplateMeasurement_GetUiDetails_CorrectUiPathUrl.vi
```
pbirkhol-ni, line 1: This test doesn't run because it requires measurementlink-labview to be installed. Our test runner currently doesn't install measurementlink-labview and instead runs the tests against source.

Source/Tests/run_tests.vi
pbirkhol-ni, line 1: This is the only VI with any real changes. I kept all of the test running and reporting the same. The only thing I changed was how tests are collected.

image

Source/Runtime/Measurement Server.lvproj
```diff
@@ -68,71 +68,43 @@
 				<Item Name="JDP Timestamp.lvlib" Type="Library" URL="/&lt;vilib&gt;/JDP Science/JDP Science Common Utilities/Timestamp/JDP Timestamp.lvlib"/>
 				<Item Name="Get LV Class Name.vi" Type="VI" URL="/&lt;vilib&gt;/Utility/LVClass/Get LV Class Name.vi"/>
 				<Item Name="Base64 Support.lvlib" Type="Library" URL="/&lt;vilib&gt;/JDP Science/JDP Science Common Utilities/Base64/Base64 Support.lvlib"/>
-				<Item Name="Simple Error Handler.vi" Type="VI" URL="/&lt;vilib&gt;/Utility/error.llb/Simple Error Handler.vi"/>
```
DelpireNI, line 71: Is all this stuff supposed to be removed from here?

DelpireNI (approved):

Source/Runtime/Measurement Server.lvproj
pbirkhol-ni: I don't know. I feel like these VIs randomly appear and disappear. I'm inclined to just trust that LabVIEW knows what it's doing.

dixonjoel: They are in the dependencies section, so I agree with @pbirkhol-ni here that these should be managed by LabVIEW.

jasonmreding: Yeah, it's annoying that the automatically calculated list of dependencies by LV is persisted in the project file such that it creates noise when trying to view "real" edits made to the project file. It can sometimes be useful if you see something from a package show up that you don't expect to depend on. However, it's mostly just noise.

PR #325, opened 2023-10-09
Initial MeasurementLink palette with MeasurementContext VIs

dixonjoel:

Add a MeasurementLink palette to our MeasurementLink Service VI package. This is a minimal palette with only measurement context VIs on it.
Looks like this:
image

On disk, this .mnu file is added when the package is installed.
image

Begin to make the MeasurementLink LabVIEW API for measurement services.

Manual editing and dropping VIs from the palettes.


Source/Runtime/MeasurementLink_FGEN.mnu
dixonjoel, line 1: Fix the short name of 'Create.vi'...

Source/Runtime/MeasurementLink.mnu
pbirkhol-ni, line 1: I can't figure out how to organize these on disk such that they show up correctly in LabVIEW. I eventually got _something_ working, but it isn't right. I realize that our installer will handle this for us, but maybe we can organize the palette files in source that same way we expect them to be installed?

Source/Runtime/MeasurementLink_DCPower.mnu
pbirkhol-ni, line 1: I don't think the Create Instrument VIs should be on the palette.

Source/Runtime/MeasurementLink.mnu
pbirkhol-ni, line 1: You also should put `Create Session.vim` on the palette.

pbirkhol-ni, line 1: You should put the 3 Get Connections.vim malleable VIs on the palette. I feel like we should hide the Create Sessions.vim and Get Connections.vim in an Advanced palette or something like that. Those VIMs are only valuable for customers who are trying to implement support for 3rd party devices.

pbirkhol-ni, line 1: We will need to do a better job customizing the palettes in the right click menu. I don't think you even attempted to do that in this PR, which is fine.

pbirkhol-ni, line 1: I don't know what to do with the `Session Reservation.Reserve Sessions.vi.` For one, it is community scoped, so customers can't even use it. However, it is valuable for code modules. Maybe we can put it on a `Code Module` subpalette and make it public? @jasonmreding has thought about this more than I have.

pbirkhol-ni, line 1: We'll eventually have a Generic Get Connections VI and a Generic Get Session Infos VI. We don't have to add those now, but we should make sure they fit your proposal, and I think they do. They could go on a different subpalette similar to the NI Digital or NI DCPOWER subpalettes.

pbirkhol-ni, line 1: It's weird that the context helper shows a generic class wire when hovering over a VI in the palette. As soon as a drop the VI on the diagram, the context help gets updated to show the correct class wire. This seems like something that is fixable, but I don't know how.

image

pbirkhol-ni (changes requested):

Source/Runtime/MeasurementLink.mnu
jasonmreding: I think we still need an explicit decision as to whether we are going to maintain individual .mnu source files using traditional mechanisms in LV to create/edit the files or whether we are going to use VIPB to create/edit the menu files. In the latter case, all of the menu files seem to be stored in the .vipb binary file itself. 

I also think we need a better understanding of how we are going to maintain the right click context menus for classes/libraries vs. the the functions palette menus. Does it require duplication of the menu file(s) or can we reference the same menu file(s) from multiple locations? There is probably some research / prototyping that needs to happen in this regard. The VIPB palette editor is much nicer in a lot of ways, but it doesn't offer all of the advanced features that LV exposes. However, I don't know if we will need/want any of these advanced features.

If we are going to maintain our .mnu files in source, then it would make more sense to me to keep them all in a directory structure like "menus\Categories\Measurement" which matches their desired installation hierarchy relative to the LV root directory. However, before finalizing that, I would like to understand the strategy for specifying the right click context menus vs. functions palette menus. It might be that it makes more sense to have the .mnu files in source next to their respective class/library file where the file is added to the class/library and configured in the class/library settings to be the palette for the class/library. Then under the functions palettes, we just link to the existing class/library palette files where appropriate. Creating/editing this will probably have to be done from a pseudo-install setup so that things are linking to <vilib> and not relative paths from the source directory.

jasonmreding: Ultimately, I think the API should be made public so that it can be used from code modules. The same is true for register/unregister session APIs. However, the positioning/naming of this palette relative to the MeasurementContext based APIs is still listed as an open issue in the HLD I posted. If you want to propose/discuss a solution here, go for it. Otherwise, we should expect to revisit whatever is done here at some point in the future.

Source/Runtime/MeasurementLink_DCPower.mnu
jasonmreding, line 1: One question that pertains to all the driver specific palettes is whether the palette/short names should match what the actual driver uses? If the answer is yes, then we need to tweak things here. It's not consistent, but most of the drivers use all caps with a space instead of a hyphen for the sub-palette name when viewed from the parent palette. The exceptions I'm aware of are "NI Digital" and "NI DAQmx" which don't use all caps. 

Once you navigate into the driver palette, the name of the palette then uses a hyphen instead of a space in the name (I don't know why). "NI DCPOWER" also becomes "NI-DCPower". Maybe this isn't worth sweating over, but I thought I would bring it up.

jasonmreding (commented):
I think @pbirkhol-ni and I are somewhat add odds at the moment in regards to feedback as to what should and should not be on the palette. I'm still open to arguments on this, but these are my current thoughts on the issue:
- Driver specific `Create Instrument` should not be on the palette.
- No .vims that are used to implement driver specific `IInstrument` plugins should be on the palette. I think it's too much noise/confusion for the average user. Instead, we can steer plugin authors to our implementations on the palette (nothing is password protected) and/or we can create a project template that we install as a getting started guide for these customers.
- I don't think we should initially put any generic `Get Connections` API on the palette that returns a variant session. I would rather wait for user feedback before adding them, as I'm not convinced they are needed and might cause confusion.

jasonmreding (changes requested):
I want to have a clearer plan on how we are going create/edit mnu files before approving.

Source/Runtime/MeasurementLink_DCPower.mnu
pbirkhol-ni: The MI drivers are so inconsistent. I'd rather be consistent within our own palette. Ideally, that would mean using the official driver names like `NI-DCPower`, `NI-FGEN`, etc.

Source/Runtime/MeasurementLink.mnu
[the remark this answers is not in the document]
pbirkhol-ni: From Jason:

> I think @pbirkhol-ni and I are somewhat add odds at the moment in regards to feedback as to what should and should not be on the palette. I'm still open to arguments on this, but these are my current thoughts on the issue:
> - Driver specific Create Instrument should not be on the palette.
> - No .vims that are used to implement driver specific IInstrument plugins should be on the palette. I think it's too much noise/confusion for the average user. Instead, we can steer plugin authors to our implementations on the palette (nothing is password protected) and/or we can create a project template that we install as a getting started guide for these customers.
> - I don't think we should initially put any generic Get Connections API on the palette that returns a variant session. I would rather wait for user feedback before adding them, as I'm not convinced they are needed and might cause confusion.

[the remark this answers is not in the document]
pbirkhol-ni: > Driver specific Create Instrument should not be on the palette.

We agree on this.

> No .vims that are used to implement driver specific IInstrument plugins should be on the palette.

I don't completely disagree here. However, 3rd party instrument support has been a requirement for every single customer that I have talked to, so we should at least consider promoting it on the palette. We could have a separate top level palette for developing 3rd party instrument support. I'm not sure that makes things any less confusing, though.

> I don't think we should initially put any generic Get Connections API on the palette

This is fine.

pbirkhol-ni (changes requested):

Source/Runtime/MeasurementLink_DCPower.mnu
dixonjoel: This is different from the HLD, right? I was trying to follow this:
https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/562608

Source/Runtime/MeasurementLink.mnu
dixonjoel: Yeah, I'm trying to understand how that even works. I can't seem to find any documentation on right-click palettes for classes / libraries / etc. Do you have any knowledge of that?

dixonjoel: This is the first I'm learning about the VIPM options for palette files. I'm playing with that, but it seems to me that the whole MeasurementLink functions palette could be defined in the .vipb and that would not require us to store .mnu files in source.

I don't know how right-click menus for classes / libraries work yet so I can't comment on that yet, but I don't see any way to define that in the .vipb unless I'm missing something. So for those, if it's true that you just place a .mnu file under the class to define it, then we'd just keep it next to the class there.

dixonjoel: My current intention is to edit the .vipb to create the MeasurementLink palette under Measurement I/O so we won't have to manage those files.

And then keep the class .mnu file next to the class in source.

For this PR, my intention is to do only the top level MeasurementLink palette and not the driver subpalettes so we can have a small chunk to review.

dixonjoel: I know how this works now and have been able to get a .mnu file working with the class. But I'll defer that to another PR.

dixonjoel: Future PR

jasonmreding: > I don't know how right-click menus for classes / libraries work yet so I can't comment on that yet, but I don't see any way to define that in the .vipb unless I'm missing something. So for those, if it's true that you just place a .mnu file under the class to define it, then we'd just keep it next to the class there.

Yes, VIPB supports this although I don't find it to be that intuitive. Check the documentation which alludes to this. If I remember correctly, I think you have to define a unique Palette Set per class/library which then automatically gets added to the class as part of the build process. I haven't played with this much so I could be wrong about how it works, but the capability is definitely supported in some fashion.

image

Source/Runtime/MeasurementLink_DCPower.mnu
jasonmreding: The HLD was a rough first pass and not a finished product. I think in the intervening time, we are leaning towards leaving it off the palette based on what we know now.

Source/Runtime/MeasurementLink.mnu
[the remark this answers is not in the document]
jasonmreding: >>    No .vims that are used to implement driver specific IInstrument plugins should be on the palette.

> I don't completely disagree here. However, 3rd party instrument support has been a requirement for every single customer that I have talked to, so we should at least consider promoting it on the palette. We could have a separate top level palette for developing 3rd party instrument support. I'm not sure that makes things any less confusing, though.

I am currently leaning towards the idea of shipping a template project for creating a third party driver plugin or just telling people to use the NI specific driver wrappers we ship as a template. I'm not necessarily opposed to placing them on the palette. However, the vast majority of users will never need them, and it would reduce noise in what is becoming a very large palette hierarchy.

dixonjoel: Later PR

dixonjoel: Postponing and starting with measurement context.

dixonjoel: To be addressed in a later PR.

dixonjoel: Later PR

Source/Runtime/MeasurementLink_DCPower.mnu
dixonjoel: I'll plan on using the 'official driver names' going forward.

Source/Runtime/MeasurementLink.mnu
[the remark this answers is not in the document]
dixonjoel: I'll keep these in mind for the future PRs

pbirkhol-ni (approved):

jasonmreding:
This seems reasonable for an initial submission. I still have questions as to whether we will want a sub-palette for the MeasurementContext APIs or not. If we do, then it seems like there is opportunity to share the same .mnu file for the functions palette and the class right click context palette.

jasonmreding (approved):

PR #326, opened 2023-10-09
Add unreserve and close to the session reservation class

pbirkhol-ni:

- Add `Unreserve Sessions.vi` to the session reservation class.
- Add `Close Sessions.vi` to the session reservation class. For now, there is only an option to close all sessions. We could add per-instrument close if desired.
- Add the `IInstrument` implementation to the private data of the Session Reservation class when sessions are created. 
- Use the State Pattern to invalidate the class when Unreserve Sessions has been called. There are two states, both of which implement `Istate.lvclass`: `Reserved State.lvclass` and `Unreserved State.lvclass`. When `Reserve Sessions.vi` is called, the `state` field in the private data gets set to `Reserved State.lvclass`. When `Unreserve Sessions.vi` is called, the `state` field gets set to `Unreserved State.lvclass`.
- The `Unreserved State.lvclass` errors for all invocations.
- The `Reserved State.lvclass` has no real logic. It just dispatches all calls back to the Session Reservation class, which handles all of the logic. The way this is implemented is a little strange. When a call to Create/Get Connections/Close/Unreserve comes in, the `Data.ctl` is unbundled from the private data, passed to the `Istate.lvclass` instance, then passed to an Impl VI that implements the logic. The Impl VI will optionally modify the `Data.ctl`, pass it back to the `Istate.lvclass` instance, and then the original call will rebundle the `Data.ctl`. Why the weird pattern? Mostly because LabVIEW does not allow recursive DVR locks. In an ideal situation, we would lock the DVR to get the `state`, then dispatch the right `Istate.lvclass` which will in turn lock the DVR again to get at the data. This will hang LabVIEW, though, hence my workaround. I am open to other suggestions.
- Update tests. I didn't have to update any tests, only the Test Inputs. We have to set the original `state` to be `Reserved State.lvclass` for all existing tests.

Only basic sanity tests for now. Automated tests will come later.


pbirkhol-ni:
> > ```
> > The way this is implemented is a little strange. When a call to Create/Get Connections/Close/Unreserve comes in, the `Data.ctl` is unbundled from the private data, passed to the `Istate.lvclass` instance, then passed to an Impl VI that implements the logic. The Impl VI will optionally modify the `Data.ctl`, pass it back to the `Istate.lvclass` instance, and then the original call will rebundle the `Data.ctl`.
> > ```
> 
> At first glance, it appears that the "Impl" VIs could be moved (or rather, have their code inlined) to the `Reserved State.lvclass`. This would remove one layer of "empty" dispatch VIs. But without having tried this for real, I don't know if there are any technical limitations you'd run into (accessing private data?).
> 
> If you did do this, the result would look like:
> 
> * `Session Reservation` - Holds data and only a set of API VIs that dispatch to the state class
> * `Reserved State` - Has a set of implementation VIs for the API as well as several supporting helper VIs
> * `Unreserved State` - Has only a set of erroring VIs for the API

That is certainly something I considered. I don't think there would be any issues with private data access. I don't have a good reason for not doing it that way. My main reasons for doing it the way I did are:
- Keeps all of the logic in a single class. The State classes do as little work as possible.
- Only the Session Reservation class modifies its private data. However, you could argue that this makes it more confusing, because it isn't accessing its private data directly, it is doing it via the Data.ctl cluster that it is passed indirectly from itself. 

Ideally, the State classes would keep a copy of the Session Reservation class, but that would lead to the nested DVR access.

jasonmreding:
> > > ```
> > > The way this is implemented is a little strange. When a call to Create/Get Connections/Close/Unreserve comes in, the `Data.ctl` is unbundled from the private data, passed to the `Istate.lvclass` instance, then passed to an Impl VI that implements the logic. The Impl VI will optionally modify the `Data.ctl`, pass it back to the `Istate.lvclass` instance, and then the original call will rebundle the `Data.ctl`.
> > > ```
> > 
> > 
> > At first glance, it appears that the "Impl" VIs could be moved (or rather, have their code inlined) to the `Reserved State.lvclass`. This would remove one layer of "empty" dispatch VIs. But without having tried this for real, I don't know if there are any technical limitations you'd run into (accessing private data?).
> > If you did do this, the result would look like:
> > 
> > * `Session Reservation` - Holds data and only a set of API VIs that dispatch to the state class
> > * `Reserved State` - Has a set of implementation VIs for the API as well as several supporting helper VIs
> > * `Unreserved State` - Has only a set of erroring VIs for the API
> 
> That is certainly something I considered. I don't think there would be any issues with private data access. I don't have a good reason for not doing it that way. My main reasons for doing it the way I did are:
> 
>     * Keeps all of the logic in a single class. The State classes do as little work as possible.
> 
>     * Only the Session Reservation class modifies its private data. However, you could argue that this makes it more confusing, because it isn't accessing its private data directly, it is doing it via the Data.ctl cluster that it is passed indirectly from itself.
> 
> 
> Ideally, the State classes would keep a copy of the Session Reservation class, but that would lead to the nested DVR access.

I don't know if this is better or worse, but it seems like you could have made the state class a simple error/validation class. Rather than having it wrap/dispatch every public API, the existing implementations would just always perform an error check with the state class first thing to determine whether they should do anything or not. That might still get somewhat tricky with functions like Close/Unreserve if the desire is for the function to always clean up in spite of any error in state, but I'm sure it can be managed.

Perhaps the biggest downside to that approach is that you always have to remember to call it first thing in every public API as an implementation detail. However, I'm not sure it's all that much different than what you have now since you have to modify the state interface each time a new public API is added and remember to use the same dispatch pattern for the implementation. However, I admit it's probably less likely to forget to do something with that approach vs. forgetting to call an error state checking VI to see if the class is in the correct state to proceed with the current function. However, it does seem much simpler.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
jasonmreding, line 1: This has the same problem we discussed in a previous PR. Since we are now storing the `IInstrument` data as part of our state and then later using the "snapshot" to close sessions, it won't reflect any state changes made to the `IInstrument` since then. That means if the class author has to provide "reference" semantics for the class if they have mutable state. I don't really see a way around this, so I think we either have to:
- Live with imposing this requirement on class authors and document it accordingly.
- Don't offer any sort of session closing APIs.
- Create a separate `IInstrumentFactory` interface that is separate from the `IInstrument` interface.

I kind of like the last option for other reasons, but I don't think it entirely solves things. It just means the `IInstrumentFactory` implementation now has to provide reference semantics if wants to maintain mutable state. We could also consider at least always storing the latest snapshot in the map when multiple sessions are created vs. the snapshot when the session was created. This would probably require some pretty expensive look ups or complicated state management / ref counting so that all sessions could point at the same reference/data for the `IInstrument`. All of this would also apply to the `Close Sessions` implementation.

At the moment, I'm inclined to just live with imposing this as a requirement. However, I also wouldn't be opposed to just saying you have to use the driver APIs explicitly to close sessions. The only real downside to that is it forces the measurement to always deal with conditional logic for how to close the session (detach vs. close).

pbirkhol-ni: For this PR, I am not going to make any changes. As we discussed at lunch, we should consider renaming IInstrument to something more descriptive. That might be IInstrumentFactory, or it might be something else. Its behavior will remain exactly the same, though. I have captured this work in AB#2537821

pbirkhol-ni:
> > > > ```
> > > > The way this is implemented is a little strange. When a call to Create/Get Connections/Close/Unreserve comes in, the `Data.ctl` is unbundled from the private data, passed to the `Istate.lvclass` instance, then passed to an Impl VI that implements the logic. The Impl VI will optionally modify the `Data.ctl`, pass it back to the `Istate.lvclass` instance, and then the original call will rebundle the `Data.ctl`.
> > > > ```
> > > 
> > > 
> > > At first glance, it appears that the "Impl" VIs could be moved (or rather, have their code inlined) to the `Reserved State.lvclass`. This would remove one layer of "empty" dispatch VIs. But without having tried this for real, I don't know if there are any technical limitations you'd run into (accessing private data?).
> > > If you did do this, the result would look like:
> > > 
> > > * `Session Reservation` - Holds data and only a set of API VIs that dispatch to the state class
> > > * `Reserved State` - Has a set of implementation VIs for the API as well as several supporting helper VIs
> > > * `Unreserved State` - Has only a set of erroring VIs for the API
> > 
> > 
> > That is certainly something I considered. I don't think there would be any issues with private data access. I don't have a good reason for not doing it that way. My main reasons for doing it the way I did are:
> > ```
> > * Keeps all of the logic in a single class. The State classes do as little work as possible.
> > 
> > * Only the Session Reservation class modifies its private data. However, you could argue that this makes it more confusing, because it isn't accessing its private data directly, it is doing it via the Data.ctl cluster that it is passed indirectly from itself.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Ideally, the State classes would keep a copy of the Session Reservation class, but that would lead to the nested DVR access.
> 
> I don't know if this is better or worse, but it seems like you could have made the state class a simple error/validation class. Rather than having it wrap/dispatch every public API, the existing implementations would just always perform an error check with the state class first thing to determine whether they should do anything or not. That might still get somewhat tricky with functions like Close/Unreserve if the desire is for the function to always clean up in spite of any error in state, but I'm sure it can be managed.
> 
> Perhaps the biggest downside to that approach is that you always have to remember to call it first thing in every public API as an implementation detail. However, I'm not sure it's all that much different than what you have now since you have to modify the state interface each time a new public API is added and remember to use the same dispatch pattern for the implementation. However, I admit it's probably less likely to forget to do something with that approach vs. forgetting to call an error state checking VI to see if the class is in the correct state to proceed with the current function. However, it does seem much simpler.

Given that there isn't an obviously better way, I'm going to submit what I have. None of this is public so we can refactor at any time.

jasonmreding:
> Given that there isn't an obviously better way, I'm going to submit what I have. None of this is public so we can refactor at any time.

Technically, all of the `IState` interface and concrete classes are all public, but I get your meaning.

pbirkhol-ni:
> > Given that there isn't an obviously better way, I'm going to submit what I have. None of this is public so we can refactor at any time.
> 
> Technically, all of the `IState` interface and concrete classes are all public, but I get your meaning.

I'd like to make them private or community scoped. But when I do that, the tests break, even after setting up proper friendship relationships. It has something to do with the json VI we are using.

PR #329, opened 2023-10-11
Creating NI-SCOPE Acquire Waveform example.

DelpireNI:

Creates an NI-SCOPE example that demonstrates how to use our new session management APIs with NI-SCOPE and acquire a waveform.

This example is modeled after our [MeasurementLink Python NI-SCOPE Acquire Waveform example](https://github.com/ni/measurementlink-python/tree/main/examples/niscope_acquire_waveform).

We want to have examples demonstrating how to use our new session management APIs with each of our device drivers. 

[#AB2540544](https://dev.azure.com/ni/DevCentral/_backlogs/backlog/ASW%20MeasurementLink/Features?showParents=false&workitem=2540544)

Example runs successfully and returns waveform data.


Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement Configuration.ctl
DelpireNI, line 1: Other examples seem to have the default values in this file. I have tried on two different occasions to set them here, but every time I save and close the project they are cleared. They do not show up when loading in InstrumentStudio.

Does anyone know what I need to do to get default values in a LabVIEW measurement?

jasonmreding: You are doing this?

image

 or

image

DelpireNI: I didn't even know that was on option! I'll give it a try, thank you!

synesthetize: You need to set your default values in the `Measurement Configuration.ctl` and then before saving you need to go to Edit -> "Make Current Values Default".

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement UI.vi
pbirkhol-ni, line 1: If you're feeling crazy, I think you can probably programmatically update the Plot Legend for the graph to match the pin inputs.  :)

pbirkhol-ni (changes requested):
I need to install NI-SCOPE to finish this review. I figured I'd post my initial comments first, though.
Make sure you also work on the TestStand aspect of the example.

DelpireNI:
> I need to install NI-SCOPE to finish this review. I figured I'd post my initial comments first, though. Make sure you also work on the TestStand aspect of the example.

I was thinking we could do the TestStand work in a separate PR.

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform.lvlib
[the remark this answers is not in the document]
DelpireNI: Fixed. Hopefully I caught all of them.

Source/Example Measurements/NI-SCOPE Aquire Waveform/Measurement Service Helpers/Measurement Service Helpers.lvlib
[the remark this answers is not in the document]
DelpireNI: I did not use anything from there. I think this is generated from our generation tool. I deleted the helpers and removed the reference from the project and the example still runs correctly.

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Get Measurement Details.vi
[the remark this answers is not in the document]
DelpireNI: Fixed:
image

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement UI.vi
DelpireNI: Felt crazy! This seems to work just fine:

image

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Run Service.vi
pbirkhol-ni, line 1: Can you customize the icon for this VI like we did for DCPower?

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement UI.vi
pbirkhol-ni: Not crazy enough. I was thinking something like this:
image

That will update the Plot Legend with the names of the pins too.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Measurement Logic.vi
pbirkhol-ni, line 1: I noticed that if I set my Pin Names to Pin1, Pin2, Pin3 and then set my Trigger Source to Pin4, I get an error about Pin4 not having any connections, which is correct since we didn't reserve Pin4. Should the measurement be adding the trigger source to the list if pins to reserve if it isn't already in the list?

pbirkhol-ni, line 1: The DCPower example requires a persistent simulated device or a real device, whereas this example creates a session-based simulated device. I don't think using a session-based simulated device is the right thing to do unless there is a discoverable way for customers to modify the example to use real hardware. I think @bkeryan has had similar conversations recently for python. Did we make any official decisions for how to handle this in python?

pbirkhol-ni (approved):

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement UI.vi
DelpireNI: That is even crazier.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Measurement Logic.vi
bkeryan: The current plan is to configure simulation by setting environment variables or creating a `.env` file: https://github.com/ni/measurementlink-python/blob/main/.env.simulation

This way, the examples will use real hardware by default, and you can enable simulation by copying one file.

Source/Example Measurements/NI-SCOPE Aquire Waveform/NI-SCOPE Acquire Waveform/Measurement UI.vi
DelpireNI: Added this. Works great. :)

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Measurement Logic.vi
DelpireNI: Okay I fixed this by adding the Trigger source to the list of pins, removing duplicates, and then passing that into reserve pins. I just realized though that this will mean all the pins will come out of Create Sessions' channel list. So I need to update it to call Get Connections for the pins we want to use.

DelpireNI: This should now be resolved. Trigger source will get reserved, and will not be used if it isn't in the Pin Names parameter.

pbirkhol-ni, line 1: I would remove the False boolean that is wired to Create Sessions. Reset is false by default, so explicitly wiring it just causes unnecessary clutter.

pbirkhol-ni, line 1: I don't think you should put Initiate and Fetch in the same for loop. That effectively serializes the measurement. It will also call Initiate multiple times on the same session. Instead, Initiate shouldn't be in a for loop and should use the session from Create Sessions. Then you should call Fetch in the for loop.

pbirkhol-ni: If you had multiple sessions, the only thing that would change is that you would Initiate in its own for loop.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Get Type Specializations.vi
pbirkhol-ni, line 1: The icon for this VI has a weird overlay.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform.lvproj
pbirkhol-ni, line 1: There is some inconsistency between your example and the NI-FGEN and NI-DCPower examples. Those examples condense the project name and library name into a single string with no spaces. I don't know why it is done that way and I don't think one way is necessarily preferable over another, but we should be consistent.

pbirkhol-ni (approved):

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Measurement Logic.vi
DelpireNI: I was going to leave it out originally, but that won't work if we have multiple sessions. I guess this won't work for multiple sessions unless you replace Create Sessions with a Multiple session variant.

I'll go ahead and remove it from the loop.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform.lvproj
DelpireNI: I'll remove the spaces.

DelpireNI: Files have been renamed. :)

pbirkhol-ni:
Also make sure you try building the measurement into an exe and running it that way. You don't necessarily have to run it as a static measurement. You can just build it into an exe, run it, and then open it using InstrumentStudio.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Measurement Logic.vi
pbirkhol-ni, line 1: The other examples are using 60000 as the timeout for Reserve.

DelpireNI: Done:
image

DelpireNI:
> Also make sure you try building the measurement into an exe and running it that way. You don't necessarily have to run it as a static measurement. You can just build it into an exe, run it, and then open it using InstrumentStudio.

Tried this out and it worked just fine. :)

PR #330, opened 2023-10-11
Close/unreserve session reservations when destroying MeasurementContext

jonathanmendez:

Auto-closes and auto-unreserves any sessions reserved on the MeasurementContext class after a measurement's logic has run.

- Stores SessionReservation classes that are created when calling Reserve Session(s) in the MeasurementContext class data
- Calls `Close Sessions.vi` and `Unreserve Sessions.vi` on all stored SessionReservations in `MeasurementContext:Destroy.vi`

Enables easier session management for measurement authors.

TODO: Test with an adapted example measurement


Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Helpers/Destroy Session Reservations.vi
pbirkhol-ni, line 1: You will need to clear error 5065 in case the customer called Unreserve on the diagram. I also wonder if we should be indexing the errors instead of using a shift register. Based on our current error implementation, the shift register will work the same as indexing. But if we change how error handling works, that may not always be the case.

image

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Helpers/Destroy Session Reservations.vi
jasonmreding, line 1: What happens if the user explicitly calls `Unreserve Sessions` or `Close Sessions` on the `SessionReservation`? It seems like the current code will generate an error and return that to the caller which doesn't seem correct. Ideally, we would be able to automatically update the `MeasurementContext` state when this happens. However, the most straight forward way to do that would add circular code dependencies between the libraries which I don't want to do. We could add an event notification via a generic interface, but that also seems like overkill. Maybe the simplest solution is to either add:
- Non-public APIs to query for valid reservation / sessions
- Non-public APIs that skip cleanup without throwing an error
- Just don't merge errors from these calls. That risks losing legitimate errors, though I'm having difficult thinking of scenarios where these methods should ever generate a legitimate error that should be forwarded to the user. However, we can't really predict what errors a driver might produce when closing the session.

pbirkhol-ni: >  It seems like the current code will generate an error and return that to the caller which doesn't seem correct. 

I mentioned this in my review above. For Close, we only close sessions that are no longer open. So there shouldn't be an error if a measurement calls close explicitly. If you called close using the driver API instead of the MeasurementLink API, then you might an error depending on the implementation of IInstrument, but maybe this is okay?

If the measurement called Unreserve, then the call to Close and Unreserve in the framework will throw error 5065. I suggested clearing that specific error as a remedy.

jasonmreding (changes requested):
At a minimum, we need to address the error behavior when the user explicitly closes sessions or calls unreserve on the `SessionReservation`. I would also like to come to a tentative agreement on auto cleanup of sessions.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Helpers/Destroy Session Reservations.vi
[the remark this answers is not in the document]
pbirkhol-ni: This is good feedback. @jonathanmendez, can you create a work item for this? We have gotten similar feedback from customers. I was thinking we would add an optional boolean to Create Sessions that will allow you to opt out of auto-close.

jasonmreding: My personal opinion is that requiring programmers to trap error codes is generally indicative of a bad API. It's not always practical to avoid this practice, but when it is, I generally prefer a different solution that is more readable, maintainable, discoverable, and intuitive.

pbirkhol-ni: An alternative to trapping an error is to check the state object of the reservation. If the state is unreserved, then skip calling close and unreserve.

I just thought of another complication, though. What if the user called unreserve but not close? The the cleanup code works now, we won't try to close the sessions. Maybe that is okay, though.

I'm becoming more and more convinced that we just shouldn't do auto cleanup. Is it really that hard for customers to do this themselves?

jonathanmendez: A further complication with attempting to auto-cleanup. The data lifetime for session reservations is controlled by the measurement logic VI. In testing, this means that the DVR stored as private data in the session reservation become invalid, and we don't have access anymore to the session information to close or unreserve.

I've been thinking about possible workarounds, but I haven't thought of anything workable/reasonable. This may be another reason to abandon auto cleanup.

jasonmreding: I'm guessing if we didn't use call by reference and instead had a virtual class method that was invoked directly that this wouldn't be an issue?

jasonmreding: Also, `Unreserve Sessions Impl` should probably explicitly destroy the DVR as well. Otherwise, if you come up with a solution to this problem, it might now result in a leak.

PR #331, opened 2023-10-11
Create NI-FGEN Standard Function Example

synesthetize:

Creates an NI-FGEN example to demonstrate the use of the new session management API with NI-FGEN, generates a waveform using standard function mode.

[AB#2533780](https://ni.visualstudio.com/DevCentral/_workitems/edit/2533780): Create Standalone MeasurementLink NI-Fgen Example

Able to run the example in InstrumentStudio and MeasurementUI Editor with a persistent simulated device.


synesthetize:
Measurement Configuration:
config

Measurement Results:
results

Measurement Logic:
measurement

FGEN Logic:
fgen logic

Measurement UI:
measui

InstrumentStudio:
instudio

MeasurementUI Editor:
uieditor

synesthetize:
I noticed that the driver definitions for enum-like controls (in this case waveform-type) are defined as rings types even if they satisfy the conditions for an enum (sequential ordering starting at val = 0). The LabVIEW service only supports auto-annotating enum types which means we won't be able to get auto-populating behavior if we use the driver defined type. We could still use the driver type but without the annotations it will show up in TestStand as a numeric instead of a ComboBox with selectable values and to get a ComboBox in .measui you would need to define a ring control for it in UI Editor.

For now in this example I've chosen to create my own enum type that maps to the values of the driver type but this is really just a workaround. We should  consider adding support for auto-annotating LabVIEW ring types if we want to support driver rings natively.

synesthetize:
I'm assuming we'll also want a TestStand .seq file as well. The existing DCPower example has some nontrivial helper VIs for setting up pinmap and sessions and I'm not sure where they came from. If our plan is to just copy these VIs into other examples then I can do that as part of this PR, otherwise I'll wait to get more info on how we want to handle that.

dixonjoel:
I'd suggest modifying the Document tab width and height so the white box around the controls is smaller.

synesthetize:
> I'd suggest modifying the Document tab width and height so the white box around the controls is smaller (in the .meaui).

Fixed:

Source/Example Measurements/NI-FGEN Standard Function/Measurement Service Helpers/Measurement Service Helpers.lvlib
jasonmreding, line 1: This comment applies to the entire Measurement Service Helpers directory/library. I thought the plan was to completely eliminate these helpers and instead only call public APIs we own from <vilib>. This means we'll eventually want to modify our code generator for a measurement. I think we have a work item to track that, but I'm not certain. In the meantime, I think you can just delete this entire directory from the project. It doesn't appear that it's being used anyway. 

@pbirkhol-ni Does this match your expectations as well?

pbirkhol-ni: I am reviewing Chris' example and have a similar comment. I think we should get rid of this library and I'll create a user story to get rid of it in the generator. If there is a VI that we are still using in the library, we should move it somewhere else.

jasonmreding:
> I'm assuming we'll also want a TestStand .seq file as well. The existing DCPower example has some nontrivial helper VIs for setting up pinmap and sessions and I'm not sure where they came from. If our plan is to just copy these VIs into other examples then I can do that as part of this PR, otherwise I'll wait to get more info on how we want to handle that.

There will be some discussions about API changes for TestStand code modules happening soon. Ultimately, I would like to centralize a lot of this code into a pin map service client library and the session management library. However, none of that work has been captured or planned yet so I'm not sure how much we'll get done before release. Hopefully there will be better guidance on this by the end of the week.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/subVIs/NI-FGEN Specific Logic.vi
jasonmreding, line 1: I can't remember, is this VI generated as a template as part of our code generation or is this an implementation detail specific to this example? I'm asking because I think ti would be better to update the connector pane to the 4 x 2 connector pane rather than 5 x 3 connector pane so it integrates better with the rest of the MeasurementLink APIs which use the 4 x 2 pattern. If it's part of the code generation, then we would need to update the template for that. Otherwise, you can just edit this VI.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/Measurement Logic.vi
jasonmreding, line 1: There is a PR posted that will add auto cleanup to the `MeasurementContext` which will automatically unreserve and close sessions when the measurement finishes. If we are going to move forward with this behavior, we should decide if our examples are going to promote this, or whether we think it is better to promote explicit cleanup and back stop that with automatic cleanup. This might be a decision in which to involve the product owners and lead users. FYI @pbirkhol-ni

jasonmreding (approved):

Source/Example Measurements/NI-FGEN Standard Function/Measurement Service Helpers/Measurement Service Helpers.lvlib
DelpireNI: I was able to successfully remove the entire directory/library from my example.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/subVIs/NI-FGEN Specific Logic.vi
synesthetize: After fixing up the connector pane:

dixonjoel: Oh, so much better. Those wire bends were bugging me.

Source/Example Measurements/NI-FGEN Standard Function/Measurement Service Helpers/Measurement Service Helpers.lvlib
synesthetize: Fixed in Commit 5

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction.measproj
pbirkhol-ni, line 1: I'm of the opinion that we shouldn't have a .measproj and a .measui for every example. Most LabVIEW measurements will use a LabVIEW. I think it is okay for just the DCPower example to have a .measui.

This would involve deleting the .measproj, deleting the .measui, and modifying Get UI Details.vi to only have Measurement UI.vi in the list of UIs.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/subVIs/NI-FGEN Specific Logic.vi
pbirkhol-ni, line 1: I think my preference would be to avoid subVIs and put everything on the Measurement Logic.vi diagram. For the 1.0 release, the examples used subVIs because the MeausurementLink code was so complicated that the diagrams become too big. However, now that the MeasurementLink API is simpler, I would put everything on the main diagram. For your example, I still would probably put the logic for detecting cancellation in a subVI since it is kind of gross.

pbirkhol-ni (approved):
As a general comment, make sure you are saving the front panel and block diagrams in the upper left of the screen. Ctrl-U is a shortcut to do this for front panels. I don't think the block diagram has a shortcut.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/subVIs/NI-FGEN Specific Logic.vi
synesthetize: Fixed and updated screenshots in my first comment

pbirkhol-ni:
Also make sure you try building the measurement into an exe and running it that way. You don't necessarily have to run it as a static measurement. You can just build it into an exe, run it, and then open it using InstrumentStudio.

synesthetize:
> Also make sure you try building the measurement into an exe and running it that way. You don't necessarily have to run it as a static measurement. You can just build it into an exe, run it, and then open it using InstrumentStudio.

I had to modify the build spec to include the additional files I added but I was able to open and run inside of InstrumentStudio after doing that.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/Measurement Logic.vi
synesthetize: This is now the last unresolved conversation. I was curious if there has been any discussion on this outside of this PR?

jasonmreding: I would submit examples with explicit cleanup for now. We're still trying to work through some of the implementation challenges to "auto cleanup". If we want to encourage use of "auto cleanup" later, we can do a second pass to update all examples for consistency.

PR #332, opened 2023-10-11
Support calling Get Connections before sessions have been created

pbirkhol-ni:

[AB#2543570](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2543570)
Using `Variant to Data.vi` with an empty variant input will always throw error -1 (input invalid). The `Get Connections from Pins.vim` calls `Variant to Data.vi` and therefore will error if the session in the `Connection with Session.ctl` hasn't been populated yet. The change in this PR will only call `Variant to Data.vi` if the session variant is non-empty. When the session variant is empty, it passes through the session from the **connection type** input. 

This new behavior matches what the python implemetation does. We could also consider throwing an error, but I like this solution better.

I am fixing this issue now because it is required for some of the Create Sessions and Close Sessions tests that I am writing.

Wrote a new test for this scenario.


Source/Runtime/MeasurementLink Session Management/Session Reservation/Get Connections from Pins.vim
pbirkhol-ni, line 1: image
image

jasonmreding (approved):

PR #334, opened 2023-10-12
Create NI-Digital SPI Example

WesleyTangNationalInstruments:

Creates a NI-Digital example that demonstrates how to use our new session management APIs with NI-Digital using the SPI shipping example based on existing standalone NI-Digital example for measurementlink-python.

[AB#2532132](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2532132)

Ran example successfully in InstrumentStudio and MeasurementUI Editor with a persistent simulated device.

image
image

### Open Issues

- Missing TestStand sequence files.
     - These will be added in a future PR.
- Redundant Pin Input
     - NI-Digital defines the list of pins that are used and what digital IO happens on those pins in the pattern file. Asking the user of the measurement to specify a pin list as input in order to reserve sessions is redundant as that information is already in the pattern. The problem is that the measurement developer is unable to extract the list of pins from the pattern without decompiling it, which can be a very heavy operation depending on the size of the pattern (on the order of minutes for large patterns). Internally, the .digipat is really just a .tdms file, so technically a measurement creator can peek inside and extract the pin list, but the fact that the .digipat is really a .tdms file underneath is an implementation detail that we do not share with external customers and that the driver reserves the right to change at any point. As such, doing so in a shipping example is not what we want to do so I've kept the redundant pin list input for this measurement.


Source/Example Measurements/NI-Digital SPI/Measurement Service Helpers/Measurement Service Helpers.lvlib
pbirkhol-ni, line 1: We have been getting rid of this library and all of its VIs in the other PRs. There is a work item to remove this library from the generator as well.

WesleyTangNationalInstruments: Removed Measurement Service Helpers.lvlib.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.sfp
```diff
@@ -0,0 +1,35 @@
...
+	<EnvoyManagerFile Id="6c31f1cca31b4b2fa7969d75081ad3bd" xmlns="http://www.ni.com/PlatformFramework">
+		<EnvoyManagerRootEnvoy Id="e948a5811cee45b3a3807dae45535633" ModelDefinitionType="EnvoyManagerRootEnvoy" Name="RootEnvoy" />
+		<NameScopingEnvoy AutomaticallyResolveUp="True" Id="78b3f18a9db84d6ea5f2e3e81919b7b2" Name="NIDigitalSPI.sfp" NameTracksFileName="True" />
+		<EmbeddedDefinitionReference Id="a4b06c4b696e4fc69fa02b8af0aa05a4" ModelDefinitionType="{http://www.ni.com/InstrumentFramework/ScreenDocument}Screen" Name="{http\://www.ni.com/InstrumentFramework/ScreenDocument}Screen.task">
+			<Screen ClientId="{4e624bc4-ee53-476d-8221-e267b8671382}" ConfigurationParameters="{&quot;@type&quot;:&quot;type.googleapis.com/ni.measurementlink.measurement.v1.MeasurementConfigurations&quot;,&quot;pinNames&quot;:[&quot;SPI_PINS&quot;]}" DisplayName="NI-Digital SPI (Py)" Id="7319682d2bea49af95dbd40688eb3193" PanelPresentation="ConfigurationWithVisualization" ServiceClass="ni.examples.NIDigitalSPI_Python" xmlns="http://www.ni.com/InstrumentFramework/ScreenDocument">
```
jasonmreding, line 17: The measurement panel appears to be pointing at a service class (the Python version) for a different service then this example provides. The snapshot of the UI captured in the sfp is also for a .measui. This isn't a big deal as it should auto update. However, if the service isn't install/running, then this is the default view of the UI that is displayed, and it doesn't match the current UI for the service.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
pbirkhol-ni: Also, the site pass/fail logic could be simplified with conditional indexing.

image

Source/Example Measurements/NI-Digital SPI/Pattern.digipat
jasonmreding, line 1: I don't have the digital driver installed to view these files, and I don't think I would know what I'm looking at anyway. You probably have the most expertise on this driver so I'm trusting you to do the right thing here.

jasonmreding (approved):

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement UI.vi
pbirkhol-ni, line 1: We've been standardizing on using NXG Style controls for UIs. As far as I can tell, most of your controls are classic controls.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.sfp
pbirkhol-ni, line 1: This .sfp isn't really working for me. It won't load when I open it. When I create a new .sfp, though, it seems to work okay.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement UI.vi
pbirkhol-ni, line 1: The pin type specialization isn't working for me and it seems like the issue is that the label and the caption don't match. The type specialization matches the caption, but it should match the label instead. I think this will also cause issues when updating the UI.

pbirkhol-ni (changes requested):

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.sfp
pbirkhol-ni: I tried to load this .sfp and it didn't really load correctly. We should just delete it and recreate it.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.measproj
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Removed as suggested.

Source/Example Measurements/NI-Digital SPI/Pattern.digipat
WesleyTangNationalInstruments: The digital pattern editor files are sourced from the measurementlink-python NI-Digital example with no additional modifications. I reviewed them and ensured that they were reasonable and I didn't find anything that was incorrect or problematic. 

The only thing to note is that it does differ slightly from Digital Pattern Editor's SPI shipping example. The files here are much simpler (and thus make creating a measurement and TestStand sequence around them simpler), but the shipping example is better documented and demonstrates more features of NI-Digital and Digital Pattern Editor. I guess it's really a choice between whether we want these examples to highlight how to use MeasurementLink while minimizing superfluous complexity versus having a fully-featured example that is very realistic and highlights as much as possible. Either way, since the python example already uses these files, it seemed fine to just maintain consistency here.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Get Type Specializations.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Fixed as suggested.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Refactored as suggested. I didn't really see the need for subVIs here so I consolidated this into "Measurement Logic.vi".

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement UI.vi
WesleyTangNationalInstruments: Replaced controls in "Measurement UI.vi" with NXG controls as suggested. I didn't realize there were multiple styles of the same control in LabVIEW.

WesleyTangNationalInstruments: Updated so that labels and captions for every control in "Measurement UI.vi" matches.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I thought about this more and I don't think moving the file paths into Create Sessions is the right thing to do here as users no longer have control over that setup process. The loading of these files is really an implementation detail of the setup process that shouldn't be a part of the Create Sessions API. 

This kind of stuff really falls into a category of "setup work for the driver that we don't want to execute multiple times". An example that also falls into this category is synchronization setup across multiple devices with NI-TClk and NI-Sync and that would not limited to just NI-Digital. NI-FGEN's arbitrary waveform generation is another example of something similar that requires loading a waveform file before being able to generate it.

It seems the problem is that we want a measurement with setup that can run both standalone and attached to sessions from TestStand. In TestStand, you would want to do this setup work as setup steps so that you can run multiple measurement setups without the overhead of setup every time, but when running the measurement standalone, the measurement itself would have to do this work as it hasn't been done yet.

_Example TestStand sequence with setup:_
image

I think the gap here is that we really need to provide the measurement author a way to say "only run this code if TestStand didn't already do it" and the previous API exposing `sessions_exists` gave the measurement an analog to that information (as we assume that if the sessions already exist, the setup work has already been done). I'm not advocating for exposing `sessions_exists` as I agree with the philosophy of wanting to distance users away from the notion of sessions as much as possible, but it was a convenient way to answer the question of "has setup already been done?" and without it we don't really have a great way of conditionally executing based on this.

I considered a few options, but I'm leaning towards adding a boolean input to the measurement configuration that is effectively "has setup been done yet?" and TestStand would always pass in true while it would default to false otherwise. Let me know what you think, @pbirkhol-ni .

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.sfp
WesleyTangNationalInstruments: Fixed by recreating this SFP file as suggested. This file was stale and I was actually updating a temporary NIDigitalSPI.sfp file I had on my desktop.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.sfp
WesleyTangNationalInstruments: Fixed by recreating this SFP file as suggested. This file was stale and I was actually updating a temporary NIDigitalSPI.sfp file I had on my desktop.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.lvproj
```diff
@@ -0,0 +1,126 @@
...
+			</Item>
+			<Item Name="niDigital_64.dll" Type="Document" URL="niDigital_64.dll">
+				<Property Name="NI.PreserveRelativePath" Type="Bool">true</Property>
+			</Item>
+			<Item Name="niDigitalLV_64.dll" Type="Document" URL="niDigitalLV_64.dll">
+				<Property Name="NI.PreserveRelativePath" Type="Bool">true</Property>
+			</Item>
+		</Item>
+		<Item Name="Build Specifications" Type="Build">
+			<Item Name="NI-Digital Measurement" Type="EXE">
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: I'm not very familiar with building in LabVIEW. It wanted me to specify a startup project so I specified "Run Service.vi". It builds successfully now but running the exe doesn't work. It flickers the "Run Service.vi" and then disappears (crashes maybe? the process hangs and I have to kill it manually) and we can't connect to it. I saw the source files were included if referenced so I tried to explicitly include them in the build, but that didn't seem to make a difference. I also tried to recreate the Build Specification, which also didn't seem to make a difference. I also looked at some of the public documentation but didn't find much guidance there. I also double-checked that running "Run Service.vi" from the lvproj does work as expected.

 @pbirkhol-ni can you give some guidance on how I should fix this?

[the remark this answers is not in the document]
dixonjoel: Did this build specification come from instantiating this service from the template? Or did you create it by hand? If the latter, you may want to compare it to one from the template.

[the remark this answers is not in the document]
dixonjoel: Or from another example project.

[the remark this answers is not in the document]
dixonjoel: E.g. Cancellable measurement
image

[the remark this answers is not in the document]
jasonmreding: Yeah, there are a bunch of always included VIs that are missing (VIs called via call by reference). I would start by copying from the template or one of the shipping examples and go from there. I'm not sure if buildspec configuration is persisted completely as XML in the lvproj file or if there is some binary data mixed in. If it's just XML, you could also try copy/pasting XML between projects by hand rather than relying on manual diffing to try and recreate everything. IIRC, I think I've seen mention of devs doing that in forums posts and generally being successful with it.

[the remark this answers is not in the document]
WesleyTangNationalInstruments: I updated the lvproj to build similar to CancellableMeasurement. The main difference was the missing files under "Always Included" and missing the Post Build Action. I'm hitting an issue where the Post-Build Action.vi is failing due to a path issue (either relative or missing). 

Where do the "Post-Build Action.vi"'s inputs come from? I opened the VI expected it to populate when the build it run, but it does not. I also set a breakpoint in that VI, but it doesn't seem to get hit so I'm having trouble seeing what the issue is.

image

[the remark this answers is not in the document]
dixonjoel: I'm not sure where the inputs come from. It looks like maybe they are filled in automatically by convention from the build specification.

This code is copying some grpc stuff. Do you have the grpc-labview packages installed?
image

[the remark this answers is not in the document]
dixonjoel: Looks like this Generate VI button fills in the front panel, so I think these are provided by the application builder framework.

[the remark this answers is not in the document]
dixonjoel: The only things I see it copying is those grpc dependencies and the .measui file so check those two things.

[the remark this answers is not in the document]
WesleyTangNationalInstruments: I have the grpc libraries installed and ensured that they are the latest versions from the latest release. I also had to copy over the latest `measurement-services-labview\Source\Runtime` into `vi.lib` for the latest session reservation stuff. 
image

The .measui and .measproj files were removed as part of a previous commit and this example uses a LabVIEW UI now. Recreating them and adding them back in did not seem to make a difference. 

I noticed during the build process, I get this warning.
image

I checked and the `Measurement\_functions_ni_measurementlink_service_2.mnu` does not exist so this might be the cause. Where do these files come from? Maybe I need to reinstall MeasurementLink?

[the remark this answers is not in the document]
dixonjoel: I recently added a link to that .mnu file. The way it works is that the class is set up to point to a file that is not around during development time, but the VI package installs that file so it links up correctly to the class. I don't think that would cause this problem, but it's possible. Are other examples able to build a .exe correctly even though they get that warning?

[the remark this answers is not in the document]
dixonjoel: To debug and figure out where the error is coming from, you can go to the Post-Build Action.vi in vi.lib and put some error dialogs in there or an error handler on some various parts of the build process.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
pbirkhol-ni: > This kind of stuff really falls into a category of "setup work for the driver that we don't want to execute multiple times".

I agree with this. Would an NI-Digital user ever not configure the file paths as shown in the example? If it is something that they are always going to do then it seems reasonable to put it in init. If it is something that they will only sometimes do (like arbitrary waveform or routing) then it doesn't belong in init. I don't know enough about NI-Digital, so I'll leave this up to you. I am okay with your alternative of putting a boolean as a measurement input; just don't expose that boolean on the InstrumentStudio UI.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.lvproj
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Paul helped me fix this this morning. The issue with "Post-Build Actions.vi" was that there was an empty element in "Get UI Details.vi".

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: > Would an NI-Digital user ever not configure the file paths as shown in the example?

Yeah, there are various ways a user may want to configure that isn't a single specs, single timing, single levels, and single pattern. 

Specs are entirely optional so a user may not have any specs to load at all. For timing/levels, it is possible that users choose to configure their timing and levels with the low-level API rather than using timing and levels files. This is generally an uncommon case as it is a pain to do it that way, but possible that some user out there may prefer using the low-level API rather than the files. For patterns, it is reasonable for a user to use NI-Digital productively without patterns such as when utilizing pins for the PPMU functionality (each digital pin can effectively act as a limited low-accuracy SMU). Additionally, a user may want to preload a single pattern, multiple patterns, or none at all depending on the size of their patterns, how many they have, and if there is sufficient vector memory available on the device to hold all their patterns simultaneously. 

On the other hand, pin maps will generally always be loaded. I have a hard time thinking of a case where a user finds it necessary to not have the pin map loaded, but I can't think of anything realistic. Maybe a user wants to manually reset their device, but then they can always just reload it. \shrug. I could see loading pin map being part of initialize, but I also don't see anything wrong with it being in the pin logic. Since it doesn't make sense for the other files, I think it makes sense to keep pin map alongside the other files.

I'll opt to add a boolean to conditionally load the files for now for all of these files.

Source/Example Measurements/NI-Digital SPI/README.md
```diff
@@ -0,0 +1,26 @@
+## NI-Digital SPI
+
+This is a MeasurementLink example that tests an SPI device using an NI Digital Pattern instrument.
+
+### Features
+
+- Uses the `nidigital` package to access the NI-Digital Pattern Driver from LabVIEW
+- Pin-aware, supporting one session, multiple pins, and multiple selected sites
+- Includes project files for Digital Pattern Editor, InstrumentStudio, MeasurementLink UI Editor
```
pbirkhol-ni, line 9: The MeasurementLink UI Editor project should have been removed.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Results.ctl
pbirkhol-ni, line 1: Make sure all of the controls are NXG controls. The new boolean you added is a "Modern" control. Which brings up another question. Should you even have the Load Files boolean on the InstrumentStudio UI? Is there ever a situation where you would not want to load the files when running from InstrumentStudio? If so, it should default to the most appropriate value, which would be to load the files.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.lvproj
pbirkhol-ni, line 1: On the Advanced section of the build spec, can you click the "Use fast file format" option? We are checking that value for all of our new build specs (the generator has already been updated so that newly created projects have it selected by default).

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement UI.vi
pbirkhol-ni, line 1: I would consider expanding the path controls so that they show more of the path. I don't know how many characters is right, but right now I only see C:\dev\ before it cuts off.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
pbirkhol-ni, line 1: In comment number 5 you mention the session manager and discovery service clients. I don't think we should mention those anymore since we are basically hiding that from customers. You can simply say "Unreserve the session".

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Results.ctl
pbirkhol-ni: On the other hand, in TestStand, you would want the default to be False. So maybe the best option is to set the default in the UI to be true and then hide the control.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
pbirkhol-ni, line 1: The site selector is not being respected. I really think we should be disabling all sites (by passing an empty string) and then only enabling the sites that are in the pin map context. I realize that this may be against NI-Digital best practices, but it makes me very uncomfortable to be generating signals on a site that wasn't explicitly specified by the customer.

pbirkhol-ni (changes requested):

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Logic.vi
WesleyTangNationalInstruments: Updated as suggested. I forgot an empty string works for disabling all sites. I had thought the fact that the `site list` parameter as required on the Disable Sites VI meant that we had to explicitly provide values in the LabVIEW API so we couldn't do this without knowing the list of all sites.

WesleyTangNationalInstruments: Fixed comment as suggested.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement UI.vi
WesleyTangNationalInstruments: I made them longer by roughly 2.5x and 3x taller. Since these are file paths, it really depends on the user's file system how big it required to display the entire path, but this displays the full path with the file locations on my system.

Source/Example Measurements/NI-Digital SPI/NIDigitalSPI.lvproj
WesleyTangNationalInstruments: Fixed as suggested.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/Measurement Results.ctl
WesleyTangNationalInstruments: Switched to using a NXG Checkbox instead and set the default in the configuration to true as suggested. Removed the control from the UI. TestStand should still be able to pass in false when loading the files.

Source/Example Measurements/NI-Digital SPI/README.md
WesleyTangNationalInstruments: Fixed as suggested. Forgot to update the readme when removing it.

Source/Example Measurements/NI-Digital SPI/README.md
```diff
@@ -0,0 +1,26 @@
+## NI-Digital SPI
+
+This is a MeasurementLink example that tests an SPI device using an NI Digital Pattern instrument.
+
+### Features
+
+- Uses the `nidigital` package to access the NI-Digital Pattern Driver from LabVIEW
```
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Reworded to "Uses the NI-Digital LabVIEW API to access the NI-Digital Pattern Driver from LabVIEW".

pbirkhol-ni (approved):

PR #335, opened 2023-10-12
RFC: Enable opt-in auto-unreserve by storing copy of session infos

jonathanmendez:

Prototype that stores copy of session infos in the measurement context to enable auto-unreserve cleanup. Does so with opt-in behavior when doing Reserve Sessions on the measurement context.


Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Destroy.vi
jasonmreding, line 1: This still doesn't really handle the case where the user explicitly calls `Session Reservation:Unreserve Sessions`. I guess that's why you added a Boolean input to the API? I'm not a huge fan of the new Boolean since:
- Ideally, we would just always cleanup.
- Nothing prevents the user from explicitly cleaning up when auto cleanup is set to True. This could technically lead to issues with sessions getting unreserved unexpectedly.

The only other ideas I've come up with so far are:
1. Create some DVR allocator singleton for reservations that is always running. Maybe a FGV would work as well? Anyway, this object would get created / initialized from within our service code and a reference would get passed during creation of the session reservation. There are probably some things from actor framework that could be borrowed for this, but I'm not really that familiar with what is available since I haven't personally used actor framework before.
2. Define an interface for use by `Session Reservation` that is implemented by `MeasurementContext`. This interface is passed to the creation of the `Session Reservation` and is effectively an event observer. When sessions are reserved/unreserved on `Session Reservation`, it synchronously calls this interface to notify the observer of the change. You could accomplish the same thing with user defined events, but I think a synchronous callback is much easier to deal with than an asynchronous callback using an event structure.
3. Just have `Session Reservation` call directly into `MeasurementContext` when sessions are reserved/unreserved, and live with the circular dependency.

The last option is the simplest implementation, but it violates good architectural principles. The second option seems like the cleanest architectural solution, but it seems like a lot of hoops to jump through for something that seems like it should be simpler.

pbirkhol-ni: Is it an error to unreserve a session that has already been unreserved? If not, then what Jonathan is proposing could work, although it might involve more gRPC calls than are necessary.

I also like Jason's option 2. It seems like it would be straight forward to implement. I don't think it solves the auto cleanup of sessions, though, if that is something that we still want to offer.

I'll be out the Friday and Monday, but I think someone should talk to Cristina, Ryan, Jack, and/or Vidya to see how important auto unreserve and auto cleanup are. It seems like something we could add later without breaking compatibility, so maybe we can just wait to see if anyone requests it. My gut instinct is that no one will since calling close VIs is natural and expected in LabVIEW. Maybe we can offer a single VI (in addition to Close and Unreserve) that does both Close and Unreserve.

jonathanmendez: > This still doesn't really handle the case where the user explicitly calls `Session Reservation:Unreserve Sessions`. I guess that's why you added a Boolean input to the API? I'm not a huge fan of the new Boolean since:
> 
>     * Ideally, we would just always cleanup.

I think always cleaning up probably makes sense, but since that idea wasn't fully resolved and it was easy to set up, I wanted to prototype what the opt-in behavior would look like.

>     * Nothing prevents the user from explicitly cleaning up when auto cleanup is set to True. This could technically lead to issues with sessions getting unreserved unexpectedly.

> Is it an error to unreserve a session that has already been unreserved?

It is indeed intentionally not an error to unreserve a session that has already been unreserved, under the idea that an unreserve request is satisfied (the session ends up in the "not reserved" state) either way. From that perspective, if the user decides to explicitly unreserve, that's not a problem for auto-cleanup.

> 
> The only other ideas I've come up with so far are:
> 
>     1. Create some DVR allocator singleton for reservations that is always running. Maybe a FGV would work as well? Anyway, this object would get created / initialized from within our service code and a reference would get passed during creation of the session reservation. There are probably some things from actor framework that could be borrowed for this, but I'm not really that familiar with what is available since I haven't personally used actor framework before.
> 
>     2. Define an interface for use by `Session Reservation` that is implemented by `MeasurementContext`. This interface is passed to the creation of the `Session Reservation` and is effectively an event observer. When sessions are reserved/unreserved on `Session Reservation`, it synchronously calls this interface to notify the observer of the change. You could accomplish the same thing with user defined events, but I think a synchronous callback is much easier to deal with than an asynchronous callback using an event structure.
> 
>     3. Just have `Session Reservation` call directly into `MeasurementContext` when sessions are reserved/unreserved, and live with the circular dependency.
> 
> 
> The last option is the simplest implementation, but it violates good architectural principles. The second option seems like the cleanest architectural solution, but it seems like a lot of hoops to jump through for something that seems like it should be simpler.

My main concern with any of these options is a notion that the `Session Reservation` class so far has been designed in a way that feels useful by itself. Certainly option 3 directly ties the `Session Reservation` to the `MeasurementContext` in violation of that design. Options 1 and 2 don't have that direct connection, but they add architecture that seems out of place except for the the purposes of `MeasurementContext`. Why would a `Session Reservation` class need to be passed a memory allocator to be able to reserve sessions? Why would a `Session Reservation` class need event observers? Of these, the latter seems least objectionable but still odd.

Comparing my prototype (grabs copies of the `Session Infos` when using the `MeasurementContext` to make a `Session Reservation`) to the event observer (stores `Session Reservations` when reserved or discards them when unreserved?), it feels like they both can accomplish one result (auto-unreserve), while only the latter can handle another (auto-close of sessions). The former has the cost of potentially extra grpc calls (practically less likely if we clearly message that the MeasurementContext will auto-unreserve for you). The latter has the cost of an extra piece of architecture for the `Session Reservation` class.

jasonmreding: > It is indeed intentionally not an error to unreserve a session that has already been unreserved, under the idea that an unreserve request is satisfied (the session ends up in the "not reserved" state) either way. From that perspective, if the user decides to explicitly unreserve, that's not a problem for auto-cleanup.

The implementation here goes directly to the service to unreserve and not a client side cache. This means the service can receive multiple unreserve calls for the same session reservation. While the service might not treat this as an error, it opens up theoretical race conditions that could lead to hard to diagnose behavior. What if another measurement call reserves a session for the same resources after the user called unreserve. Then we execute our auto cleanup code which tries to unreserve again. However, now rather than a harmless noop, we're actually unreserving another measurement call's reservation.

jonathanmendez: > > It is indeed intentionally not an error to unreserve a session that has already been unreserved, under the idea that an unreserve request is satisfied (the session ends up in the "not reserved" state) either way. From that perspective, if the user decides to explicitly unreserve, that's not a problem for auto-cleanup.
> 
> The implementation here goes directly to the service to unreserve and not a client side cache. This means the service can receive multiple unreserve calls for the same session reservation. While the service might not treat this as an error, it opens up theoretical race conditions that could lead to hard to diagnose behavior. What if another measurement call reserves a session for the same resources after the user called unreserve. Then we execute our auto cleanup code which tries to unreserve again. However, now rather than a harmless noop, we're actually unreserving another measurement call's reservation.

Oof, yeah that's a good point. It makes me wonder if we should do something like add a client token to session reservations, though that's probably a deferable concern as long as we don't actively introduce client-side problems like this.

> > My main concern with any of these options is a notion that the Session Reservation class so far has been designed in a way that feels useful by itself.
> 
> I would expect the use of the observer interface would be optional and not required. The default interface implementation would just be all noops. I don't know if that allays your concerns or not.

I figured the observer would be optional. It just leaves it feeling odd to me as opposed to problematic, at least for session reservation where we can store copies of the data. I'm not sure I see if it'd be able to handle session closing, however, since the data lifetime for the session refnums would still belong to the measurement logic VI hierarchy. Maybe you aren't proposing this as a way to auto-close, but we should circle back with folks as Paul suggested to see what auto cleanup behavior is important for us to address.

jasonmreding: > Maybe you aren't proposing this as a way to auto-close, but we should circle back with folks as Paul suggested to see what auto cleanup behavior is important for us to address.

I was mostly thinking of this as a solution to auto unreserve and not auto close of sessions. Auto close of sessions is more complicated because of different driver refnum behavior to cleanup in LV. Also, not all driver APIs will be based on refnums (e.g. third party drivers). 

For grpc-device based drivers, I'm not sure if the LV refnum cleanup even matters. I would guess it doesn't matter for MI drivers, but we would have to test it.  I suspect the refnum is just a facade around a string in that case, and that string is still valid on the server. I guess it depends on how the session name is tracked within the session refnum. I can't remember if the refnum wire carries the resource name in this case or the session name. If it's the session name, then I wouldn't be surprised if everything still works even though LV has "cleaned up the refnum". I guess if the driver cleans up the underlying grpc channel on the client side as part of refnum cleanup, that might still be a problem. Ultimately, I don't yet know enough about how the drivers are implemented in this regard to feel confident about any solution.

jonathanmendez: I implemented a quick and dirty version of the observer implementation in #340

PR #336, opened 2023-10-13
Add sub-palette for 'Context' i.e. MeasurementContext and link it to the class

dixonjoel:

Put the MeasurementContext VIs in a subpalette under MeasurementLink. This is to facilitate sharing of that sub-palette in the Functions >> Measurement I/O to be the palette shown on right-click of a MeasurementContext wire.

image

Give the MeasurementContext class wire a specific palette instead of the default generated one.

Interactive use of the palette with a MeasurementContext wire.


Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
```diff
@@ -2,6 +2,7 @@
 <LVClass LVVersion="20008000">
 	<Property Name="NI.Lib.ContainingLib" Type="Str">MeasurementLink Measurement Server.lvlib</Property>
 	<Property Name="NI.Lib.ContainingLibPath" Type="Str">../../../MeasurementLink Measurement Server.lvlib</Property>
+	<Property Name="NI.Lib.DefaultMenu" Type="Str">_functions_ni_measurementlink_service_2.mnu</Property>
```
jasonmreding, line 5: Did you just end up making this edit by hand after installing the built VIPM? I didn't see any changes in the VIPB XML that looks like it associating this palette with a class. Does the built VIPM actually include a copy of this menu file next to the class where the menu is also a member of the class library? Do we know how stable this generated menu name is?

dixonjoel:
@jasonmreding @pbirkhol-ni The strategy I'm going for now is just to fully define a single palette set in the .vipb. This will be the 'MeasurementLink' palette. Any sub-palette that we want to share or reuse should be a sub-palette with only those reusable things. In the .vipb, we won't use the 'Add palette to library or class' at all. We will simply make the class or library point to the location where the package *will* install the palette file. Hopefully that makes sense.

The sharing / reuse of the MeasurementContext palette is the reason I made a sub-palette for its contents under the 'MeasurementLInk' palette.

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/MeasurementContext.lvclass
dixonjoel: I installed a locally built version of the vi package with these menus and then edited the MeasurementContext.lvclass to add that appropriate .mnu file to the class and point to it as the default palette. So no hand editing.

dixonjoel: The build installs the .mnu to <LV>\menus\Categories\Measurement\<subdirectory>. The class points to it in that location, so no there's no .mnu file next to the installed class.

I don't yet know how stable the .mnu name is, but I wouldn't expect it to change if we just add other palettes in the .vipb. It probably will change if we significantly edit the hierarchy or anything. But I guess we can just watch out for this and see how much of a hassle it is.

jasonmreding: > I installed a locally built version of the vi package with these menus and then edited the MeasurementContext.lvclass to add that appropriate .mnu file to the class and point to it as the default palette. So no hand editing.

Yeah, that's what I meant. I was really trying to get at whether VIPB made this change for you or whether you had to make a manual change so I can get a feel for how brittle or magical this approach is.

> The build installs the .mnu to \menus\Categories\Measurement<subdirectory>. The class points to it in that location, so no there's no .mnu file next to the installed class.

That's what I was wondering, but I don't see anything in the XML that has path information (just a file name). Does the class code search menus directory to resolve this? That seems kind of scary as it implies all mnu files have to be named uniquely throughout the hierarchy. Either that or the path information is stored somewhere in the binary information...

dixonjoel: Oh, the path for this item is down on Line 65.

dixonjoel: @jasonmreding Did you see the path down there? Is this PR ready to go then?

dixonjoel: Ok. I'll go ahead with this since I don't see a way to share the common MeasurementContext palette with the other approach.

jasonmreding: The other approach would be to have the class on its own mnu/palette for the right click context palette. The create a second set top level palette for the functions palette. The functions palette would then link to the existing class palette as a sub-palette. I haven't tried this myself, but I believe it's possible. You still have to deal with copying everything to vilib so that the pathing works out right, but from my understanding, you also have to do that with the approach used in this PR.

PR #337, opened 2023-10-16
Adding Register, Unregister, and Reserve Registered Sessions APIs

jasonmreding:

Adds the following public APIs to the Session Reservation class
1. Reserve Sessions - This is now a poly VI. The existing VI was renamed to the "NPin" poly instance. The "1Pin" instance, just calls the "NPin" variation.
2. Reserve Registered Sessions
3. Register Sessions
4. Unregister Sessions - This VI executes regardless of error in

In addition, APIs that can be called in parallel from a measurement service that can make gRPC calls have been updated to have reentrant execution with shared clones. APIs that are expected to be called from code modules are still marked as non-reentrant since parallelism isn't an expectation in this context. This resulted in a lot of non source code changes to the VIs.

Reentrant APIs:
- Reserve Sessions VIs (Both Measurement Context and Session Reservation classes)
- Unreserve Sessions VIs (Both Session Reservation and IState classes)
- Create Sessions VIs (Both Session Reservation and IState classes) - includes dispatcher and implementation VIs
- Close Sessions VIs (Both Session Reservation and IState classes) - includes dispatcher and implementation VIs

Non-Reentrant APIs:
- Reserve Registered Sessions
- Register Sessions
- Unregister Sessions
- Get Connections

These APIs simplify code module development which will make writing examples easier.

None at this time.


dixonjoel:
At least Reserved State.Unregister Sessions.vi was not close to the top-left of the screen. Please make a pass to fix this.

dixonjoel (approved):

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Registered Sessions.vi
pbirkhol-ni, line 1: This poly VI doesn't have documentation. This doesn't matter when hovering over the poly VI on the block diagram because it will show the context help of the selected instance VI. It matters for the palette, though, because context help will not show up unless the poly VI has context help.

Get Connections has this same problem.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions - 1Pin.vi
pbirkhol-ni, line 1: I don't know how I feel about `pin or relay name` being required for the `1Pin` instance, but `pin or relay names` being recommended for the `NPin` instance. I understand why you would do it that way, because if the input was recommended in both cases, then both instances would behave identical when unwired. 

It leads to strange behavior, though. Try doing this. Drop the poly VI, which should default to the `1Pin` instance. Then right click on the `pin map context` input and create a constant. You will notice that the poly instance changes to `NPin`. I assume it does this because the `NPin` instance isn't broken. However, that is unintuitive and, in my opinion, undesirable behavior.

FWIW, Get Connections has a similar problem. `pin or relay name` is recommended for both the `1Pin MConn` case and the `NPin MConn` case. This means that when unwired, both of those poly instances have identical behavior.

This is a long comment just to say that I think you should make the `pin or relay name` input recommended for the `1Pin` instance.

pbirkhol-ni (approved):
Some general feedback:
- You should Ctrl-U all of the new VI front panels to arrange the controls/indicators and to save in the upper left.
- I hope there is going to be a followup PR with some tests. You should be able to use the session manager directly if that helps you write tests. No need to mock it.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions - 1Pin.vi
jasonmreding: I made it required because you have to wire a valid pin name in order not to get an error when running the VI. An empty array is valid, but an empty/blank pin name isn't valid. If that's the behavior you want, then use the NPin poly instance.

I hadn't considered auto adapt oddities because of the required input. I wasn't aware LV would adapt the selection based on wiring inputs that are identical between the instances based on which instances are broken or not. I suspect LV didn't really realize this would happen either, but I understand the dilemma they are in. If we think this behavior is too weird/confusing, we could turn off adapt to type for this poly instance. I don't know if enabling adapt to type for some poly VIs but not other poly VIs in the same API would be equally confusing or not. So I think the options are:
1. Live with the current behavior
2. Disable adapt to type for this poly VI
3. Disable adapt to type for all poly VIs

jasonmreding:
> At least Reserved State.Unregister Sessions.vi was not close to the top-left of the screen. Please make a pass to fix this.

This is what it looks like on my machine. What am I missing?

image

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Registered Sessions.vi
jasonmreding: Updated reserve sessions documentation. I think I got fooled into thinking poly VIs couldn't be documented since right clicking on the icon to get at VI properties doesn't work. You have to right click the project item or go the files menu instead.

jasonmreding:
> * I hope there is going to be a followup PR with some tests. You should be able to use the session manager directly if that helps you write tests. No need to mock it.

That is my plan. I want to unblock example development as soon as we can though.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions - 1Pin.vi
pbirkhol-ni: The 4th option is what we do for the other new APIs, which is to treat an empty pin name as meaning "All Pins". We even have a helper VI in the Session Reservation class to help with this: `Pin to Pin Array.vi`, which will check for an empty string and then convert it to an empty array to pass to the `NPin` poly instance.

If you don't like the 4th option, then I will begrudgingly choose option 1.

pbirkhol-ni:
> > At least Reserved State.Unregister Sessions.vi was not close to the top-left of the screen. Please make a pass to fix this.
> 
> @dixonjoel This is what it looks like on my machine. What am I missing?
> 
> > You should Ctrl-U all of the new VI front panels to arrange the controls/indicators and to save in the upper left.
> 
> @pbirkhol-ni I do try and do this although it's possible I forgot on some of them. I've reviewed all of the new VIs, and they all appear to be laid out correctly to me. Can you reference specific examples where that's not the case? I'm worried there is some weird LV thing going on unless I'm just crazy.
> 
> image

The main thing I noticed is that the front panels seemed to be saved in the middle of my screen. So while the layout seemed fine, Ctrl-U will also move the front panel to the upper left of the screen.

dixonjoel:
> > At least Reserved State.Unregister Sessions.vi was not close to the top-left of the screen. Please make a pass to fix this.
> 
> @dixonjoel This is what it looks like on my machine. What am I missing?
> 
> > You should Ctrl-U all of the new VI front panels to arrange the controls/indicators and to save in the upper left.
> 
> @pbirkhol-ni I do try and do this although it's possible I forgot on some of them. I've reviewed all of the new VIs, and they all appear to be laid out correctly to me. Can you reference specific examples where that's not the case? I'm worried there is some weird LV thing going on unless I'm just crazy.
> 
> image

Sorry, I should have been more clear. I was looking at the block diagram of Unregister Sessions.vi and it was far to the right of my monitor - I could barely grab it and relocate. I don't know if we have a standard practice of where to put block diagrams.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions - 1Pin.vi
jasonmreding: OK, so I went with option 5 which is to effectively punt on this issue and not offer a poly VI for reserving from a single pin right now. This matches the service API, and the primary use from within a code module will always want to reserve for all pins for a given instrument type. I did keep the file name as if it were part of a poly VI in case we want to add a poly VI at some point in the future. However, I can change that if you think it's too weird.

PR #340, opened 2023-10-16
RFC: Use observer interface to track session reservation and unreservations for auto-unreserve

jonathanmendez:

Builds on #335 to observe when session reservations and unreservations happen, so that we don't accidentally unreserve sessions belonging to a different client, per the race condition [called out here](https://github.com/ni/measurementlink-labview/pull/335#discussion_r1358405678).


Source/Runtime/MeasurementLink Session Management/Session Reservation/Unreserve Sessions.vi
jasonmreding, line 1: I find it a little odd the unreserve callback is fired before we actually unreserve. Similarly, I would try to avoid firing the callback from inside the DVR for similar reasons you generally try to avoid holding locks while firing callbacks in other languages.

Source/Runtime/MeasurementLink Measurement Server/Reservation Observer/Reservation Observer.lvclass
jasonmreding, line 1: My initial thought is that we should just implement the interface directly from `MeasurementContext` to simplify things. However, maybe there is some value in hiding this detail from users of the class? Is there a particular reason you chose to implement it this way?

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Destroy.vi
jasonmreding, line 1: This should check for empty array on info to unreserve before calling unreserve on the session management service.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions.vi
jasonmreding, line 1: FYI, I am renaming this VI in [this PR](https://github.com/ni/measurementlink-labview/pull/337). That PR also makes the API public, and I don't think we should be adding this to the public API. Instead, we should have a community scoped API overload that is only accessible from `MeasurementContext`.

Source/Runtime/Measurement Server.lvproj
jasonmreding, line 1: The general approach looks reasonable here. I assume map lookups using the `SessionReservation` class still work as expected even though the DVR within the class is no longer valid? I don't really see why it wouldn't, but I thought I would double check. Due to the RFC tag, I only looked at the larger pieces. I will probably have additional cosmetic level feedback before submitting. Please reset my review status if you convert this to a non-RFC PR for submission.

jasonmreding (approved):

Source/Runtime/MeasurementLink Session Management/Session Reservation/Unreserve Sessions.vi
jonathanmendez: > I find it a little odd the unreserve callback is fired before we actually unreserve.

I agree it's a bit odd. I originally felt the need to set it up to happen before the unreserve since I anticipated using data of the session reservation in the callback, which unreserve will destroy. Putting it afterwards with the current signature means passing basically a bad object to the callback.

Putting the callback before the unreserve doesn't fully avoid the problem either, as someone could call `Unreserve Sessions` on an already unreserved reservation - the state will later throw an error but in the meantime we've passed a bad object to the callback.

This all doesn't practically cause issues in the current implementation since all the callback does is treat the session reservation object as a key to remove from a map. But I don't like the design.

I think an alternative that would work is, instead of storing a map<session reservation, session info[]> we could just store a set of session info
  - Assumes (currently correctly) that there wouldn't be overlap of any session infos across session reservations
  - Lets the callbacks happen after successful reservation/unreservation (including outside the DVR lock)
  - Theoretically more work to adapt/extend in the future if we want more smarts in the callback related to the session reservation itself and not just the session infos.. but that's just more work, not something we'll have prevented, so YAGNI says not to worry about it

Source/Runtime/MeasurementLink Measurement Server/Reservation Observer/Reservation Observer.lvclass
jonathanmendez: I didn't implement it separately for any particular reason. I was in prototype mode and so the first thing my mind went to is "I need an implementation of this interface, let's create one". So, I think having `MeasurementContext` do the implementation seems fine, unless we discover some scoping issue we care about and can't work around.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Reserve Sessions.vi
jonathanmendez: Having a separate community-scoped overload seems reasonable enough, and will let us avoid your dreaded `optional` terminal that I intentionally used to here to make it more hidden 😄

Source/Runtime/Measurement Server.lvproj
jonathanmendez: Yes, I did test that the map lookups with `SessionReservation` still work even if it happens after the DVR has been cleared out (although as per the other discussion, we may not stick with the map after all).

PR #343, opened 2023-10-16
Adding pin map client library

jasonmreding:

Adds pin map library to centralize logic for TestStand integration and provide higher level starting point around grpc-labview generated code for the pin map service.

Makes it easier to write TestStand code modules and single sources logic for examples we intend to ship.

None at this time since this work is gating example development. Will add tests at a later time as needed.


Source/Runtime/MeasurementLink Pin Map/Resolve Pin Map Path for Sequence.vi
jasonmreding, line 1: As best as I can tell, this VI appears to be a wrapper around the TestStand `FindFile` function that I have also seen used within TestStand expression step types in some of our examples. That would imply we're doing the lookup/resolve work twice. Does anyone know if this is true or what the preferred usage should be? I'm wondering if this API should be a private implementation detail rather than a public API.

dixonjoel (approved):

dixonjoel:
If you haven't built and installed the runtime package to verify this is installed correctly and not broken, you should consider doing that.

Source/Runtime/MeasurementLink Pin Map/Resolve Pin Map Path for Sequence.vi
pbirkhol-ni: This VI gets the absolute path to the pin map file, which we currently also use as the pin map id. Then later in the code module, we store the pin map id in a temporary global so that we can look it up later without having to recompute it. My understanding is that we do this so that we don't have to resolve the pin map id multiple times.

My opinion is that the VIs that use a TestStand sequence shouldn't be in this library. I view this library as a way to simplify access to the pin map service. What happens when we create a TestStand step that configures the pin map and registers sessions? We'll still have these TestStand sequence VIs in the library and won't be able to remove them without breaking compatibility. 

Maybe consider putting the Query Resource Access Information VIs in one library and the code modules helpers in another library?

Source/Runtime/MeasurementLink Pin Map/Query Resource Access Information - 1Pin 1Rsrc.vi
pbirkhol-ni, line 1: `resource` is spelled wrong in the context help.

Source/Runtime/MeasurementLink Pin Map/Query Resource Access Information.vi
pbirkhol-ni, line 1: I am generally against adding this poly VI until it is requested of us. We haven't spent much time thinking about this and if it is how we would even want to expose the pin map information to customers. For session management, we don't really expose the channel mappings because it isn't always necessary for customers and they found it too complicated. If we were to release this, we would have to support it forever. I'd prefer to make sure we're doing it the right way, and as we found out with session management, that isn't always obvious at first glance.

pbirkhol-ni (changes requested):

pbirkhol-ni:
I didn't do a thorough review of this PR. I want to nail down exactly what we should include in this release and what we shouldn't include.

Source/Runtime/MeasurementLink Pin Map/Resolve Pin Map Path for Sequence.vi
jasonmreding: OK, I've renamed the library to `MeasurementLink TestStand Integration` and removed the pin query APIs. I'm submitting the remainder of the APIs as public for now. We can go back and make some of them private if we determine a good way to reference VIs in <vilib> directly from TestStand with being locked to a specific absolute path or LV version.

Source/Runtime/MeasurementLink Pin Map/Query Resource Access Information - 1Pin 1Rsrc.vi
jasonmreding: VI no longer exists.

Source/Runtime/MeasurementLink Pin Map/Query Resource Access Information.vi
jasonmreding: VI no longer exists.

Source/Runtime/MeasurementLink TestStand Integration/Set Pin Map ID for Sequence.vi
pbirkhol-ni, line 1: You should remove the underscores from the `pin_map_id` input.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink TestStand Integration/Set Pin Map ID for Sequence.vi
jasonmreding: I thought I had gotten rid of all of those, but I guess I missed one.

PR #349, opened 2023-10-18
Adding support to register measurement service on specific  version

svm-ni:

Modifying the code to explicitly specify the version of the Measurement Service, which is not supported currently, allows users to register services based on their specific needs, ensuring they can choose the desired service version.

- Renamed MeasurementLinkServerUserEvents.ctl to MeasurementLinkServerUserSettings.ctl
- A string array has been introduced in "MeasurementLinkServerUserSettings.ctl" to accommodate the Measurement Service Version.
- "Create Measurement Services.vi" has been modified as follows:
   - A new control for specifying the Measurement Service Version has been added to the Front Panel.
   - When the array is empty, all supported Measurement Service classes will be utilized.
   - If the array contains version information, the corresponding Measurement Service will be used based on the provided version.
- The "Run Service.vi" has been updated to unbundle the Measurement Service Version and Pass it to the "Create Measurement Services.vi."

Run Service.vi
image

image

Create Measurement Services.vi
- Measurement Service Version Array Empry.
  All supported Measurement Service classes will be utilized.
  
image

- Measurement Service Version Array not Empty:
 Measurement Service will be used based on the provided version.
 
image

By implementing this modification, users can explicitly designate the version of the Measurement Service, a feature not currently supported. This enhancement empowers users to register services according to their unique requirements, enabling them to select the preferred service version.

The existing test cases with respect to the discovery service were executed and they passed.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Create Measurement Services.vi
pbirkhol-ni, line 1: Can you uncheck the `Show as Icon` option for the new control?

pbirkhol-ni, line 1: I think we can make this slightly more maintainable by removing the case structure that you added. Instead, if the `Measurement Service Version` is empty, create an array with all of the supported services. I've included a screenshot below of what this might look like:
image

pbirkhol-ni, line 1: For the default case, consider throwing an error instead of returning the v2 interface. You could also consider using an enum instead of a string. Then we would be prevented from providing an bad value.

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
pbirkhol-ni, line 1: Make sure the front panel of this VI is saved exactly how it was before. This is a user visible VI. This is how I see it after your changes:
image

It should look like this. Notice that there is more space between the bottom of the stop button and the edge of the window.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/TypeDefs/MeasurementLinkServerUserSettings.ctl
pbirkhol-ni, line 1: The new control should be plural: Measurement Service Version**s**

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
pbirkhol-ni, line 1: You should unbundle the Stop Serve and the Measurement Service Versions in the same unbundle node. That will clean up the diagram.

pbirkhol-ni (changes requested):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Create Measurement Services.vi
jasonmreding: FYI, there is also a right click configuration on the case structure to perform case insensitive comparisons which will clean things up a little more. The case structure displays a glyph in the lower left when configured in this mode.

Source/Runtime/MeasurementLink Measurement Server/Helpers/TypeDefs/MeasurementLinkServerUserSettings.ctl
jasonmreding: @pbirkhol-ni I can't remember when we added this input terminal, but doesn't this rename break source compatibility of the API? I guess we don't really expect any current customers to have it wired so maybe we're OK with that. However, doesn't this break our test suite?

pbirkhol-ni: Yes, it breaks source compatibility and it breaks our test suite. Since this new input is being added for our test suite, I'm not concerned about that breakage. Sharath will fix it after our next measurementlink-labview release. We considered adding a new input to `Run Service.vi` to not break compatibility for customers, but I am fairly confident that we are the only clients so I am not too concerned with breaking compatibility. You could also make an argument that this input is not an official part of our API; instead, it's a test seam and we won't guarantee compatibility.

jasonmreding: I think that's a reasonable position for now. My main concern is that we recognized the test quite breakage and are planning accordingly rather than reacting.

jasonmreding (approved):

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
svm-ni: Updated the code

Source/Runtime/MeasurementLink Measurement Server/Helpers/Create Measurement Services.vi
svm-ni: Modified the control to utilize enums instead of strings.

When the array is empty, a constant value is passed, which will register both v1 and v2. If the array is not empty, the appropriate measurement service version case will be applied.

svm-ni: I've revised the code to utilize enums, ensuring that no invalid values are passed.

svm-ni: Done

pbirkhol-ni, line 1: You need to make the enum a typedef. Then replace all instances of the enum with the typedef.

pbirkhol-ni, line 1: Since you are using an enum now, you can delete the "Default" from the case structure. There should only be two cases: v1 and v2.

svm-ni: Removed "Default" from case structure

svm-ni: - Converted Measurement Service Version to typedef and named as **MeasurementServiceVersions.ctl**.
- Added MeasurementServiceVersions.ctl in MeasurementLinkServerUserSettings.ctl.
- Added MeasurementServiceVersions.ctl in front panel of Create Measurement Services.vi
- Added MeasurementServiceVersions.ctl in the block diagram of Create Measurement Services.vi. This will be used when the array is empty.

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
pbirkhol-ni: It looks like you made the changes in all the right places, but you made the wrong thing a typedef. You should make the enum itself a typedef, not the array.

dixonjoel:
@pbirkhol-ni isn't this a breaking change for existing measurement services? @svm-ni have you tried opening a measurement service created with an older version of the framework with your new changes?

Also, I don't quite understand how the user's supposed to use this feature. Do they show this user settings control on the Run Service.vi front panel and add a specific version to the array if they only want, e.g. V2?

Source/Runtime/MeasurementLink Measurement Server/Helpers/TypeDefs/MeasurementServiceVersions.ctl
dixonjoel, line 1: Most other places use a capital 'V1' and 'V2'. I think 'Version 1' and 'Version 2' would be even clearer, but I'd be ok with either. @pbirkhol-ni what do you think about capitalizing this or changing the 'Version 1'?

dixonjoel (changes requested):

Source/Runtime/MeasurementLink Measurement Server/Helpers/TypeDefs/MeasurementServiceVersions.ctl
pbirkhol-ni: I am okay with capital `V1` and capital `V2`. I don't like `Version 1` and `Version 2`.

pbirkhol-ni:
> @pbirkhol-ni isn't this a breaking change for existing measurement services? @svm-ni have you tried opening a measurement service created with an older version of the framework with your new changes?
> 
> Also, I don't quite understand how the user's supposed to use this feature. Do they show this user settings control on the Run Service.vi front panel and add a specific version to the array if they only want, e.g. V2?

This change is just for enabling testing in our test suite so that we can make sure both V1 and V2 services work. Otherwise, LabVIEW will always advertise both versions and our client will always pick V2. 

I would be surprised if any customers were depending on this input. It is a recommended input, so if you aren't wiring it (and I doubt any customer is since our generator doesn't wire it), then there is no issue. The defaults do the right thing. If you were wiring it, this would break compatibility because of the name change. Since the fix is trivial and I suspect no one is using this feature anyway, I asked Sharath to just modify and rename the existing typedef instead of creating a new one.

Source/Runtime/MeasurementLink Measurement Server/Run Service.vi
svm-ni: Made the Enum control as typedef

Source/Runtime/MeasurementLink Measurement Server/Helpers/Create Measurement Services.vi
pbirkhol-ni, line 1: You should make the `Measurement Service Versions` input a required input.

svm-ni: Updated the VI

dixonjoel (approved):

PR #351, opened 2023-10-18
Edit of MeasurementLink functions palette, class palette for the session reservation class

dixonjoel:

Add a palette for measurement context, session reservation, pin maps, and MI instruments. Install it with the measurement service VI package. Reuse the same palette for the right-click context palette for a session reservation class wire.
image

image

Creates an initial API for the MeasurementLink LabVIEW palette.

Built and installed the .vip locally and tested.


Source/Build Specs/MeasurementLink Service.vipb
jasonmreding, line 1: Some thoughts here:
- Move `Reserve Registered Sessions` to the left most column under `Reserve Sessions`. This mirrors what driver APIs typically do when there are multiple overload for creating a session/object.
- Move `Unreserve Sessions` to the far right column. Since this is the closing/terminating API for the class. it would mirror how drivers and other APIs typically put the create/init in the left most column and the destroy/close in the right most column on the same row.

So the final layout would be:

| | | |
| --- | --- | --- |
| Reserve Sessions | Close Sessions | Unreserve Sessions |
| Reserve Registered Sessions | Register Sessions | Unregister Sessions |

Another option would be:
| | | | | |
| --- | --- | --- | --- | --- |
| Reserve Sessions | Register Sessions | Close Sessions | Unregister Sessions | Unreserve Sessions |
| Reserve Registered Sessions | 

And one final option would be:
| | | | | |
| --- | --- | --- | --- | --- |
| Reserve Sessions | Create Sessions Palette | Get Connections Palette | Close Sessions | Unreserve Sessions |
| Reserve Registered Sessions | Register Sessions | Unregister Sessions |

Here the driver specific APIs for creating sessions and getting connections would get place into two separate sub-palettes. This layout then forms the expected call sequence for the most common use case within a measurement service. I'm still not sure how much I like this, but I thought I would throw out some ideas and see if anything sticks.

jasonmreding (approved):

Source/Build Specs/MeasurementLink Service.vipb
pbirkhol-ni: There are a few workflows for session reservation:
- The measurement workflow:
    - Reserve Sessions (this is on the MeasurementContext class)
    - Use the reservation
    - Close Sessions
    - Unreserve Sessions
- The setup code module workflow:
    - Reserve Sessions (this is on the Session Reservation class)
    - Register Sessions
    - Unreserve Sessions
- The teardown code module workflow:
    - Reserve Registered Sessions (this is on the Session Reservation class)
    - Unregister Sessions
    - Unreserve Sessions

My main concern is that customers will try to use the Session Reservation Reserve Sessions when they should be using the MeasurementContext Reserve Sessions. I'd almost prefer having a code module palette that has the VIs for that workflow.

dixonjoel: What about Unreserve sessions? Would that be on the sub-palette? It's used for both workflows.

Would this mean
* Close Sessions
* Unreserve Sessions
* Code Modules
  * Reserve Sessions
  * Register Sessions
  * Reserve Registered Sessions
  * Unregister Sessions

dixonjoel: See the image in the description. This is a first pass and we can always tweak it later.

pbirkhol-ni:
> What did you do for the right click menu for the MeasurementContext palette? Does it still have the same right click menu as before?

I see now. It looks like you removed it.

dixonjoel:
> What did you do for the right click menu for the MeasurementContext palette? Does it still have the same right click menu as before?

For this PR, I just removed it since it didn't seem like we had an appropriate one to link to from the .vipb palettes.

PR #352, opened 2023-10-18
Add tests for Close Sessions

pbirkhol-ni:

#AB2520987

This PR adds tests for Close Sessions

Tests pass


Source/Tests/Tests.Runtime/Measurement Server/Get Connections/Helpers/Session Json to Data - Get Connections.vi
pbirkhol-ni, line 1: This VI had some logic to populate sessions in the Connection Map. I single sourced that logic into a new subVI called `Add Sessions to Connection Map.vi`. So the only change to this VI is to replace the existing logic with a subVI that contains the same logic.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Fake Instrument/Close MeasurementLink Session.vi
pbirkhol-ni, line 1: I added the ability for this VI to throw an error.
image

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/Test Inputs/Close - 2 Fake Sessions Registered 1 Fake Session Unregistered First Session Bad.vi
jasonmreding, line 1: This VI is under the Test Inputs folder on disk but under the Helpers folder in the project. I'm not sure which is correct, but please make them consistent.

image

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/test_MultipleSessionsReserved_CallCloseSessionsWithErrorInput_ConnectionMapSessionsCleared.vi
jasonmreding, line 1: We should consider packing the `Fake Instrument` the same way we plan to for real drivers. In other words, create `Create Session` and `Get Connections` VIs as needed. I think this will simplify and improve readability of some of the test code. I don't consider it a blocker for this PR though.

jasonmreding, line 1: Rather than use `Assert Equal.vim` consider creating another helper called something like `Assert Sessions Cleared` which will just inspect the variant sessions field. This will reduce coupling/duplication between constants on the test's diagram and the test input VI.

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/Test Inputs/Close - 2 Fake Sessions Registered 1 Fake Session Unregistered First Session Bad.vi
pbirkhol-ni: That's a mistake. I fixed it in the latest commit.

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/test_MultipleSessionsReserved_CallCloseSessions_SessionsClosed.vi
jasonmreding, line 1: This is probably subjective, but I probably would have collapsed the "SessionsClosed" and "ConnectionMapSessionsCleared" test cases into a single test. That would mean instead of four tests here (error and no error variations), there are now just two. You can take or leave this suggestion.

jasonmreding, line 1: This is as good of a place as any to add this nit. If the intent is for tests to be grouped manually/logically by the developer underneath the folder in the project, then groups tests with the same setup next to each other. Otherwise, configure the project folder so items under it are arranged alphabetically. If nothing else, it makes it easier match files in github PR with files in the project (since github doesn't let you resize the files pane and the names are so long they are always truncated).

jasonmreding (approved):

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/test_MultipleSessionsReserved_CallCloseSessions_SessionsClosed.vi
pbirkhol-ni: I thought about it. And now that I am creating an assertion VI that just checks the session field, I think I'll do it.

pbirkhol-ni: I am going to create a follow-up work item to better standardize our tests. They have grown organically with different best practices over the past few weeks. For this PR, I am not going to change organization, but I will add it as an action item to the user story.

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/test_MultipleSessionsReserved_CallCloseSessionsWithErrorInput_ConnectionMapSessionsCleared.vi
pbirkhol-ni: I will add this to the follow-up test user story that I mentioned below.

pbirkhol-ni: Okay, I created an `Assert Sessions in Connection.vim`. I made it way more flexible than I probably needed to and fell into a bunch of VIM pitfalls but it works now.

I also replaced the session logic in a few other tests that weren't originally part of this PR.

Source/Tests/Tests.Runtime/Measurement Server/Close Sessions/test_MultipleSessionsReserved_CallCloseSessions_SessionsClosed.vi
pbirkhol-ni: Done.

pbirkhol-ni: AB#2557951

PR #353, opened 2023-10-18
Adding TestStand Example Files for NI-Digital

WesleyTangNationalInstruments:

- Adds the TestStand sequence file for Ni-Digital example.
- Adds TestStand Code Module VIs.
- Removed usage of Helpers library. 

Note that this NI-Digital example has more complexity than the examples for the other drivers due to the additional setup steps required. 

[AB#2532132](https://ni.visualstudio.com/DevCentral/_backlogs/backlog/InstrumentStudio%20Platform/Features/?workitem=2532132)

- TestStand sequence executes and passes.
image

- Measurement works in InstrumentStudio.
image


Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/subVIs/Get Pin Map ID.vi
jasonmreding, line 1: All of the sub VIs in this folder should be deleted. Instead you should use the new APIs recently added to `MeasurementLink TestStand Integration` library. 

image

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Create and Register Sessions.vi
jasonmreding, line 1: Please utilize the new APIs added to `Session Reservation` class along with the driver specific `Create Sessions` wrappers to implement this logic. It should simplify the diagram significantly.

image

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Update Pin Map.vi
jasonmreding, line 1: You should be able to replace the contents of this VI with this single VI. 

image

Either that or replace it with the contents of that VI if you think that looks too weird. Ultimately, this code should go away in favor of a build in step type that does this.

image

The recommendation from TestStand folks is to not call <vilib> VIs directly from TestStand.

jasonmreding (changes requested):
I will probably be OOO by the time you post updates, so I'll let other owners take it from here. If necessary, feel free to remove me from the review if that's needed to satisfy the PR gate.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Update Pin Map.vi
WesleyTangNationalInstruments: I replaced the contents of this VI to just call into the new "Update Pin Map for Sequence.vi" as suggested.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/subVIs/Get Pin Map ID.vi
WesleyTangNationalInstruments: Removed subVIs and replaced with "MeasurementLink TestStand Integration.lvlib" ones as suggested.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Create and Register Sessions.vi
WesleyTangNationalInstruments: Switched to new Session Management API for the TestStand Code Module VIs as suggested.

pbirkhol-ni, line 1: I think we should put all of the file loading into this VI. That means there will be two VIs: one for updating the pin map and one for doing all initialization. Maybe you can rename this VI to something like `Initialize Sessions.vi`.

The reason why we split up `Update Pin Map.vi` and `Create and Register Sessions.vi` is because you update the pin map for all instruments at the same time but you would probably want to initialize sessions for each driver separately. However, that means we have to split up everything into separate code modules. I think it is okay to do more than just Create and Register in this VI. It will certainly reduce the amount of boilerplate you have to write.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Unregister and Destroy Sessions.vi
pbirkhol-ni, line 1: Since the default timeout is 0, I don't think we have to explicitly wire it. You can also consider removing the blurb about a 0 timeout from the comment.

pbirkhol-ni, line 1: Same comment as before about not needing to clear or merge errors. You also don't have to mention clearing errors in the comment.

pbirkhol-ni (approved):

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Unregister and Destroy Sessions.vi
WesleyTangNationalInstruments: Updated to wire errors as normal for this and other VIs. Updated comments to no longer reference this.

WesleyTangNationalInstruments: Removed explicit `timeout=0` for this and other VIs. Updated comments to no longer reference this.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Create and Register Sessions.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Updated to wire errors as normal for this and other VIs. Updated comments to no longer reference this.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Update Pin Map.vi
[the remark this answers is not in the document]
WesleyTangNationalInstruments: Renamed all inputs with lower case for these VIs. Outputs are already all be lowercase.

Source/Example Measurements/NI-Digital SPI/NI-Digital Measurement/TestStand Code Modules/Create and Register Sessions.vi
WesleyTangNationalInstruments: I think I could go either way on this. I could see a user either preferring having smaller test steps with more granularity versus having larger test steps to reduce boilerplate code and the number of code modules or vice versa. The python NI-Digital MeasurementLink example loads files in multiple test steps and I think there is some value is having consistency between these two examples. Personally, I think the granular test steps is slightly more useful as having separation clearly visible in TestStand through multiple test steps makes it easier to work with, but I can see how having a bunch of test steps is much more cumbersome in LabVIEW than in python. 

I've opted to leave this as-is for now with the multiple test steps for file loading so we can get this PR in today without another big change, but if we want these examples to push users towards using less granular test steps to avoid the boilerplate overhead, let me know and we can create a work item to address that in a separate PR.

PR #354, opened 2023-10-19
Remove Waveform.ctl

gpachecoNI:

When opening the NIFgenStandardFunction.instudioproj you could quickly see Labview trying to look for the Waveform.ctl 
go ahead and remove the file _Waveform.ctl_  and a ring with no type Def. Use the ring from the _NI-FGEN Configure Standard Waveform_ to replace the one on _Measreument Logic.vi_

A quick fix to avoid a weird search from labview to look for the _Waveform.ctl_ when opening IS.

Rerun Example:
image

Change of Ring
image


Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/Measurement Configuration.ctl
pbirkhol-ni, line 1: You should make the caption for `Waveform` have a lowercase `w`.

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction.sfp
```diff
@@ -1,18 +1,18 @@
...
-			<Screen ClientId="{4311038b-c332-4083-815f-ec9fb7485a5f}" ConfigurationParameters="{&quot;@type&quot;:&quot;type.googleapis.com/ConfigurationParameter&quot;,&quot;pinName&quot;:&quot;Pin1&quot;,&quot;waveform&quot;:0,&quot;frequency&quot;:1000000.0,&quot;amplitude&quot;:1.0,&quot;duration&quot;:1.0}" DisplayName="NI-FGEN Standard Function (LabVIEW)" Id="8cfefc2aa3804417adfaaa04f5bc994f" PanelPresentation="ConfigurationWithVisualization" ServiceClass="ni.examples.NIFgenStandardFunction_LabVIEW" xmlns="http://www.ni.com/InstrumentFramework/ScreenDocument">
+			<Screen ClientId="{c38c1f48-fd94-47f4-a40c-99f1fe886dab}" ConfigurationParameters="{&quot;@type&quot;:&quot;type.googleapis.com/ConfigurationParameter&quot;,&quot;pinName&quot;:&quot;Pin1&quot;,&quot;waveform&quot;:2,&quot;frequency&quot;:1000000.0,&quot;amplitude&quot;:1.0,&quot;duration&quot;:1.0}" DisplayName="NI-FGEN Standard Function (LabVIEW)" Id="8cfefc2aa3804417adfaaa04f5bc994f" PanelPresentation="ConfigurationWithVisualization" ServiceClass="ni.examples.NIFgenStandardFunction_LabVIEW" xmlns="http://www.ni.com/InstrumentFramework/ScreenDocument">
```
pbirkhol-ni, line 15: Can you change the default waveform back to Sine?

pbirkhol-ni (approved):

Source/Example Measurements/NI-FGEN Standard Function/NIFgenStandardFunction/Waveform.ctl
synesthetize, line 1: I mentioned this in my initial PR for the FGEN example but the primary reason for adding `Waveform.ctl` was to work around the limitation that the driver enums are largely implemented as Ring types and we currently don't support automatically annotating LabVIEW ring types. If we want to remove this control then we should seriously consider adding that capability to the MeasurementLink LabVIEW service. Otherwise, this parameter will show up in TestStand steps as an integer instead of using the enum combobox styling.

e.g. when using the enum type:
before202736918723

e.g. when using the ring type:

gpachecoNI: @pbirkhol-ni 
Do you want me then to go ahead and implement the build as the Keysight 34401A DMM Measurement example?

pbirkhol-ni: No, let's go ahead and do what Everett was doing in his original implementation, but just don't put the control in the data folder. Instead, create a new destination directory in the that is next to the exe. Similar to what we had discussed yesterday. Then do the same for DMM.

pbirkhol-ni: The other question I have is whether we could implement the typedef as an enum, even if the drivers don't do it that way. That wouldn't work for the DMM example because of the resolution, but it might work for FGEN. Even if we can use an enum, though, it is probably best to be consistent and use the additional destination directory for both examples.

gpachecoNI:
Abandon this PR, we would implement it in a different way.

PR #363, opened 2023-10-23
Add controls back and change build adding a Controls Folder

gpachecoNI:

Fixes the build and adds back the controls so they can be used for TestStand. 

[AB#2533874](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2533874)

Re-ran example

image
image


Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement/ResolutionInDigits.ctl
pbirkhol-ni, line 1: Since this is a ring I assume that it won't show up as a dropdown in TestStand, right? One way you could get this to work in TestStand is to make this an enum instead and then create a subVI that you call from `Measurement Logic.vi` that converts from the enum to the ring.

gpachecoNI: Yeah is being used as a double I will make the change.

pbirkhol-ni: I don't think this will work. I'd be surprised if the measurement even ran without error. You will need to use the same enum on the UI and in the Measurement Configuration.ctl. Then you'll have to convert the enum to the driver's ring control before wiring it into the driver VI.

gpachecoNI: image

gpachecoNI:
Will be merging these results with this other pull - https://github.com/ni/measurementlink-labview/pull/364

PR #369, opened 2023-10-24
Include a constant for DCPower type ID

dixonjoel:

Adds a Type ID constant for the 'niDCPower' string. Add this to the DCPower palette and use it in the code modules. We can use this PR to set a pattern and a follow up PR will add constants for the remaining 4 drivers.
image
image

Use this constant in multiple places instead of hard-coding the string all over.
Start of [AB#2552370](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2552370)

Installed the .vip built locally and ensure the DC Power example isn't broken.


Source/Runtime/Measurement Server.lvproj
```diff
@@ -132,6 +132,7 @@
 			<Item Name="niFgen_64.dll" Type="Document" URL="niFgen_64.dll">
 				<Property Name="NI.PreserveRelativePath" Type="Bool">true</Property>
 			</Item>
+			<Item Name="Initialize and Close Session Behavior.ctl" Type="VI" URL="../MeasurementLink Session Management/Session Reservation/Initialize and Close Session Behavior.ctl"/>
```
pbirkhol-ni, line 135: Strange. This is in Jonathan's PR, which hasn't even been submitted yet.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Session Reservation.vi
pbirkhol-ni, line 1: Why did this VI change? I think you can revert it.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink niDCPower/Instrument Type ID.vi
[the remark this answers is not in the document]
jasonmreding: - Consider making this an inline VI
- Consider making this part of the class library instead.
- I don't know what sort of iconography is in flight for this. Lacking anything official, my recommendation would be to try and make it look more like the string constants on the string palette. That means pink outline and not header/banner. Not sure what would go in the middle, but ideally height would be the same.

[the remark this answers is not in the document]
dixonjoel: > Consider making this part of the class library instead.

You mean a, e.g. public VI under the class, right?

> I don't know what sort of iconography is in flight for this. Lacking anything official, my recommendation would be to try and make it look more like the string constants on the string palette. That means pink outline and not header/banner. Not sure what would go in the middle, but ideally height would be the same.

For now, I think I'll try to make it look like a string constant and put the driver name inside like it has in the header now? We wanted some homage to the driver. Maybe we can have a pink outline, but a driver-like fill color? Anyway, I'll play around with it and like paul said, we should get icons made for us anyway.

Source/Runtime/Measurement Server.lvproj
dixonjoel: Weird. I think I'll clean things out and kind of start over on this.

PR #374, opened 2023-10-25
Add instrument type id constants for the remaining 4 instruments

dixonjoel:

Adding instrument type id for the remaining instruments - Digital, DMM, FGEN, and Scope. Same as #369 but repeated for these instruments. Here's a typical usage:
image

Ease of use so the customers don't have to remember these magic strings.

Interactive testing of the examples.


pbirkhol-ni (approved):
I didn't check the palettes. I figures those will get enough reviews later.

PR #376, opened 2023-10-25
Validate inputs for Reserve Sessions

pbirkhol-ni:

[AB#2532052](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2532052)

`Reserve Sessions.vi` should error if a user passes an empty pin string or a zero element pin array. This PR adds the new errors and tests.

Auto tests


Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Is Pin Array Valid.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Classes/MeasurementContext/Reserve Sessions - 1Pin.vi
pbirkhol-ni, line 1: image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
dixonjoel, line 1: This seems pretty wordy.
> The pin or relay name(s) input is not valid. When specifying a single pin, the name must be non-empty. When specifying multiple pins, the array must have at least one name and every name must be non-empty.

How about something like this?
> The pin or relay name(s) input is not valid. It must be a pin or array of pins and the pin names cannot be empty strings.

dixonjoel (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
dixonjoel: Ehh I realized you need to tell them it can't be an empty array. Just leave it as is.

PR #377, opened 2023-10-26
Update TestStand code modules in examples to use the session initialization behavior

dixonjoel:

Updates all of our examples to use the new session initialization behavior for initializing and closing. These examples were previously explicitly closing or detaching the sessions directly using the driver APIs.

Updates examples to latest best practices.

None. Simple code refactoring.


pbirkhol-ni (changes requested):
You should make a similar change for the Keysight example. Because this particular example is a little tricky, I want to see the changes before you submit. Here is what I think you should do:
- The Register code module should match the other code modules. That means you should remove the True boolean wired to Reset since wiring Initialize and Detach will do a reset.
- Consider making an instrument type id constant like we do for the MI drivers. I realize this is out of scope of this PR, but I think it will be easier to just sneak it in.
- The Unregister code module should match the other code modules, I think. It is a little strange, because we'll essentially be creating a session only to immediately close it. However, knowing to shortcut the Attach and Close behavior would rely on us knowing the internal implementation details. I'd rather show best practices, even if it seems a little silly.

dixonjoel:
> You should make a similar change for the Keysight example. Because this particular example is a little tricky, I want to see the changes before you submit. Here is what I think you should do:
> 
> * The Register code module should match the other code modules. That means you should remove the True boolean wired to Reset since wiring Initialize and Detach will do a reset.
> * Consider making an instrument type id constant like we do for the MI drivers. I realize this is out of scope of this PR, but I think it will be easier to just sneak it in.
> * The Unregister code module should match the other code modules, I think. It is a little strange, because we'll essentially be creating a session only to immediately close it. However, knowing to shortcut the Attach and Close behavior would rely on us knowing the internal implementation details. I'd rather show best practices, even if it seems a little silly.

@pbirkhol-ni Can I do this in a separate PR?

pbirkhol-ni (approved):

pbirkhol-ni:
> > You should make a similar change for the Keysight example. Because this particular example is a little tricky, I want to see the changes before you submit. Here is what I think you should do:
> > 
> > * The Register code module should match the other code modules. That means you should remove the True boolean wired to Reset since wiring Initialize and Detach will do a reset.
> > * Consider making an instrument type id constant like we do for the MI drivers. I realize this is out of scope of this PR, but I think it will be easier to just sneak it in.
> > * The Unregister code module should match the other code modules, I think. It is a little strange, because we'll essentially be creating a session only to immediately close it. However, knowing to shortcut the Attach and Close behavior would rely on us knowing the internal implementation details. I'd rather show best practices, even if it seems a little silly.
> 
> @pbirkhol-ni Can I do this in a separate PR?

Yep!

PR #381, opened 2023-10-26
Fix the TS code modules for 34401 example

dixonjoel:

Update the 34401 example to use the session initialization behavior just like our other examples.
image

Also adding a constant for the instrument type id.

Makes the example code standard with our API. Also prepares us to easily change the implementation when we have grpc-device support for VISA.

None. Simple code refactoring.


Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Instrument/Session Factory/Instrument Type ID.vi
pbirkhol-ni, line 1: You should move this VI to the library.

pbirkhol-ni, line 1: You should inline this VI.

Source/Example Measurements/Keysight 34401A DMM Measurement/Keysight 34401A DMM Measurement/TestStand Code Modules/Create and Register Sessions.vi
pbirkhol-ni, line 1: For the comment in point 2, I wouldn't talk about resetting the session. It used to happen explicitly, but now it is implicit.

pbirkhol-ni (approved):

PR #383, opened 2023-10-27
Add a new way to invalidate the client pool and location pool caches

pbirkhol-ni:

This PR creates client wrappers for the Discovery Service, Session Management Service, and Pin Map Service and then uses these client wrappers in the `Session Reservation` class and the `MeasurementLink TestStand Integration` library. The interface to the client wrappers are nearly identical to the grpc-labview interface. Inside the implementation of each wrapper VI, we get the `grpc client id` from an FGV and then proceed to call the associated client VI. If there is an error, then we close the client, open it again, and attempt the original call.

This PR also creates a similar mechanism for the location pool. The location pool is currently only used when calling Initialize on driver sessions, so a retry mechanism was added here as well.

Thorough automated testing.


Source/Runtime/Measurement Server.lvproj
jasonmreding, line 1: Using the project file to make general comments in a threaded fashion:

> The drawback to this implementation is that it is open to race conditions if you try to access the same service in parallel and there is an error. For example, if I call Reserve Sessions in parallel and the first call errors, we will clear the cache and close the grpc client. The second instance may have already received the cached client id, though, which will fail. 

I think we can solve most of the legitimate race conditions if the "destroy client" action also takes in the existing client id to be destroyed. If that ID doesn't match the copy stored in the FGV, then just ignore the request and assume the client has already been destroyed. This would prevent "valid" clients from erroneously getting destroyed from sequences like:

1. Client1: Get Client (Id=1)
2. Client2: Get Client (Id=1)
3. Client1: Invoke RPC that errors (Id=1)
4. Client2: Invoke RPC that errors (Id=1)
5. Client1: Destroy Client (Id=1)
6. Client3: Get Client (Id=2)
7. Client2: Destroy Client (ends up destroying Id=2 when Id=1 was used)
8. Client3: Invoke RPC errors when it should have succeeded

> However, perhaps this isn't so bad. It was likely to fail anyway, so the only issue is that the error returned might be confusing. 

I think you are probably talking about a scenario like below, which I agree isn't that above. However, the situation I described is bad and something we should strive to prevent.

1. Client1: Get Client (Id=1)
2. Client2: Get Client (Id=1)
3. Client1: Invoke RPC that errors (Id=1)
4. Client1: Destroy Client (Id=1)
5. Client2: Invoke RPC that errors (Id=1). However the error is about an invalid client ID rather than an RPC error from the server.

> We would also have to make sure that grpc-labview can handle these scenarios and not crash. For example, what happens if I am accessing a grpc client at the same time another thread is closing that client?

These are already race conditions / issues that grpc-labview needs to handle. We should do our due diligence with testing and file bugs as necessary. However, I don't think it should impact our design unless we find things are horribly broken and bug fixes can't be completed in time to fit our schedule.

jasonmreding, line 1: While tedious, this approach seems reasonable. However, I don't really like the idea of a bunch of new top level libraries for the wrappers. I think I would prefer a single client library that contains all of the code, but I could probably be persuaded otherwise. My initial thought is that most of this should be an implementation detail and could remain private. However, there is probably some case I'm not considering where that's not feasible. If that's the case, then we could either create another nested library or just a folder called "wrappers" or something like that where it's easy to manage the access scope as a group via the folder. I really cringe at having to add even more circular references between libraries just to maintain friendship status for community scoped stuff.

jasonmreding, line 1: Another thing I think we should consider is only clearing the cache for specific, well known errors which indicate the service isn't available at the given address/location. This may require some trial/error testing with grpc-labview to see how reliable/consistent the error reporting is.

jasonmreding: If we had something like above, it would also be more practical to employ retry logic for each RPC. This would likely provide a better experience in the case where the user kills the discovery service for some reason in between runs. With the current code, the user will have to rerun/retry at least once in the best case to clear out all of the bad, cached connections. However, it could be more. For example, consider the following when running from TestStand after killing the discovery service:
1. User tries to run the sequence.
2. The process setup step for updating the pin map might generates due to a stale address/port and generates an error. This clears the cached pin map id and kills the sequence run. 
3. User tries running again.
4. The step to update the pin map is fine, but now the steps to register sessions with the session manager fail. This clears the cached session manager id and kills the sequence run.
5. User tries running again.
6. User now runs into the cached service address for each driver when creating sessions as part of process setup. This PR doesn't address this issue so they would be dead in the water unless they restart TestStand or know how to unload all code modules (I think).
7. Assuming we have an invalidation solution for drivers, the user would have to repeat steps 5 and 6 for each driver used before they could get back to a running state.

On the other hand, performing a retry for every RPC could really slow things down to a crawl when services will never be available so it's not without it's risks either.

jasonmreding, line 1: I wish grpc-labview offered an API for gRPC interceptors. I suspect they could be used pretty succinctly in this case to implement cache invalidation with the "channel pool" itself. However, we would still have to figure out retries as I don't think it would help with that.

pbirkhol-ni: > I think we can solve most of the legitimate race conditions if the "destroy client" action also takes in the existing client id to be destroyed.

Unfortunately, you are not guaranteed to get a unique grpc client id when recreating a client. I found this out a few months ago when trying to write tests for the client pool. It is possible to destroy a client and then create a new client that has the same client id. In my testing, I got the same client id ~10% of the time.

One thought I had was to do the entire grpc call within the FGV. It would have a case for every API call (register, reserve, etc). Each case would create the client if necessary, invoke the call, and destroy the client in an error. This would avoid race conditions, but would serialize all calls to the service.

pbirkhol-ni: We could certainly add these VIs to the grpc-labview generated libraries. This would mean that we are modifying the grpc-labview generated libraries, but it is purely additive, which maybe isn't so bad. I have a code generator script that I've been working on to automate the grpc-labview code generation. I can probably also automatically add additional VIs to the generated libraries without much issue.

pbirkhol-ni: I guess we could generate our own unique ID every time we initialize a new client. So the output of the FGV would be a grpc client id and a unique id that we create. If you have to destroy the client then you pass in the unique id instead of the grpc client id.

pbirkhol-ni: We could also detect errors in the FGV when using the Destroy Client action. If there is an error, we could clear it and generate a new client. Then the calling code could retry. This would make the implementation more tedious, and it would couple the FGV to its calling code, but that might be okay since this would all be private anyway. I'll play around with it.

jasonmreding: I was thinking that we would add it to existing or newly added client libraries that already sit on top of the libraries containing the generated APIs.

jasonmreding: > Unfortunately, you are not guaranteed to get a unique grpc client id when recreating a client. I found this out a few months ago when trying to write tests for the client pool. It is possible to destroy a client and then create a new client that has the same client id. In my testing, I got the same client id ~10% of the time.

This is unfortunate and probably something for which we should file a bug, especially if it makes it possible for a grpc ID for a destroyed client for service X to get mapped to newly created client for service Y.

> One thought I had was to do the entire grpc call within the FGV. It would have a case for every API call (register, reserve, etc). Each case would create the client if necessary, invoke the call, and destroy the client in an error. This would avoid race conditions, but would serialize all calls to the service.

I would not be in favor of this.

> I guess we could generate our own unique ID every time we initialize a new client. So the output of the FGV would be a grpc client id and a unique id that we create. If you have to destroy the client then you pass in the unique id instead of the grpc client id.

This might be reasonable if we can keep it an implementation detail. However, it seems like we would effectively be creating our own class type/wrapper around the tuple and duplicating the public methods as part of the declaration of this new type.

> We could also detect errors in the FGV when using the Destroy Client action. If there is an error, we could clear it and generate a new client. Then the calling code could retry. This would make the implementation more tedious, and it would couple the FGV to its calling code, but that might be okay since this would all be private anyway. I'll play around with it.

I don't think I'm following this train of thought. In what situations would the Destroy Client action generate an error, and how would generating a new client in that situation help solve some of the scenarios described above?

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
jasonmreding, line 1: Should the FGV not be configured for non-reentrant execution? It's currently configured for preallocated clones. If there is exactly one call site, it probably doesn't matter, but it's confusing.

Source/Runtime/MeasurementLink gRPC Utilities/Generic Client FGV.vi
jasonmreding, line 1: It appears that all of the implementations for the recreate vi reference method are the same. Would it make sense to only supply a create vi reference and provide a common implementation for recreate here? Closing the existing client can be done generically using the underlying grpc-labview APIs, and we can just call the create vi reference again to recreate it.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
jasonmreding, line 1: This should probably be community scope for consistency. Given all the typedefs used on the conn pane are community scoped, I don't think anyone outside the community could call it anyway.

Source/Runtime/MeasurementLink Discovery/Create Client.vi
jasonmreding, line 1: This still calls into Measurement Server library to get the discovery service address which creates a circular dependency. I know we already have a bug on this. However, I wasn't sure if you were intending to fix this as part of this PR, or if you want to fix it in a follow on PR.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
jasonmreding, line 1: The VI reference for the Resolve vi threw me for a loop at first. Then I realized it was probably done this way to avoid circular references between the libraries. However, since this VI already uses request/response messages from discovery_service.proto, it's already tied to the discovery service. Given that, I would just move this VI to the MeasurementLink Discovery library and collapse the two VIs into one.

At that point, all that's left of MeasurementLink gRPC Utilities is a generic client pool. At that point, I can see either:
1. Collapsing this library into the discovery library. Everything is currently community scope anyway so we could always move it later.
2. Update the name of the library to something like `gRPC Utilities` or `NI gRPC Utilities` to indicate there is nothing specific to MeasurementLink about this library. We could even look into distributing it as part of grpc-labview if we wanted to promote broader usage of this concept.
3. Leave the library name as is.

I don't know that I have a strong opinion here. I probably lean slightly towards option 1 if it simplifies our friends management between all of the libraries.

Source/Runtime/MeasurementLink Discovery/Create Client.vi
jasonmreding, line 1: nit: I would update the name of this virtual folder to something like "Client Pool" to help avoid confusion with the grpc-labview generated types/code from discovery_service.proto.

Source/Runtime/MeasurementLink TestStand Integration/Pin Map Client Wrapper/Create Client.vi
jasonmreding, line 1: For a typedef cluster constant that doesn't represent default values for the typedef, I would generally lean towards not displaying the cluster as an icon as I think it hides details that are important for readability. I could also see wrapping the constant in an inlined sub VI call if it improves readability. I'm open to counter arguments on this. I imagine this applies to all of the client wrappers introduced in this PR.

Source/Runtime/MeasurementLink Session Management/Session Management Client Wrapper/Create Client.vi
jasonmreding, line 1: Is there any reason you chose to add the wrapper VIs to this library vs. `Session Reservation.lvclass`? I was kind of expecting to see them on the latter since that should be the only consumer of the API outside of tests. If not for tests, I would say these should be private implementation details of the class.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
jasonmreding, line 1: Should this VI be reentrant? It looks like most/all of the driver specific APIs which call this are reentrant.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Initialize Sessions with Retry.vi
jasonmreding, line 1: Same question about reentrancy here. It seems like this should be a reentrant VI if it invokes grpc calls.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Create Sessions Impl.vi
pbirkhol-ni: Yeah, this was my mistake. I set it to non-reentrant to help with debugging and forgot to set it back. I fixed it.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Initialize Sessions with Retry.vi
pbirkhol-ni: Yeah, this was my mistake. I set it to non-reentrant to help with debugging and forgot to set it back. I fixed it.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
pbirkhol-ni: At one point that was true, but I had so much success testing `Generic Client FGV.vi` with mocks that I decided to do the same with this FGV. This "shared" FGV is used in `MeasurementLink Discovery.lvlib::Service Location FGV.vi`, which is set to non-reentrant. For testing, there is a `gRPC Utilities Tests.lvlib::Fake Location FGV.vi` that wraps the shared FGV.

I don't think I invented this pattern of pre-allocated clone reentrant FGVs that are wrapped in non-reentrant VIs. However, I can't find the blog post where I read about it.

pbirkhol-ni: The VI reference is not to prevent circular references. It is to allow easy unit testing. The original tests that I wrote started and stopped services and took a long time. The new unit tests with mocks are very quick and allow more thorough testing. I still have some integration tests that start and stop actual services, but there aren't many.

Tomorrow I'll think more about the library layout. This is an area I figured you would have feedback.

Source/Runtime/MeasurementLink Discovery/Create Client.vi
pbirkhol-ni: Yep. I am not planning on fixing this in this PR. This PR is already too big. When fixing the circular dependency bug, I will evaluate all of our dependencies and fix them.

Source/Runtime/MeasurementLink Session Management/Session Management Client Wrapper/Create Client.vi
pbirkhol-ni: Hmmm...that's a good point. I suppose I could move them to the class. I feel like the Session Reservation class is getting overwhelming with the number of VIs. I might try to organize them better with virtual folder.

pbirkhol-ni: I'll also have to rename the wrapper VIs, which is probably okay.

Source/Runtime/MeasurementLink TestStand Integration/Pin Map Client Wrapper/Create Client.vi
pbirkhol-ni: I generally agree with you. I collapsed the cluster because it is really big when expanded. However, the VIs that resolve services are pretty small, so I will just expand it.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
jasonmreding: I haven't had a chance to look at tests yet so I hadn't considered that. I think you could still move this into the `MeasurementLink Discovery` library and only expose this at community scope for testing purposes. I'm not sure if that's better or worse depending on whether or not you like the idea of collapsing the `MeasurementLink gRPC Utilities` and `MeasurementLink Discovery` libraries. I just know when I first started looking at the changes, having two VIs named `Service Location FGV.vi` in two different libraries was a bit confusing, and it took me a while to figure out why the separation existed. Maybe I just need to get used to the idea of using VI references in the API for testing/mocking purposes.

Source/Runtime/MeasurementLink Session Management/Session Management Client Wrapper/Create Client.vi
jasonmreding: Yeah, I was expecting you would still want a virtual folder called "wrappers" or something like that. I don't know that I care if there is a matching folder on disk or not. Maybe add a "with Retry" suffix if you need to uniquify the names.

Source/Runtime/MeasurementLink gRPC Utilities/Generic Client FGV.vi
pbirkhol-ni: I see what you mean. IMO, the duplication is minimal. The alternative that you are suggesting is relying on an implementation detail of grpc-labview that could change in the future. I'm going to leave it as is.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
pbirkhol-ni: The VI references were not originally meant for unit testing. I added them to support sharing the FGV amongst the client pools. Howwever, when I went to write tests for the client pools, I realized the VI references allowed for easy mocking. Because of this, I added VI references to the location pool as well, even though it technically it isn't a shared FGV.

pbirkhol-ni: I think I can alleviate your concerns by moving this VI to the Discovery library and changing the name. I would still have two VIs (the "shared" FGV and the non-reentrant VI that wraps the shared FGV), but they would have clearer names.

pbirkhol-ni: You're right. In this case, the "shared" FGV could be non-reentrant and it wouldn't' matter because the production code and the tests would never be running in the same application context at the same time. However, since the "shared" FGV could technically be used by another VI (similar to how the client pool "shared" FGV is used 3 different times), my preference is to keep it pre-allocated reentrant.

jasonmreding: The `Generic Client FGV.vi` only contains storage for a single client so it has to use pre-allocated clones so that you get one cached instance per type of client. The `Service Location FGV.vi` uses a map to cache data for N service location lookups. Given that, I don't understand why you wouldn't want to share the cache as widely as possible rather than create multiple instances of it. The only reason I can think of is so that you can perform parallel lookups of different services across gRPC. However, if that's a concern, I'd rather almost go with a single cached location per instance similar to `Generic Client FGV.vi`. Otherwise, why not just share the cache across all callers?

Source/Runtime/MeasurementLink Discovery/Create Client.vi
pbirkhol-ni: I changed the virtual folder to be `Discovery Client Wrapper`. I also moved the Generic Client Pool VIs to the Discovery library (and renamed them to be `gRPC Client Cache *` since it isn't really a pool anymore.

Source/Runtime/MeasurementLink Session Management/Session Management Client Wrapper/Create Client.vi
pbirkhol-ni: I moved these VIs to the class. I kept the VIs in a flat hierarchy on disc but added a virtual folder called `Client Wrapper` in the project. We're inconsistent on whether the virtual folder hierarchy and disc hierarchy match. About 10 years ago I went to an NI Tech presentation on large LabVIEW app development using OOP. They strongly suggested using a flat disc hierarchy, although I don't recall the rationale.

Source/Runtime/MeasurementLink gRPC Utilities/Service Location FGV.vi
pbirkhol-ni: I moved everything in the gRPC Utilities lib to the Discovery lib. I also renamed the Service Location Pool "shared" FGV to be `Service Location FGV Impl.vi`.

pbirkhol-ni: Jason and I discussed this offline and still couldn't agree on the reentrancy settings. Ultimately, the reentrancy setting makes no difference given the current state of the code, so I just made the VI non-reentrant. If this VI ever gets used somewhere else, we will have to evaluate at that time whether it needs to be pre-allocated clone or stay non-reentrant.

Source/Tests/Tests.Runtime/Measurement Server/gRPC Utilities/gRPC Utilities Tests.lvlib
jasonmreding, line 1: It looks like this library accidentally got left behind through some refactoring. I think you can delete/revert it.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Pin Map Service.vi
jasonmreding, line 1: nit: Calling this "Session Pin Map Service" is a bit weird. I would suggest just calling it "Pin Map Service".

pbirkhol-ni: That was just a naming oversite. I did Save As Copy of the `Kill Session Management Service.vi` and must have forgotten to remove `Session` from the name.

jasonmreding (approved):

PR #386, opened 2023-11-03
Adding tests to exercise new public APIs added to Session Reservation class

jasonmreding:

Adds new tests to exercise public APIs for `Session Reservation.lvclass`
- Reserve Sessions.vi
- Reserve Registered Sessions.vi
- Register Sessions.vi
- Unregister Sessions.vi

As part of this PR, I modified the test organization for the new tests slightly from the norm. If we like this convention, we can gradually apply the same structure to the other tests that exist.
- Created virtual folder hierarchy to match corresponding library/class hierarchy in the source rather than having a bunch of top level folders named after functionality being tested.
- Created Tests library rather than a virtual folder to contain the actual tests.
- Moved a couple of existing tests with similar functionality into the new library
- Tweaked existing test names somewhat for consistency and brevity
- Added new test utility library. The idea is the test utility library/libraries would be the libraries that maintain friend status as necessary with product code to shield every test class/library from having to establish its own friendship (friendship isn't extended to nested libraries in LV).

- Increases code coverage of automated regressions tests
- Slight improvements to test organization and test utilities

Newly added tests passed when run locally.


Source/Runtime/MeasurementLink Session Management/Session Reservation/Test Constructor.vi
jasonmreding, line 1: The test constructor was modified to also populate the `connection map` based on the input data. I didn't end up needing this, but I decided to leave it in anyway.

pbirkhol-ni: Some of the `Json to Data` VIs also do this. It isn't necessarily a problem, but now some tests populate the connection map twice.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsRegistered_ReserveRegisteredSessions_ReservesAllSessions.vi
pbirkhol-ni, line 1: For the restart discovery service VI, you either need to wire the error wire to the edge of the case structure or turn of automatic error handling. Otherwise, if that VI runs and errors, it will hang the test suite with a modal dialog.

pbirkhol-ni, line 1: Should you be reserving the sessions before registering them? Does the session management service enforce that or is it just considered best practice?

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/Assert Session Exists.vi
dixonjoel, line 1: The behavior of this VI seems more like `Assert All Sessions Exist` or `Assert Session Exists For All Sessions`. But I could see how your VI name could be understandable. So whatever you think is the best name.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsRegistered_ReserveRegisteredSessions_ReservesAllSessions.vi
pbirkhol-ni, line 1: This is just a personal preference, but the logic for restarting the discovery service is confusing to me. The logic is too negative. I'd prefer the Restart Discovery Service VI be in the True case and the logic be `if not pass or error`. Feel free to disagree and leave it, though.

dixonjoel (approved):

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/Assert Session Count.vi
pbirkhol-ni, line 1: This and the other helper VI seem like they could be general assertion VIs. Should those go in test utilities?

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsUnregistered_ReserveSessions_IndicatesSessionsDoNotExist.vi
pbirkhol-ni, line 1: This is just a general comment about our API. This test is confusing on the surface. You register the sessions and then verify that they do not exist. However, by registering them, they _should_ exist. The reason they don't is because the reservation doesn't get updated with the latest session manager state. I don't think this is necessarily a problem with the test or with our API since it is unlikely customers will be writing code like this.

pbirkhol-ni (approved):
This organization seems reasonable. I think there are still some open questions, but nothing major at the moment.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsRegistered_ReserveRegisteredSessions_ReservesAllSessions.vi
jasonmreding: I am going to explicitly disable automatic error handling for all test VIs in this library. However, we should consider modifying our test runner so that it disables the automatic error handling property for each test VI it runs.

image

However, maybe that is still problematic for sub-VIs called by the test VI. Maybe we should add something to the PR gate that prevents any VIs under the test folder from being checked in with these properties enabled?

jasonmreding: It's not enforced. It's considered best practice as to avoid theoretical race conditions of two threads trying register sessions at the same time for the same session names. However, I think with the latest driver changes for Init/Attach semantics, you'll still get an error in this case even if you don't reserve. I omitted it here just to eliminate two grpc calls in the test and make things a little faster.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/Assert Session Count.vi
jasonmreding: I was planning to leave them as part of this test library for now and move them to the utilities library if/when they have broader use across test libraries.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsRegistered_ReserveRegisteredSessions_ReservesAllSessions.vi
pbirkhol-ni: I'm less concerned about the tests since they only get called from the framework, which will have the error out wired. In fact, I'd prefer you _not_ change that setting on all tests now since it will cause merge conflicts for me.

As for the subVIs, let's turn off automatic error handling for now. If this becomes a problem in the future we can consider adding a PR check then. I just noticed that in your PR it might be a problem because I have seen that problem in the PR build before.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsUnregistered_ReserveSessions_IndicatesSessionsDoNotExist.vi
jasonmreding: Yeah, this test is just meant to be a sanity test that the client wrapper API connects up to the service correctly. To do that, I reserve three times and assert the session_exists toggles state with each subsequent reserve. I could have omitted the first assert and just asserted the final state is the session doesn't exist. However, that wouldn't detect a bug where the session_exists state is always False. The first assert is really a precondition, but I didn't want to spend any effort on trying to add test helpers to differentiate asserts that are preconditions and which are asserts trying to validate the purpose of the test. I guess I could also have relied on the fact that we have a separate test for validating registration of sessions works as expected so don't bother asserting it here. If you think that's better, I can do that, but it feels kind of like a meh to me.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Test Constructor.vi
jasonmreding: I may take a pass at this in a subsequent PR and update existing tests to use the new test utility VI instead.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/Assert Session Exists.vi
jasonmreding: Renamed to Assert All Sessions Exist.

Source/Tests/Tests.Runtime/Measurement Server/Session Management/Session Reservation/test_SessionsRegistered_ReserveRegisteredSessions_ReservesAllSessions.vi
jasonmreding: I'm leaving them for now. The output of that node also updates the `pass` indicator. So I would have to add another `not` somewhere in there which to make the case structure logic seem positive which seems even more negative logic then before.

In general, I think we should reconsider the `pass`, `result`, and `error out` connector pane pattern used by tests. I find it a bit tedious and confusing to manage since the test will fail with either an error or false for `pass`, and it looks weird for a test to report `pass`=True and then still fail because of `error out`. I wonder if we could just get by with `error out`, but maybe this is just as tedious?

pbirkhol-ni: > it looks weird for a test to report pass=True

I agree with this. In general, I am not opposed to rethinking the connector pane. This is where I think we should lean on other LabVIEW test frameworks to see what they do. I wouldn't be surprised if they do something similar to what we're already doing.

jasonmreding: I agree we should look at how existing test frameworks solve this. If there is a good existing solution, one thing I thought about was developing a similar test class and mixin hierarchy to that used by our C# code. That would allow us to have assert functions that store of the failures so you don't have to deal with the tedium of chaining the wires around and conditionally updating the `pass` and `result` outputs. That might also have challenges if you want to stop executing a test immediately after a failed assertion. However, you could probably have Boolean output or look at force aborting the VI. I'm not sure how/if aborting the VI from within itself works. That also raises other questions about test cleanup to put things back into a good state. I'm sure that could probably be solved with similar startup/cleanup semantics we have in our other test suite, but this is getting miles ahead of where we are.

PR #387, opened 2023-11-03
Limit configuration data types to 32 and 64 bit types

dixonjoel:

grpc-labview has [an issue](https://github.com/ni/grpc-labview/issues/323) where it incorrectly serializes Int16 (or Int8) data. Leads to various problems but [AB#2562693](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2562693) represents one. This PR limits the data types in the configuration cluster to 32 and 64 bit numeric types. When a LabVIEW service is run, the user will see an error like this.
image

It turns out that we were already advertising only 32 and 64 bit support. This enforces that in the measurement service
https://www.ni.com/docs/en-US/bundle/measurementlink/page/supported-datatypes.html

Prevent confusion and strange errors in the default values and other places of the measurement service.

Built the packages and ran the measurement service attached to [AB#2562693](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2562693).


Source/Runtime/MeasurementLink Measurement Server/Helpers/Validate Configuration Metadata.vi
dixonjoel, line 1: New data type checking.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel, line 1: Took out the 8 and 16 bit types in the cases.
image

jasonmreding: What about enums of sizes other than 32 bits? I believe the protobuf spec only supports encoding as 32 bits. Does this have the same issue? I see there is still a "Enum U64". I don't know if the original list also had the U8 and U16 versions of enums and you just forgot to remove support for this variation.

jasonmreding: There is also some code in the gplugin used to host the LV UI for the service which includes mappings for these types. Now that the service will fail to start because of this condition, this should more or less be dead code. Are you planning to update that code as well to avoid confusion?

jasonmreding (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel: > I don't know if the original list also had the U8 and U16 versions of enums and you just forgot to remove support for this variation.

Yes, the list had Enum U8 and Enum U16. I didn't think about Enum U64, but it seems like you're right. I should take that out as well.

dixonjoel: Hmm LabVIEW 2020 SP1 doesn't even let me select U64 for an enum. Guess we should axe that.
image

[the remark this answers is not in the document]
pbirkhol-ni: If you implement my comment below, you won't need the subVI because we will only throw this error in a single place.

pbirkhol-ni (changes requested):
Ignore this comment. Github wouldn't let me post my review without it.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Validate Configuration Metadata.vi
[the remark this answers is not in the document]
dixonjoel: Renamed this to Validate Metadata since it's used for results as well. Removed the error handling from Get Metadata*

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel: Keeping Enum 64. I'll update the 'LabVIEW DataType to Parameter Type.vi' in a separate PR.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Parameters And Check Data Types.vi
pbirkhol-ni, line 1: This could all be replaced with an indexing terminal at the output of the loop, right? I know this is existing code, but it is very confusingly doing a simple thing.

image

pbirkhol-ni, line 1: Make sure the input is required and the control indicator names are all lower case.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
jasonmreding: I'm a little confused by the tail end of this thread. LV may support a 64 bit enum, but gRPC never will, right? That was the whole point of removing the other enum types to begin with.

dixonjoel: @pbirkhol-ni based on this, should I go ahead and remove it with another PR?

pbirkhol-ni: Yes, you can make the change. I didn't know that grpc enums are always 32bit, but just confirmed in the spec.
`Enumerator constants must be in the range of a 32-bit integer.`
https://protobuf.dev/programming-guides/proto3/#enum

PR #389, opened 2023-11-07
Remove usages of enums that were 16 bit so examples are runnable

dixonjoel:

These examples were using typedefs that were U16s. This is no longer allowed as of #387 . I'm disconnecting from the typedefs where they were driver-defined and just changing the representation for those that are owned by us in our examples.

Fix examples

Ran each example affected to ensure the Run Service.vi will run without error.


Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement.lvproj
pbirkhol-ni, line 1: My Measurement UI.vi wouldn't load until I saved it. Are you seeing a similar issue? Can you just open the Measurement UI.vi and save it and I can try again?

Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement/Measurement Configuration.ctl
pbirkhol-ni, line 1: I'm surprised this works. I thought the data types in the Measurement Configuration had to match the data types in the UI exactly. The Measurement Type control in the Measurement Configuration is a U32 but the Measurement Type control on the UI is a U16. @jasonmreding, do you know more about how closely we expect data types to match?

Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement.lvproj
pbirkhol-ni: This is true of other examples as well. Can you save the Measurement UI.vi for all modified examples?

pbirkhol-ni (changes requested):
Can you reset me when you've made the changes so I can make sure that the UI is loading on my machine?

Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement/Measurement Configuration.ctl
dixonjoel: Updated Measurement UIs to match the types.

Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement.lvproj
dixonjoel: Updated Measurement UIs to match the types.

pbirkhol-ni (approved):

Source/Example Measurements/NI-DMM Measurement/NIDmmMeasurement/Measurement Configuration.ctl
jasonmreding: The UI validation/mapping currently happens at run time. That logic is in the gplugin, and Joel is fixing that in this [PR](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/593658). We could also entertain adding build time checks to the UI buildspec. However, there isn't usually a build spec for simple UIs, and I'm not sure if or how we would add that sort of check to the build spec for the measurement service. It would also only work for a service implemented in LV so we would probably still need the runtime checks in the measurement client as a fall back.

pbirkhol-ni: Okay. I understand now. The gplugin used to treat U8, U16, and U32 LabVIEW types as UINT32 gRPC types. So if your configuration had a U8 and the UI had U32, those were seen as the same type. But with Joel's change, 8 bit and 16 bit types will no longer be supported in the UI. I don't think this is necessarily wrong, but maybe a little heavy handed? I'll comment on his PR.

jasonmreding: > I don't think this is necessarily wrong, but maybe a little heavy handed?

I'm not really sure what the alternative is. If we want to accept u8, u16, and u32 control/indicator, we could map it to a 32 bit data type in gRPC like we were doing before. For controls, this should generally work fine since we are going from a smaller type to a larger type so there is no risk of overflow. Doing the same thing for an indicator is more dangerous since in theory the backing service logic could erroneously send a larger data value that will overflow the value of the control. However, I'm pretty sure LV just truncates the data in this case without throwing an error. Hopefully, the service author would eventually figure this out through testing.

If all we were dealing with was single scalar values, allowing this would be somewhat reasonable. However, the real problem comes in with arrays and clusters. In both of these cases, you have to deal with packing of the data in the underlying LV data structure, and you have to map that to/from the correct gRPC data structure. This will be more expensive from a performance perspective (especially for array data) since you can't just block copy the data. To support different size data between the cluster used by the UI and gRPC, we would either need to make changes to grpc-labview or figure out another way to do it on our end.

In grpc-device, it appears that driver APIs that use u8 or u16 scalar data types just aren't supported. Driver APIs that use u8 or u16 array data are mapped to either repeated byte or u32 data. This is based on looking at the DAQmx read APIs for digital data. Presumably, this means the driver also has logic that throws an error if the user asks for an odd number of u16 samples. We could entertain similar types of mappings, but I don't think it's worth it. At least I wouldn't try and do anything until we have more concrete user feedback.

pbirkhol-ni: I copied your reply to azdo. Let's continue there.

Issue #292, opened 2023-09-26
Circular dependency between the Utilities library and the Measurement Server library


pbirkhol-ni:
# Bug Report

The MeasurementLink Utilities library is not supposed to depend on any other libraries. However, it currently has a dependency on the MeasurementLink Server library. The `gRPC Client Pool.lvclass:Create.vi` calls `Get Discovery Service Address.vi`, which is part of the MeasurementLink Server library. Since `Get Discovery Service Address.vi` is technically a public VI and is used by our examples, we can't just move it to a new library without breaking compatibility.

[AB#2533602](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2533602)

PR #392, opened 2023-11-09
Sever the circular dependency between the Discovery lib and the Measurement Server lib

pbirkhol-ni:

#292 
[AB#2533602](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2533602)

This PR removes the circular dependency between the Discovery library and the Measurement Server library. Prior to this change, the Discovery library depended on the Measurement Server library for `Get Discovery Service Address.vi`. This change moves `Get Discovery Service Address.vi` (and all of its private subVIs) to the Discovery library and updates all callers to use the new location. I kept the VI in the old location too so that we wouldn't break compatibility. However, I added a red X to the icon and mentioned in the context help that it is deprecated and to use the version in the new location.

I also moved an existing test to `Discovery Tests.lvlib`.

Unit tests. I also opened a measurement that was using the old `Get Discovery Service Address.vi` and verified that it still works.


jasonmreding (approved):
Looks good. I also verified this also fixes an exe build spec problem I was having with some changes I was trying to test.

Issue #350, opened 2023-10-18
Openning project with no driver installed, then install, does not clear the error of no driver until specific VI's are reopen.


gpachecoNI:
# Bug Report

Opening the project with no drivers installed in your machine will throw an error of "The NI-DMM driver is not installed". The user will then install the driver and would expect that the next time the VI is run the error to go away but it doesn't. 
Seems that the type specialization structure is under the _Initialize MeasurmentLink Session.vi_ and the _Close MeasurmentLink Session.vi_ wont get fix until someone opens the vi's. 

## Repro or Code Sample

1.- Open any of the new examples on a machine with no drivers.  _NI-FGEN Standard Function_ _NI-SCOPE Acquire Waveform_
2,- Run the example and get the expected error _"The NI-xxx driver is not installed"_
3.- Install the expected driver
4.- Re-run the example 

You will be getting the same error. The workaround is to open the VI's that have a type specialization structure. 

## Expected Behavior

The error should have been clear. 

## Current Behavior

It is impossible to remove the error of  "Driver is not installed" until you open the VI's that have a type specialization structure. 

## Possible Solution

N/A

## Context

N/A

## Your Environment

No drivers when starting adding them later.

[AB#2557116](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2557116)

jonathanmendez:
Another workaround is to force-compile the `Measurement Logic.vi` hierarchy (ctrl-shift click on the run arrow).

PR #394, opened 2023-11-10
Issue #350 - Driver errors not cleared after installing driver

jasonmreding:

Fixes Issue #350 

This initial push only includes a fix for the DCPower driver since that is the only driver I already had installed. I wanted to get feedback on the approach before updating the remaining drivers.

With this approach, all calls to driver VIs are wrapped in a `Call by Reference` node instead of calling the VI directly. The node is configured with a static reference by right clicking the node and configuring the "Call Setup" dialog. This maintains a static edit time reference which will still trigger search dialog to find missing dependency. However, if the dependency isn't found or is broken, it doesn't actually break the caller at edit time. Instead, a runtime error is generated. This ensures none of the `ISession Factory` implementations are broken when loaded into memory due to the driver not being installed. This in turn means not having one driver installed no longer breaks usages of the other drivers that are installed. Because the node maintains a static reference, it also means users don't have to explicitly include driver dependencies when building exes, source distributions, etc that you would otherwise have to do when calling VIs dynamically at run time.

Fixes Issue #350

For DCPower, manually tested:
- The driver call by ref calls work when the driver VIs are installed to <instrlib>
- The driver call by ref calls error appropriately when the driver VIs are not installed to <instrlib>
- The driver call by ref calls work as expected when the same VIs are run from LV 2020 and LV 2023. LV 2023 uses version independent addons instead of a versioned <instrlib> directory.
- A built exe runs successfully without having to manually include the driver VIs.


Source/Runtime/MeasurementLink Instrument/ISession Factory/Helpers/Merge Driver Call by Reference Error.vi
jasonmreding, line 1: image

Source/Runtime/MeasurementLink niDCPower/Session Factory/Initialize MeasurementLink Session.vi
jasonmreding, line 1: image

image

Source/Runtime/MeasurementLink niDCPower/Session Factory/Close MeasurementLink Session.vi
jasonmreding, line 1: image

Source/Runtime/MeasurementLink Instrument/ISession Factory/Helpers/Merge Driver Call by Reference Error.vi
jasonmreding: Merging Call by Ref errors is a little awkward. I decided that any error from the called VI itself would take precedence over errors reported by the call by ref node itself. I suspect this is mostly a non-issue since I expect the Call by Ref node will just report default values for the data type if the call itself fails, but I didn't test this. Because of this, I also don't wire `error in` of the Call by Ref node itself since I want the call to always execute, and then let the underlying VI decide what it wants to do if there is an error on its `error in`.

I left the error 5052 description unmodified. Technically, there are a number of reason why the call could fail:
- Driver isn't installed
- Driver is installed, but driver support for this version of LV isn't installed
- Driver is installed, but some other dependency used by the driver is missing
- Driver is installed, but running in the runtime engine and the build didn't copy over all required dependencies. We could do something custom here with a conditional compilation structure if we really wanted to.

That's a lot to try and put in an error description so I just stuck with the most plausible answer. However, if others want to take this opportunity to do some more word smithing with the error explanation, I'm happy to update it.

Source/Runtime/MeasurementLink niDCPower/Session Factory/Initialize MeasurementLink Session.vi
pbirkhol-ni: The call by ref path uses an absolute path. If we can't get it to work with a path like <instr.lib>\niDCPower\nidcpower.llb\niDCPower Initialize With Independent Channels for gRPC session.vi then this solution is likely a non-starter.

pbirkhol-ni: nit: There is a lot of empty space to the right of the call by ref node that could be removed.

Source/Runtime/MeasurementLink Instrument/ISession Factory/Helpers/Merge Driver Call by Reference Error.vi
pbirkhol-ni: FWIW, I did something similar with the client cache and location pool. The VI error took precedence over the call by ref node error.

As for the error message, it seems fine to me.

Source/Runtime/Measurement Server.lvproj
pbirkhol-ni, line 1: Adding a general comment here.

What is the developer experience for this? 
- What happens if I open this project when a driver isn't installed? Do I get a search dialog? Do we get a dialog like before that says some drivers aren't installed? IMO, the search dialog is worse.
- What happens if I try to open one of the initialize or close VIs when the driver isn't installed? Do I get a search dialog? If so, that is a step back in developer experience since we didn't have the issue with the Type Specialization structure.

I'm still not convinced that this solution is better than the conditional disable. It's not clear to me what advantage it gives us, but it comes with the potential disadvantages that I mention above.

pbirkhol-ni (changes requested):

Source/Runtime/MeasurementLink niDCPower/Session Factory/Initialize MeasurementLink Session.vi
jasonmreding: > The call by ref path uses an absolute path. If we can't get it to work with a path like <instr.lib>\niDCPower\nidcpower.llb\niDCPower Initialize With Independent Channels for gRPC session.vi then this solution is likely a non-starter.

It does work with <instrlib>, and it's something I tried. However, LV seems to always convert it to an absolute path for display purposes even though it maintains a relative link to <instrlib> internally. I also verified that these VIs install to <vilib> of both LV 2020 and LV 2023 work. The DCPower VIs used by LV 2023 have a completely different path which is why I say LV must be maintaining this as desired internally even though they always display the absolute path.

> nit: There is a lot of empty space to the right of the call by ref node that could be removed.

I did it that way so the case labels aren't truncated. The extra space is annoying, but I thought the readability without having to hover the mouse over the case label was probably more important. I don't feel strongly about that so if you feel differently, I'm willing to remove it.

pbirkhol-ni: Thanks for the explanations about relative paths.

As for the extra space, maybe you can at least move the call by ref node to the middle? I guess I'll leave it up to you. I don't care that much.  :)

Source/Runtime/Measurement Server.lvproj
jasonmreding: > What happens if I open this project when a driver isn't installed?

You get a dialog that looks like the following, but no search dialog.

image

image

> What happens if I try to open one of the initialize or close VIs when the driver isn't installed?

Nothing happens outside of the ordinary. There are no dialogs that get presented. You'll just see something like this instead.

image

> I'm still not convinced that this solution is better than the conditional disable. It's not clear to me what advantage it gives us, but it comes with the potential disadvantages that I mention above.

The advantage over the conditional disable structure approach is that an `ISession Factory` is never loaded into memory broken. That means accidental loading of a driver specific VI for a driver that isn't installed won't break all the other drivers. Once LV loads a class into an application context, it basically never unloads it until the application context goes away. That means the user has to reload the project or restart LV is they are using the main application context in order to fix this problem.

Source/Runtime/MeasurementLink niScope/Session Factory/Close MeasurementLink Session.vi
pbirkhol-ni, line 1: nit: This isn't new to your PR, but the driver string for the error should be `NI-SCOPE` instead of `NI-Scope`.

pbirkhol-ni (approved):
This looks great. As far as I can tell, it has all the benefits of the old solution with none of the drawbacks. Did you test that it works when building an exe?

Edit: Nevermind. I see you mentioned a built exe in the PR description.

PR #396, opened 2023-11-13
Update generated grpc code for Measurement Service Base V2

dixonjoel:

To get the new 'message type' field for configuration parameters and output parameters, we need to update to a new measurement_service.proto.

Allows us to move forward with supporting XY data returned from a measurement service.

None. Should be backwards compatible. G tests will run on checkin.


Source/Runtime/MeasurementLink Measurement Server Internal/RPC Service/MeasurementService/RPC Methods/Set MeasurementService Measure Response.vi
pbirkhol-ni, line 1: If you only regenerated V2, this shouldn't have changed.

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Create Metadata Response.vi
pbirkhol-ni, line 1: I don't think any of the 5 modified files in this library should have changed either, although I'm less sure of this.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Classes/IMeasurementService/Create Metadata Response.vi
dixonjoel: The Measurement Server has its own copies of ConfigurationParameter.ctl and OutputParameter.ctl (that are version agnostic). However, they didn't have the message_type field, so I had to add it. The rest are type propagation changes.

PR #399, opened 2023-11-14
Allow a LabVIEW service to return DoubleXYData

dixonjoel:

This change allows a LabVIEW service to return DoubleXYData in its outputs. 'DoubleXYData' is defined as matching the 'ni_protobuf_types_DoubleXYData' typedef that we are now shipping. This was generated from the xydata.proto file. If it is a cluster of that type, the parameter type will be TypeMessage and the message type will be ni.protobuf.types.DoubleXYData.

Allows LabVIEW services to return DoubleXYData similar to how C# and Python services already can.

Tested interactively with a new 'Game of Life' example LabVIEW service I'm working on.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Cluster Is DoubleXYData.vi
dixonjoel, line 1: Duck typing for checking if it is DoubleXYData.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel, line 1: Hard-coding that clusters are supported. We'll need a bit of code for validating the metadata to say that the cluster *is* a DoubleXYData cluster.

dixonjoel, line 1: Hard-coding that clusters are supported. We'll need a bit of code for validating the metadata to say that the cluster *is* a DoubleXYData cluster.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Results.vi
dixonjoel, line 1: If it's not an array, check if it's a cluster of DoubleXYData and return the appropriate message type.
image

pbirkhol-ni, line 1: Is an array of double XY data something we need to handle?

Source/Runtime/MeasurementLink Measurement Server/Helpers/Cluster Is DoubleXYData.vi
pbirkhol-ni: Does Get Cluster Information throw an error if the data type is not a cluster? I think I'd prefer to first get the type and if it is a cluster, then inspect it further.

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
pbirkhol-ni: It seems like you could modify this VI to contain most of the validation logic that you have in `Cluster is DoubleXYData.vi`. So it would also return a string for the `message_type`, which would be blank in all cases except XY data.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Cluster Is DoubleXYData.vi
pbirkhol-ni, line 1: I have a few comments that might be more obvious if I had been paying more attention to this feature from the beginning.
- What if I want XY data that isn't a double? Integer XY data is valid, even if it has to be converted to a double to send over the wire. Maybe this is another one of those situations where we have to enforce a particular data type on customers because of gRPC and/or proto file definitions. However, I don't like how much our implementation is starting to leak out to customers.
- Is there a plan to support generic clusters in the future? I'm wondering how what you have here would interoperate with that. I'm not even sure how supporting generic clusters would work. I guess we could just recursively traverse the cluster to figure out how to serialize and deserialize it? Would a generic solution like that alleviate the need at all for a `message_type` field?

dixonjoel: Yes, that's what it does and what I'm using the Get Cluster Information for. I suppose we could call Get Type Information first, but then we'd have to call Get Cluster Information anyway. The Error case just returns False here.

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel: I did consider that. I just wasn't sure if I wanted to pass in the variant itself into this VI and change its signature. Another aspect of the way I did it is that we don't yet want to support XYData for inputs. And the Get Metadata from Configuratons.vi also calls this basic datatype VI. When I do the whole thing with validation, I'll consider how to best consolidate the validation with the Get Metadata VIs.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Results.vi
dixonjoel: We don't really support array of double xy data. There's nothing you can bind it to in the UI right now anyway.

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
jasonmreding: As is, this change will indicate support for all clusters and not just XY Data. That seems bad. I think it would be better to pass in the variant instead and push some of the existing logic into a common sub VI.

image

In general, I don't think we should be restricting support for certain types to inputs vs. outputs. However, if it's really necessary, we should be able to add an input for this as well.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Cluster Is DoubleXYData.vi
jasonmreding, line 1: In general, having a VI like this is fine. However, I would prefer to also have a higher level method that is able to check for any valid message type / cluster type. As we scale up the number of message types we support, the performance and maintenance burden to match an "anonymous" type might become untenable. To avoid that, we might want to consider the user has to use the typedef generated for the message type from grpc-labview. At that point, we just have to match the typedef string and not every field in arbitrary order of the cluster recursively.

jasonmreding: > What if I want XY data that isn't a double? Integer XY data is valid, even if it has to be converted to a double to send over the wire. Maybe this is another one of those situations where we have to enforce a particular data type on customers because of gRPC and/or proto file definitions. However, I don't like how much our implementation is starting to leak out to customers.

I assume this really means what if I want that in my UI? At the service level, we need to restrict the data that defined by the protobuf message types. If customers really want/need integer, then we should add `Int32XYData` in addition to `DoubleXYData`. 

At the UI level. we could still support integer XY data that is mapped to double XY data at the service level. However, there will a cost for us to implement and maintain all of these conversions. It will also come at the cost of reduced performance for the user. Whether providing this convenience is worth the risk of the customer possibly shooting themselves in the foot with poor streaming performance, I don't know.

jasonmreding: > Is there a plan to support generic clusters in the future? I'm wondering how what you have here would interoperate with that. I'm not even sure how supporting generic clusters would work. I guess we could just recursively traverse the cluster to figure out how to serialize and deserialize it? 

This is as much as I've thought through it, and the answer really depends on how you want map clusters to the UI. If we restricted mapping UI controls to a single field of a cluster and didn't support cluster controls directly, we could probably support clusters at the service level today without a lot of extra work. We would still have to figure out a way to register the cluster types with LV ahead of time or take over serialization from grpc-labview and do it ourselves (similar to what we do in Python and C#) code. grpc-labview would probably have to provide us with more advance / dynamic APIs so that we could do this.

> Would a generic solution like that alleviate the need at all for a message_type field?

Maybe, If we are supporting cluster controls in the UI that expect specific data values/types for each cluster type, then we have to be able to construct those message types / values in order to give the control a value type that it expects. There are 3 general solutions that I can think of for this:
1. Hard code knowledge ahead of time of all the possible message types that can be used. This is the path we are headed down with these "well known" types.
2. Support server reflection with the service. The service then has to publish reflection information for all of the custom message types it wants to using a separate reflection service endpoint that is more or less a de facto standard for gRPC. I've messed around with this a little bit, and the support / feasibility for this per language varies.
3. Add support for dynamic message types at run time rather than rely on generated protoc message types at compile time. This is similar to the previous point, but it doesn't actually require server reflection. Some language like Java already have libraries that let you do this. However, C#, LV, and Python don't have anything that supports this currently. C# kind of has support for something akin to this using the "code first" gRPC library, but I haven't tried to actually use it to accomplish this. I suspect it is still missing some features to make this practical. However, I'm sure the maintainer is open to outside contributions.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Results.vi
jasonmreding: Again, we could support this at the UI level even if we don't support it at the service level. However, there will be a cost to converting the data from the service to the UI.

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
dixonjoel: Yeah, that makes sense. I'll update this converter to check the cluster to see if it's supported.

dixonjoel:
@pbirkhol-ni @jasonmreding I updated the data type checking. I'm turning this into an actual PR now. It can't be tested and checked in until my other PR #400 is submitted.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Is Supported Cluster.vi
jasonmreding, line 1: Consider replacing the single equality check with a case structure. That will make it easier to extend as new message types are supported.

Source/Runtime/MeasurementLink Measurement Server/Helpers/LabVIEW datatype to Measurement datatype.vi
jasonmreding, line 1: For cluster types, you are duplicating logic for the `message_type` name. You should just be able to use the output parameters from `Is Supported Cluster.vi`.

jasonmreding: image

Should the False case here report a `grpc_type` of `TYPE_UNKNOWN` instead?

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Configurations.vi
jasonmreding, line 1: image


For the array case, I would also set the `message_type` field in the cluster based on the output parameter from the array's element type. This also applies to `Get MetaData from Results.vi`.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Is Supported Cluster.vi
jasonmreding, line 1: image

You should put the message_type string constant inside the case structure as well.

jasonmreding (approved):

PR #400, opened 2023-11-15
Add generated code and .proto for ni.protobuf.types.DoubleXYData

dixonjoel:

To support LabVIEW services returning DoubleXYData, we need to have a canonical definition of the cluster for ni.protobuf.types.DoubleXYData. This PR adds a project and library for ni.protobuf.types and xydata.proto.

It also adds a build spec for the 'NI gRPC Types' package. The .vip name is 'ni_protobuf_types-1.0.0.2.vip'. The content is installed to vi.lib\gRPC\Generated APIs\ni\protobuf\types\
image

This package currently contains no palette. We intend to add a dependency on this package and a palette item referencing the DoubleXYData.ctl to the Measurement Service.vipb.

Precursor to supporting DoubleXYData in LabVIEW services.

Built and installed the Common gRPC Types.vipb and ensured that the xydata types are in the correct place in vi.lib


Source/Runtime/ni.protobuf.types.lvproj
dixonjoel, line 1: Here's what the ni.protobuf.types project looks like.
image

Source/Runtime/ni.protobuf.types/xydata/RPC Messages/ni_protobuf_types_DoubleXYData.ctl
dixonjoel, line 1: image

dixonjoel: Dropped on a VI from the .ctl in vi.lib:
image

Source/Runtime/ni.protobuf.types/xydata/ni.protobuf.types.xydata.lvlib
jasonmreding, line 1: I know we're still trying to standardize on some of this stuff on the fly, but my preference at the moment would be for the library name to match the package name in the proto file. In this case, that would mean `ni.protobuf.types.lvlib`. As more proto files get added to this package over time, that means we would need to generate all of the proto files from the pacakge at once so they are included in the same library. From a file/directory structure, I would delete the xydata folder and move everything up one directory. Thoughts/concerns with this approach? This might be something to discuss internally with the gRPC chapter sooner rather than later if we're looking to release public support for this.

dixonjoel: I saw that in your spec and I didn't see how that would work. My concern was that the generated structure is this:
image

And if we generated multiple .proto files into the same library, it seemed like all the messages would get generated together in that directory and it would get messy.

Plus like you said we would have to regenerate all protobuf/types .proto files at the same time. So if we added a .proto file, we'd have to generate all of them again even if they didn't change.

But I think either way could work. Since the generated package names won't change, maybe the organization into libraries could also change later without disruption? If so, I could go either way for this first one.

dixonjoel: The way the code generation is now, it seems like grpc-labview wants to go with one library per .proto file.

Source/Build Specs/MeasurementLink Service.vipb
jasonmreding, line 1: If this is going to be an NI wide package, it should really be in a package that's independent of measurement-link. I don't know that we want to take time to spin up a repo just for that now. However, we could at least create a new package, and add it as a dependency to the measurement-link package so that it has the long term name/packaging that we would want. This might create some dev friction though since it might require install of that package to develop. However, maybe there is still some relative pathing magic we can figure out so everything still links correctly from source?

dixonjoel: Maybe once this PR goes in, we should be adding the ni-apis repo as a submodule instead of checking in this .proto here. However, it's still internal.
https://github.com/ni/ni-apis/pull/6

bkeryan: You cannot add ni-apis as a submodule in the public LV and Python repos until we make ni-apis public.

Source/Runtime/ni.protobuf.types/xydata/RPC Messages/ni_protobuf_types_DoubleXYData.ctl
dixonjoel: Moved to a different location.

Source/Build Specs/MeasurementLink Service.vipb
dixonjoel: I moved the project to Source\protobuf\ni.protobuf.types and added a build spec for this package.

I checked in the proto file under .\ni-apis\ni\protobuf\types\xydata.proto, in anticipation of this being a sub-repo directory of the ni-apis repo at some point.

Source/Runtime/ni.protobuf.types/xydata/ni.protobuf.types.xydata.lvlib
dixonjoel: Changed over to a single ni.protobuf.types library.

Source/Build Specs/Common gRPC Types.vipb
```diff
@@ -0,0 +1,391 @@
+<VI_Package_Builder_Settings Version="2020.1" Created_Date="2023-01-23 14:05:21" Modified_Date="2023-11-15 14:16:56" Creator="lvadmin" Comments="" ID="2ad2328419eeb300544b1df2aa0dc54a">
+  <Library_General_Settings>
+    <Package_File_Name>ni_protobuf_types</Package_File_Name>
```
bkeryan, line 3: > [@dixonjoel](https://github.com/dixonjoel) pushed 1 commit.
>
>https://github.com/ni/measurementlink-labview/commit/509ee59ddb81ef91d0e630740debd6a250b294c5 Add build spec for ni_common_types-x.x.x.x.vip

Is the package name `ni_common_types` or `ni_protobuf_types`?

Why don't the product name and package file name correspond to each other?

dixonjoel: It is ni_protobuf_types. I used 'Common' and 'gRPC' in a couple places that are user-facing since I didn't know how well customers would understand that. But we can standardize on 'Protobuf' perhaps. How about these?
* NI Protobuf Types.vipb
* ni_protobuf_types-x.x.x.x.vip
* Product Name: NI Protobuf Types
* Description: Protobuf types for use across NI products

Anything else that I missed?

bkeryan: > I used 'Common' and 'gRPC' in a couple places that are user-facing since I didn't know how well customers would understand that.

I see. 

You could expand "Protobuf" to "Protocol Buffers" in the English name/description, but I don't know how much that helps.

Another approach is to lean on ni-apis and use some variation on "NI API message types" 

> Description: Protobuf types for use across NI products

Ideas for clarifying this:
- "for use across" -> "used by" or "shared by"?
- "NI products" -> "NI APIs" or "NI gRPC APIs"?
- "types" -> "data types" or "message types"?

bkeryan (approved):

ni-apis/ni/protobuf/types/xydata.proto
jasonmreding, line 1: I know next to nothing about git submodules, but is there some reason we couldn't just map the ni-apis repo as the "source/protos/ni" directory. Assuming that's possible, then I would think this proto file should be added there instead of here. Otherwise, we'll need to specify additional import directories when generating code. It's not the end of the world, but it's also not clear to me that it's necessary.

Source/protobuf/ni.protobuf.types/Client API/Create Client.vi
jasonmreding, line 1: I think you can remove the Client API VIs as I don't think they are of any use. There is no service to connect a client to. Ideally, grpc-labview would do a better job of supporting this use case. I assume if this proto file is imported from another proto file which uses these message types that the metadata for the messages defined in this package are also included in the metadata for that service?

Source/Build Specs/Common gRPC Types.vipb
jasonmreding: I would personally try to stick to a convention based naming scheme for a lot of this stuff in terms of the file names and visible package names. If there are user facing descriptions / names within the contents of the vipb/vipm files, then using friendlier naming / descriptions at that point seems fine. While that may initially lead to package names that seem less friendly, I think it will make it easier to navigate/browse for packages in the long term once you understand the packaging convention. Hopefully, the VIPM store has sophisticated enough search that users can still find what they are looking for if they make search queries using English phrases rather than exact product names.

Source/Build Specs/NI gRPC Types.vipb
jasonmreding, line 1: Shouldn't there be a dependency to the measurement service VIPB for this VIPB?

Source/protobuf/ni.protobuf.types/ni.protobuf.types.lvproj
jasonmreding, line 1: This applies to the directory structure and not specifically this file. Thinking about this more, I suspect a hierarchical folder structure would work out better than a flat folder structure (e.g. ni/protobuf/types instead of ni.protobuf.types). I think that would lead to a palette hierarchy in vilib that more closely mirrors the package sub-name hierarchy used for the proto files themselves. However, that's kind of a pain to keep up with without changes from grpc-labview that would just enforce some if this layout. Ultimately, I think somebody would have to experiment with grpc-labview and VIPB a little more to create more than one package and see what the end result looks like and what achieves the best result with the least effort from the developer. This also raises questions as to whether MeasurementLink should be installing its generated code in a similar fashion. However, for now, I'm OK moving forward with what is here and continuing to iterate on this off line. I think some of this would be good fodder for the next grpc chapter meeting.

ni-apis/ni/protobuf/types/xydata.proto
bkeryan: Git submodules are typically mapped to something like `third_party/$repo_name` (e.g. `third_party/ni-apis`). 

The ni-apis repo already contains an `/ni` directory. You can change the directory name so it doesn't match the repo name, so you could map ni-apis to `/source/protos/ni`, but this would result in `/source/protos/ni/ni`.

Likewise, the protobuf repo contains `/src/google/protobuf`. You could map protobuf to `/source/protos/google/protobuf`, but this would result in `/source/protos/google/protobuf/src/google/protobuf`.

For ni-apis and ni-internal apis, mapping to `/source/protos` would remove the extra directory name, but you can't map multiple repos to the same directory. Also, `/source/protos` + `/src/google/protobuf` still adds an extra `src`.

I am not aware of a way to map a subdirectory of another repo as a submodule.

Source/Build Specs/NI gRPC Types.vipb
dixonjoel: I would think that Measurement Service.vipb might depend on this .vipb but I'm not sure. So far, in my other PR, I don't think I depend on the typedef defined in this PR - I have just been using a string. However, I was considering making Measurement Service have some code that included an instance of the ni_protobuf_types_DoubleXYData.ctl and fished out the 'typedef name' from that to use in various places instead of hard-coding a string.

ni-apis/ni/protobuf/types/xydata.proto
dixonjoel: Well, for now, how I about I just abandon the ni-apis folder here and just put it next to the others in source\protos\ni\protobuf\types?

Source/Build Specs/NI gRPC Types.vipb
dixonjoel: We can add that dependency if necessary in the follow-up PR where I'm going to use the generated .ctl.

Source/Build Specs/Common gRPC Types.vipb
dixonjoel: Package name is currently ni_protobuf_types-<version>.vip. So I think I'm following this. For the .vipb name, I currently have 'NI gRPC Types.vipb'. Should I change that to, e.g. ni_protobuf_types.vipb perhaps?

ni-apis/ni/protobuf/types/xydata.proto
jasonmreding: That seems reasonable. I assume whenever the ni-apis repo comes online, it will also contain the existing proto files from this repo. So at that point, we'll have to decide on a mapping, but all of the NI stuff will still be under a single top level directory. For google proto files, I assume the plan is to git submodule their repo as well into a new top level folder (requiring additional import search directory) and not continue to copy their proto files into this repo or the ni-apis repo.

Source/Build Specs/Common gRPC Types.vipb
jasonmreding: I think keeping them the same would be less confusing.

ni-apis/ni/protobuf/types/xydata.proto
bkeryan: > Well, for now, how I about I just abandon the ni-apis folder here and just put it next to the others in source\protos\ni\protobuf\types?

I agree, this is a good plan.

> So at that point, we'll have to decide on a mapping, but all of the NI stuff will still be under a single top level directory. For google proto files, I assume the plan is to git submodule their repo as well into a new top level folder (requiring additional import search directory) and not continue to copy their proto files into this repo or the ni-apis repo.

Yes, all public NI APIs will be in `/third_party/ni-apis` and Google APIs will be in `/third_party/protobuf` and `/third_party/googleapis`. We will add these directories (or the `src/` subdirectory in the case of protobuf) to the import search path.

bkeryan: Also, I would like set up Renovate in this repo to update the Git submodules once a month. (It will also update Git actions when a new version is released that doesn't match the tag we are using, like `@v2`.)

jasonmreding (approved):

Source/Build Specs/ni_protobuf_types.vipb
```diff
@@ -0,0 +1,187 @@
+<VI_Package_Builder_Settings Version="2020.1" Created_Date="2023-01-23 14:05:21" Modified_Date="2023-11-16 10:58:35" Creator="lvadmin" Comments="" ID="7205f398b8afd492ca4d6a2f13d92a99">
+  <Library_General_Settings>
+    <Package_File_Name>ni_protobuf_types</Package_File_Name>
+    <Library_Version>1.0.0.8</Library_Version>
+    <Auto_Increment_Version>false</Auto_Increment_Version>
+    <Library_Source_Folder>..\protobuf\ni.protobuf.types</Library_Source_Folder>
```
bkeryan, line 6: @dixonjoel This directory doesn't exist:

image

Did you move this `.vipb` file or have a stale `Source\ni.protobuf.types` directory on your hard disk?

dixonjoel: Yes, I moved it to https://github.com/ni/measurementlink-labview/tree/main/Source/gRPC/Generated%20APIs/ni/protobuf/types and forgot to update this. I see you have a PR to fix that.

PR #404, opened 2023-11-16
Finalize the measurementlink-labview palettes

pbirkhol-ni:

This PR finalizes the palettes. It does the following:
- Moves away from using NIPB for palettes. Instead, we are hand maintaining our own palettes. This solves a whole bunch of issues:
    - We control the naming of the palettes instead of NIPB. It is important to have consistent names for our library and class palettes.
    - We can put the palettes into source, which was not possible when using VIPB. Since our palettes have to be members of libraries and classes, they were always listed as missing in the project when using VIPB for palettes. This had real ramifications (can't build exes, potential increase in PR build times, etc). This is no longer an issue when the palettes are in source.
    - The VIPB editor is really buggy and frustrating to use.
    - More that I won't detail here. 
- Adds documentation to all sub palettes.
- Finalizes palette icons
- Adds the driver subpalettes to our palettes for quick access.
- Single sources palette files by adding a pre build step to find all of the palette files in source and put them in the proper directory structure for VIPB to harvest. This was done because VIPB doesn't allow installing the same file to two locations. It also means that we are installing palette files multiple times to different locations. This is what allows us to have good right click menus, a good palette, and the ability to hide driver palettes if they aren't installed.
- Add a post install step to fix up the .lvlibs that have palette files. VIPB somehow mangles the palette path during installation, which breaks the right click palette menu.

Tested in LabVIEW 2020 and LabVIEW 2023. This change also makes it easier to test palette files without having to build a VIP. Do the following:
- Run `measurementlink-labview\Source\Build Specs\Build Custom Actions\Pre-Build Custom Action.vi`. This isn't strictly necessary, but it is the easiest way to find all of the palette files to copy.
- Copy the dir and single palette file from `measurementlink-labview\Source\Runtime\MeasurementLink Palette Files\Measurement` to `<labview>\menus\Categories\Measurement`
- Restart LabVIEW (or somehow reload the palettes
- Play around with the palettes.
- Run `measurementlink-labview\Source\Build Specs\Build Custom Actions\Post-Build Custom Action.vi` to cleanup.


Source/Build Specs/MeasurementLink Service.vipb
```diff
@@ -1,7 +1,7 @@
-<VI_Package_Builder_Settings Version="2020.1" Created_Date="2023-01-23 14:05:21" Modified_Date="2023-11-14 15:14:33" Creator="lvadmin" Comments="" ID="d6a8d70e3546d75f7536825b8c1e1299">
+<VI_Package_Builder_Settings Version="2020.1" Created_Date="2023-01-23 14:05:21" Modified_Date="2023-11-16 10:10:39" Creator="lvadmin" Comments="" ID="8e7965136f751ce2ff2ce750b3189051">
   <Library_General_Settings>
     <Package_File_Name>NI_MeasurementLink_Service</Package_File_Name>
-    <Library_Version>1.3.0.4</Library_Version>
+    <Library_Version>1.3.0.2</Library_Version>
```
pbirkhol-ni, line 4: This version has been incremented unintentionally for various reasons. I put it back to the version it should be at.

dixonjoel:
Looks to me like running the G tests took approx. half the time as before.

dixonjoel (approved):

dixonjoel:
Is this expected when the package is not installed?
image

pbirkhol-ni:
> Is this expected when the package is not installed? image

Yes. The palettes link to VIs in vi.lib so they will not work for development. I don't know if that is something we can easily fix or not.

dixonjoel:
Why is this measurement context wire thicker than, e.g. the session reservation wire?
image

pbirkhol-ni:
> Why is this measurement context wire thicker than, e.g. the session reservation wire? image

Joel and I discussed this offline. The refnum wire gets thicker when the terminal is set to dynamic dispatch. I had no idea this was a thing and I'm not sure I like it.  :D

jasonmreding (approved):

PR #410, opened 2023-11-16
Use type definition name of actual DoubleXYData .ctl

dixonjoel:

Add a dependency to the NI gRPC Types VI package >= 1.0.0.1 to Measurement Service.vipb.

Ensure customers have the NI gRPC Types package installed as needed.

Interactive testing with a LV service.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Type Definition Name To Message Name.vi
jasonmreding, line 1: You could simplify this with a regular expression like so:

image

jasonmreding, line 1: - For consistency with the rest of the code, your controls/indicators on the block diagram should have labels on the left/right respectively rather than on top.
- I would make the `type definition name` input parameter required. You may also want to clear out the default value currently assigned so that it's an empty string.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Is Supported Cluster.vi
jasonmreding, line 1: image

Consider extracting the above into a sub VI and possibly making it a FGV so you only have to build the data once. Then move the call inside the No Error case of the first case structure. You may event want to change the data type to a set so you don't have to perform a linear search for the lookup.

jasonmreding (approved):

Source/Build Specs/MeasurementLink Service.vipb
```diff
@@ -19,6 +19,7 @@
     <Package_Dependencies>
       <Additional_External_Dependencies>ni_lib_labview_grpc_library =1.0.0.1</Additional_External_Dependencies>
       <Additional_External_Dependencies>ni_lib_labview_grpc_servicer =1.0.0.1</Additional_External_Dependencies>
+      <Additional_External_Dependencies>ni_protobuf_types &gt;=1.0.0.7</Additional_External_Dependencies>
```
pbirkhol-ni, line 22: Brad is changing this back to 1.0.0.1 in another PR so you should change this version here as well

dixonjoel:
@jasonmreding @pbirkhol-ni I find myself reconsidering whether this is better than what we have. A couple drawbacks are that it's just more complicated logic and it also gives us a compile-time dependency on the NI gRPC Types package. I'm not totally convinced it's better than this simple case structure logic that we can expand over time.
image

What do y'all think?

PR #413, opened 2023-11-17
Add a VISA plugin for generic VISA support

pbirkhol-ni:

- Add a MeasurementLink VISA library with all of the normal APIs plus a palette file. There is one main difference between this API and other driver APIs:
    - Both the Initialize and Get Connections VIs require an instrument type id input. Since we are making a generic VISA API, the customer gets to define their own instrument type id in the pin map so we can't know what it is unless they tell us.
- Add a VISA ISessionFactory. This has a few differences from normal ISessionFactory implementations
    - The instrument type id is not a static value so it must be passed in during construction and then it gets returned by a property.
    - The `Close MeasurementLink Session.vi` ignores the `initialize and close session behavior` input. It just unconditionally calls `VISA Close` because the close behavior is defined during init.
    - The `VISA Open for gRPC Session.vi` has an `Open / Close Behavior` input that has an `Auto` enum value. However, the ISessionFactory `initialize and close session behavior` enum does not have an `Auto` value. What this means is that we never pass `Auto` to `VISA Open for gRPC Session.vi`, which I think is okay but I can't 100% convince myself.
- Add all of the standard tests that we write for all of our plugins. Some of them are disabled for two main reasons:
    - They require a pre-release VISA driver on spawningpool. My preference would be to wait for the official release before we install it.
    - Even with the prerelease driver installed, however, a few tests still fail due to [AB#2582855](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2582855). Once that bug is fixed, those tests _should_ pass, although I had to take an educated guess on what error code we should be expecting.

Auto tests.


Source/Runtime/MeasurementLink VISA/Session Factory/Session Factory.lvclass
pbirkhol-ni, line 1: Now that I think about it, I will probably have to override `Clean Up After Init Error.vi`. For example, if you Initialize VISA with `Initialize and Detach` and another driver init fails, the default cleanup will simply close the session, which will detach. However, that will leave the grpc-device server in a different state than it started. What should happen is that it closes the session (which will detach) then reinitializes it with the `Attach and Close` option and then closes the session. That will leave everything in the same state as before.

pbirkhol-ni: Hmmm....this is pretty straight forward for 3 of the cases. For `Initialize and Close`, `Attach and Detach`, we just call `Close MeasurementLink Session.vi` and it will do what we want. For `Initialize and Detach`, we do what I described above. However, for `Attach and Close`, it is unclear what to do. I can certainly recreate the session, but it almost certainly won't be recreated with the same configuration as before. I don't think there is any easy way for us to "undo" in that case. Which makes me wonder whether we should attempt to "undo" in any of these cases.

pbirkhol-ni: Jason and I discussed this on a call. We should cater the behavior of `Clean Up After Init Error.vi` to work well with our process setup and process cleanup. So if there is an error during `Initialize and Detach`, we should Close the session. If there is an error during `Attach and Close`, we should also Close the session. This will give customers the best chance of success when using MeasurementLink.

jasonmreding, line 1: nit: This factory class has a light purple banner. All of the other instrument factory classes use a green banner.

Source/Runtime/MeasurementLink VISA/Initialize Sessions - 1Sess.vi
jasonmreding, line 1: Do you think the context help should include any information about the `instrument type id` parameter? This would apply to all of the poly VI instances.

Source/Tests/skip_tests.lst
jasonmreding, line 1: Looks like [AB#2582855](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2582855) is pending validation now if you want to verify everything works as expected before check in. If you are planning to do this in a follow up PR, that is fine too.

jasonmreding (approved):

Source/Runtime/MeasurementLink VISA/Session Factory/Close MeasurementLink Session.vi
jasonmreding, line 1: Should the `VISA Close` not also be wrapped in a call by reference node? Or is `VISA Close` a primitive installed as part of LV and the gRPC open variant is a VI installed by the driver even though the regular open is a primitive installed by LV?

Source/Tests/skip_tests.lst
pbirkhol-ni: That bug fixed the tests so I reverted this file.

Source/Runtime/MeasurementLink VISA/Session Factory/Close MeasurementLink Session.vi
jasonmreding: Primitives are part of the LV binary and are always installed. The biggest difference between a primitive and a VI is that primitives can adapt to type at edit time and do other fun polymorphic things with terminal data types. That's what originally led to the introduction of XNodes and subsequently malleable VIs. Otherwise, you are sort of forced to create poly VIs with instance VIs for each data type you want to support. But of course you can't do that for an infinite number of types.

I'm pretty sure there is no way to call a primitive by reference. I'm not even sure what that would mean. The only thing you could do is wrap it in a VI and then call that. I guess the question is how does the primitive behave if VISA isn't installed. Does the primitive break the diagram at edit time? If so, we might have to resort to doing this sort of thing. I don't think the data types for any of the terminals for the primitive(s) in question are dynamic?

pbirkhol-ni: If LabVIEW is installed, VISA Close is also installed. The same is not true for `VISA Intialize for gRPC`. So I think what I have now is correct.

Source/Runtime/MeasurementLink VISA/Initialize Sessions - 1Sess.vi
pbirkhol-ni: Sure. How does this sound? 

Create a VISA session from the <b>session reservation in</b> and the <b>instrument type id</b>. The <b>instrument type id</b> must match the Instrument Type Id for the custom VISA instrument in the pin map.

Source/Runtime/MeasurementLink VISA/Session Factory/Session Factory.lvclass
pbirkhol-ni: I am going to make this change in a follow up PR. It will likely involve modifying the interface and also could be a little controversial.

Source/Runtime/MeasurementLink VISA/Initialize Sessions - 1Sess.vi
jasonmreding: Sounds good.

PR #417, opened 2023-11-28
Cleanup After Init behavior should better match MeasurementLink workflow

pbirkhol-ni:

`ISessionFactory` has an optional override called `Clean Up After Init Error.vi`. Historically, this VI has attempted to put the grpc-device server back into the same state it was in before initialize was called. However, this can cause issues with the MeasurementLink workflow. For example, during ProcessCleanup, we initialize sessions with Attach and Close. If that step fails, we still want those sessions to be closed. However, the current behavior would leave those sessions open on the grpc-device server. This would cause rerunning the TestStand sequence to fail when creating the sessions. This PR attempts to fix this workflow, although I think some of the decisions I made will be controversial. 

Below are the behaviors that I added to this PR:
- `Init and Close`: On error, don't initialize any more sessions. Call close to clean up any sessions that had been created before the error.
- `Init and Detach`: On error, don't initialize any more sessions. Call close to clean up any sessions that had been created before the error.
- `Attach and Detach`: On error, don't attach to any more sessions. Call detach to clean up any sessions that had been attached before the error.
- `Attach and Close`: On error, keep trying to attach to sessions. Once we have tried to attach to all sessions, call close on any sessions that were successfully attached.

The `Init and Close` and `Attach and Detach` behaviors aren't controversial. I also don't think `Init and Detach` is too controversial. Rather than detaching, we are closing the sessions so that the instruments are in the same state as before we called Initialize. 

For `Attach and Close`, I modified `Initialize Sessions Impl.vi` such that we try to attach to the session even if a previous attach had failed. This gives us the best chance at successfully cleaning up all sessions so that when we run again it doesn't fail because the session still exists on the grpc-device server. However, it could lead to issues where all sessions fail to Attach and so it takes a long time to fail.

Automated tests. I modified some existing automated tests rather than write new ones. We already had existing automated tests for the various init/close behaviors. I just added more sessions to these tests to verify that in the `Attach and Close` case that we try to attach to all sessions, even after failure.


Source/Runtime/MeasurementLink Session Management/Session Reservation/Clear Error For Attach and Close.vi
jasonmreding, line 1: Not sure what's going on here. I pulled down your branch, but I don't see this change. Similarly, I see other changes that aren't present in this view (e.g. `ISession Factory.lvclass:Cleanup Up After Init Error.vi`). Any idea what's going on here?

Source/Runtime/MeasurementLink Instrument/ISession Factory/Helpers/To Cleanup Initialize and Close Session Behavior.vi
jasonmreding, line 1: nit: The comment about Attach and Detach not being common and not used in our work flows is a bit misleading. It is the common case for the measurement logic when running measurements from TestStand.

pbirkhol-ni: Agreed. This comment has changed in the latest revision. Do a `git pull`.

Source/Runtime/MeasurementLink Session Management/Session Reservation/Initialize Sessions Impl.vi
jasonmreding, line 1: I didn't see any logic here that would continue to try and initialize sessions for the `Attach and Close` case. I imagine this will be pretty ugly to implement and is the only thing that gives me some pause with these changes. I think the existing code would need to be refactored with some logic pushed into sub VIs to try and do this in a sane/readable way.

jasonmreding (approved):

PR #422, opened 2023-11-30
Fix error in getting array element type

dixonjoel:

I discovered that array elements were returning TypeUnknown for a LabVIEW service. This was a mistake in earlier refactoring.

I refactored the logic between Get Metadata from Configurations and Get Metadata from Results so they re-use the code that fetches the parameter information from the cluster element.

Fixes LabVIEW services with arrays. Simplifies logic in Get Metadata.

I'll build the packages and test with them to ensure what I was seeing is fixed.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Configurations.vi
dixonjoel, line 1: Before:
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Results.vi
dixonjoel, line 1: Before:
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get MetaData from Configurations.vi
dixonjoel: After:
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Get Parameter Information.vi
dixonjoel, line 1: image

pbirkhol-ni: The danger in using variants. Not saying we have a choice here, but I've done similar things many times. The coercion dot has usually alerted me.

dixonjoel:
> I wish we had more tests in areas like this. This code is definitely testable. I am not going to require tests to submit this PR, but if you want to take a shot at writing some tests, I won't stop you. :)

How do we handle the fact that these are private VIs inside the library when we write tests?

pbirkhol-ni:
> > I wish we had more tests in areas like this. This code is definitely testable. I am not going to require tests to submit this PR, but if you want to take a shot at writing some tests, I won't stop you. :)
> 
> @pbirkhol-ni How do we handle the fact that these are private VIs inside the library when we write tests?

Make them community scope and then add the test library as a friend.

PR #428, opened 2023-12-04
Create PR comments with images for changes to G code

aepete:

When RUN_G_DIFF_CHECK is specified in PR description, the automated diff process now produces images that highlight VI differences for each changed VI, then automatically publishes them to file-scoped comments.

Graphical diff facilitates both code authors and reviewers by automatically highlighting the functional changes at hand, rather than requiring a tedious manual process of capturing screenshots for those who elect to do so.  The red (before) and green (after) ovals also help reviewers focus on precisely what about the G code has changed.

#### Implementation detail
The diff images are initially produced by LabVIEW's comparison method, then several image processing utilities proceed to optimize the images for viewability.  The idea here is to essentially merge / overlay similar diff images in such way as to not overwhelm the viewer a massive dump of before & after images.  Results are pretty good as-is and can be further refined later.

In order to render the images in the PR, the .png files first needed to be posted to an accessible hosting location.  Per https://stackoverflow.com/questions/18360714/official-image-host-for-github-projects, there is no clear hosting location to use, nor does github document how to replicate its own hosting of users' copy-pasted images...  Previous NI solutions had posted files to the repo, which was also deemed undesirable here.  _Potentially controversial?_: the proposed solution is to post & keep the images nearby, tucked away as assets in [a long-forgotten pre-1.0 release](https://github.com/ni/measurementlink-labview/releases/tag/v0.12.1)

- Used #419 as a workspace to make incremental progress on both the imagery and the workflow / PR sides of this implementation, which involved substantial manual testing.
- Ran diff utility on other sets of contrived G changes with mostly good results.
   - Perhaps understandably, the merging implementation can struggle to optimize cases where there are stacked changes living in the same coordinate space of the same diagram (e.g. changing overlapping code in two otherwise-similar cases of the same case structure).


.github/workflows/run_pr_commenter.yml
aepete, line 1: Since this diff runs quicker than the test, perhaps we should list diff first in the pr workflow..?

Source/Tools/run_diff.vi
jasonmreding, line 1: Is there anything we can do so that clicking on the image opens the picture directly in the browser (probably a new tab/page) rather than downloading a file? The images are small enough when embedded here that I don't imagine the diagrams will ever be readable enough without enlarging (except for the simplest VIs). If that's the case, then maybe it would be better to just put the before and after side by side in one picture and expect that people will need to open them in a new tab anyway?

.github/workflows/run_pr_commenter.yml
pbirkhol-ni: They run in parallel, right? If so, then I don't think it matters.

Source/Tools/run_diff.py
```diff
@@ -135,6 +149,47 @@ def post_github_pr_text_comment(text, pr_number, token):
     return response.status_code
 
 
+def post_github_pr_file_scoped_comment_with_images(file_id, directory_with_images, pr_number, token, commit_id):
+    # First, upload all the images.  Since PRs do not support assets, this code uploads images
```
pbirkhol-ni, line 153: This is unfortunate. I did some googling, and while I found some alternative solutions, none of them are obviously better than what you have here.

Source/Tools/run_diff.vi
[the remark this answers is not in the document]
pbirkhol-ni: Is the 3rd picture an example of a merge /overlay? In general, if I saw this image, I would probably open the VI and look at it manually because I can't make sense of what's happening. I realize that will have to happen occasionally, especially for larger refactors. I'm unsure if this is one of those cases.

Source/Tools/subVIs/Parse Diff Report to Text.vi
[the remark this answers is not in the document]
pbirkhol-ni: FWIW, I like this diff. It's a simple change and it is very obvious what has changed.

pbirkhol-ni (approved):
I have no real concerns. I'm not really sure what you mean by a merge/overlay diff. The pessimist in me thinks that a merge/overlay might cause more trouble than it's worth. But we can give it a shot.

Source/Tools/run_diff.vi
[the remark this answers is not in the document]
aepete: The merge/overlay produces as few images as it can for both the before case and for the after case.  For this particular case, it produced 1 image (the top-most) for before and 2 images for after.  FWIW, you get a hint about this if you hover over the picture.

I'm pretty sure what happened here is that the "after" picture that LV gave us for one of the changes -- perhaps the deleted content -- didn't graphically align with the rest of the after pictures. This is probably inevitable for larger diagrams since we're leaving it in the LV compare method's hands.

I guess we could try to optimize out any picture that doesn't contain a diff oval...?  That's probably easy-enough and reliable enough to detect, and clearly it's a bit confusing if the automated differ is posting images that have neither green nor red.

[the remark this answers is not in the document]
aepete: I could also make some improvements to how the images get formatted within the comment (e.g. a table with columns for before & after and rows for diagram & panel), though I had planned to revisit that in a possible later PR.

aepete: I suspect I can modify the way we script the comment content so it'll open in a new tab.  Lemme try out a few ideas.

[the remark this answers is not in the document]
pbirkhol-ni: I almost want to say that if there are multiple images for a diff that we just completely bail. I'd rather have no diffs than confusing diffs. However, I don't know how common multiple images are.

aepete:
> I have no real concerns. I'm not really sure what you mean by a merge/overlay diff. The pessimist in me thinks that a merge/overlay might cause more trouble than it's worth. But we can give it a shot.

The post-processing is actually pretty essential to producing images that are easily & helpfully digestible.  Effectively, LV gives us two pictures for every individual line-item change.  As an example, the number of images produced by the LV compare method for this very set of changes to `run_diff.vi` would otherwise be 64 .png files (32 before & 32 after).  The collate step takes it down to 38 (19 before & 19 after).  Then the merge drops it to 3 (1 before, 2 after).

dixonjoel:
Have you seen documented if there's any limit to the number of assets on the release that we're using? As long as it's a reasonable limit, I guess we could come in once in a while and delete the older assets, but then I guess we'd lose some history on the PRs. Anyway, concerned about the long term usage of that release.

Source/Tools/run_diff.vi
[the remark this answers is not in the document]
dixonjoel: What's with the weird rendering of the rectangular area on the arrows?

dixonjoel (approved):

Source/Tools/run_diff.vi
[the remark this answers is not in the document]
aepete: I guess LabVIEW draws a bounding rectangle for those arrows.  Not sure why.

aepete:
> Have you seen documented if there's any limit to the number of assets on the release that we're using? As long as it's a reasonable limit, I guess we could come in once in a while and delete the older assets, but then I guess we'd lose some history on the PRs. Anyway, concerned about the long term usage of that release.

Github doesn't appear to have any limits on release assets that would hinder this kind of usage: https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas

Source/Tools/run_diff.vi
aepete: Joel and I took a closer look at this.  So far, there doesn't appear to be a clear means to reference the image within a release assets directory other than the download URL.

A whole different approach would be to host the images elsewhere (other than release assets), e.g. https://stackoverflow.com/questions/61537403/how-to-host-image-at-https-user-images-githubusercontent-com-path-filename .  I don't think we want images stored as files in the repo itself...  And I've not found a documented way to directly upload to githubusercontent as is done when pasting an image into a PR.

pbirkhol-ni: I saw a stack overflow post yesterday that suggested creating a new repo for PR images. Maybe we can create an NI assets repo just for this purpose.

.github/workflows/run_pr_commenter.yml
aepete: From observation, they appear to be serialized.  But I don't think there's a guaranteed order, unless we were to add a dependency.  Probably fine to leave it as is, can revisit it it's troublesome.

aepete:
This review stands in an approved state, and all open ideas / concerns in the comments have either been addressed or added as follow-up tasks to the associated work item.  I'm hereby submitting this PR.

I encourage developers to evaluate this diffing capability as they author new G changes; please let me know if anyone encounters a diff result that could be refined or improved upon!

PR #429, opened 2023-12-04
Update old examples to use new Run Service.vi

pbirkhol-ni:

@jasonmreding changed how `Run Service.vi` works in #267. However, we didn't update the examples to use the new recommended way. This PR updates the two remaining examples that still use the old `Run Service.vi`. It also adds (and hides) the `MeasurementLink Server User Settings` control to the DCPower example so that we can run it in our ATS.

Ran examples locally to make sure they work.


pbirkhol-ni:
The front panel and block diagram now look like this:

image

jasonmreding (approved):

Issue #397, opened 2023-11-13
test_PinMapWrapper_StopPinMapService_WrapperMethodsWorkWithoutError is in the skip list


dixonjoel:
# Bug Report

test_PinMapWrapper_StopPinMapService_WrapperMethodsWorkWithoutError is going in the skip list for tests. It apparently puts the services into a bad state and causes other tests to fail.

## Repro or Code Sample

Run test_PinMapWrapper_StopPinMapService_WrapperMethodsWorkWithoutError and then test_CreatePinmapWith3Pins_QueryPins_3PinsReported.vi quickly in sequence. You'll get a -1002 stream removed error.

## Expected Behavior

Tests should all pass, regardless of running order.

## Current Behavior

As stated above

## Possible Solution

Clear out the discovery client FGV before running the second test.

[AB#2578870](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2578870)

PR #431, opened 2023-12-05
Close the Discovery Client before restarting the Discovery Service

pbirkhol-ni:

#397 

The following sequence will fail with a stream removed error:
- Start discovery service
- Create a discovery client
- Resolve a service
- Restart the discovery service
- Create a discovery client
- Resolve a service ---> Fails with stream removed

The reason why it fails is because the first discovery client doesn't get closed. If you close the discovery client before restarting the discovery service then there is no issue. You can also simply wait a few seconds after the restarting the Discovery Service and everything will work, even if you don't close the client. In our PR build, we have two tests that run consecutively that effectively did this sequence of steps. This PR fixes the issue by closing any discovery clients that are open in the `Discovery Client FGV.vi`.

Will this be a problem for customers? I don't think so. All of our accesses to the Discovery Service happen using an API that will retry on error. 

What is happening? I'm not entirely sure. I looked through the grpc-labview code and it keeps a map of grpcIds, each with its own channel. Since the first and second clients have different clientIds, I wouldn't expect that they would conflict. My guess is that the underlying C++ gRPC implementation has a channel cache that somehow gets invalidated and that effectively invalidates the channel for the new grpc session as well.

Tests now pass when run consecutively.


Source/Tests/Tests.Runtime/Measurement Server/Helpers/Restart Discovery Service.vi
pbirkhol-ni, line 1: image

dixonjoel (approved):

jasonmreding (approved):

PR #433, opened 2023-12-06
Rename/refactor of MeasurementContext.lvclass

jasonmreding:

- Modify rename/refactor scripting code so that it handles dependencies that link directly to the renamed class/library rather than to files within the class/library.
- Renames `MeasurementContext.lvclass` to `Measure Call Context.lvclass`. The path on disk was modified to match, and the class is no longer located in a "Classes" folder.
- Renames `Read Pin Map Context.vi` to `Get Pin Map Context.vi`.

Note: palette files are not updated with this change nor or control/indicator names that link to the renamed class. These changes will be made in follow on PRs.

Changes to the examples and template was performed by the scripting utilities. Changes to the server project was made by the LV IDE when renaming and moving files via the files pane in the project.

Promotes greater consistency of file names/paths.

Manually opened all example projects and verified `Run Service.vi` isn't broken.


pbirkhol-ni (approved):
In this case, opening just the `Run Service.vi` isn't good enough since the `Measurement Logic.vi` is the likely VI to be broken. I did a handful of sanity checks on `Measurement Logic.vi`, though, and everything looked good. I'm not doing thorough checks for now since we'll do example testing at the end.

PR #434, opened 2023-12-06
Skip VISA tests for LV 2020

MounikaBattu17:

VISA tests uses VISA 2024 Q1 which doesn't support LabVIEW 2020. Thus skipping the VISA tests for LV 2020 seems to be the right thing do. So I have made the following changes in `run_tests.vi`.
- The VISA tests are identified in the basis of "`VISA`" string in test name.
- I have used Application Version property node and matched pattern with string `20.0` to find the LV version on which the VI is running.
- Added a case structure to search for `VISA` in the test case name.
- Finally, `AND` gate to validate the resulting booleans and skip tests. So the tests will be skipped if either they are added in `skip_tests.lst `or the test name contains VISA when running in LV 2020.

Here's the diff of old and new run_tests.vi
image
new1

Ideally, VISA tests should be skipped in LV 2020 as it is not supported. Also, the upcoming measlinkbot has higher versions which can be used to run VISA tests.

Manually ran tests in LV 2020, 2023 and observed VISA tests are skipped in LV 2020 alone.


dixonjoel (approved):

Source/Tests/run_tests.vi
pbirkhol-ni, line 1: You should make the version regex 20\.0 so that it only matches 20.0. As is, the . will match any character.

pbirkhol-ni, line 1: Skipping these tests is fine, but not until we have a test runner that uses a different version of LabVIEW. As far as I know, we only have `spawningpool`, which only runs tests on LabVIEW 2020. If we have other test runners, we should add them to the PR build before submitting this PR. If we don't have other test runners, we should hold off on submitting this PR.

pbirkhol-ni, line 1: Looking for `VISA` in the test name is probably too fragile for a long term solution. However, I anticipate that we'll be upgrading the minimum LabVIEW version to 2021 in the near future so this should be okay for now. We have considered splitting our tests into more libraries, which could make this type of filtering easier. Anyway, no action item for this PR.

pbirkhol-ni (changes requested):
See comments for requested changes.

Source/Tests/run_tests.vi
dixonjoel: If spawningpool is running the tests on LV 2020, are they passing now? I'm confused.

MounikaBattu17: Some VISA VIs are manually copy pasted in spawningpool to run tests in LV 2020. That's why the VISA tests are passing in spawningpool.

MounikaBattu17: >If we have other test runners, we should add them to the PR build before submitting this PR. If we don't have other test runners, we should hold off on submitting this PR.

@pbirkhol-ni Currently PRs are ongoing for measlinkbot which has LV 2020, 2021, 2022, 2023. So once the PRs are approved, measlinkbot with labview setup can be deployed to AWS and the same can be hosted as a runner here. The issue in holding this PR until the above steps happen is that we run tests as a part of build stage of measlinkbot, so currently the LV 2020 VISA tests are failing in the build stage. Without this change the build stage of measlinkbot cannot succeed. There are few shortcuts that can be made for the measlinkbot build stage to succeed.
1) We can copy paste the VIs manually in measlinkbot also
2) Currently, the tests are running from main branch source code, instead we can change the target branch to this branch and run tests.

Note: These changes can only be temporary and should be reverted back once the runner is set up here.

Please share your thoughts.

pbirkhol-ni: You're change looks reasonable. However, I'd rather the logic be something like this:
```
if 20.0 and not spawningpool:
    if not VISA:
        run test
```

After you make that change, go ahead and submit.

PR #439, opened 2023-12-06
Upgrade to grpc-labview 1.0.1.1

pbirkhol-ni:

This PR upgrades grpc-labview to 1.0.1.1. Most of the changes are just regenerating the code, so you can ignore those. There are a few changes to our infrastructure that are required for this upgrade.
- Modify `Register Configuration and Results.vi` to supply the enum metadata to grpc-labview.
- Modify `Validate Metadata.vi` to make sure the enum strings are valid. Enum strings cannot use a ";" or a "=". This is because of how grpc-labview does tokenization of enum metadata. This isn't ideal, but also doesn't seem like too big of a restriction.

I wrote some tests for this change. We haven't written any tests for the `Measurement Plugin Service` class yet, but I followed the new organization paradigm that @jasonmreding established. I also created a fake measurement plugin that we can use to inject plugin VIs for writing more tests in the future.

There are other solutions to this problem, some of which I will detail below:
- Have grpc-labview treat enums as integers. This would require a small modification to grpc-labview. I am not sure if the grpc-labview team would accept this change. It would be opt-in, of course. By default, grpc-labview would treat enums like enums.
- Change the grpc-labview enum tokenization to use a standard format like json. This seems possible, but would require a lot more work.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Register Configuration and Results.vi
pbirkhol-ni, line 1: This added code registers enum metadata with grpc-labview.
image

Source/Runtime/MeasurementLink Measurement Server/Helpers/Validate Enum Strings.vi
pbirkhol-ni, line 1: image

pbirkhol-ni:
I think the PR build will have lots of failures until we update the version of grpc-labview on spawningpool. Ignore the failures for now.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
jasonmreding, line 1: nit: I would change "Enums strings" to "Enum strings" as that reads better to me.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Validate Enum Strings.vi
jasonmreding: nit: you could move the call to `1D String Array to Delimited String.vi` inside the False case of the case structure.

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
pbirkhol-ni: That was just a typo

pbirkhol-ni:
I made the changes Jason suggested and also wrote some tests. I am going to reset reviewers.

Source/Runtime/Discovery Client.lvproj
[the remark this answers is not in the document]
pbirkhol-ni: https://github.com/ni/grpc-labview/issues/332

PR #441, opened 2023-12-06
Rename/refactor of session management and discovery libraries

jasonmreding:

- Update refactor utilities to properly handle nested typedefs. Only the top most typedef instance can be modified.
- Update refactor data for rename/refactor scripting utilities.
- Renames the following libraries
  - MeasurementLink Discovery.lvlib -> MeasurementLink Discovery V1.lvlib
  - MeasurementLink Session Management.lvlib ->MeasurementLink Session Management V1.lvlib

Changes to the server project were made by the LV IDE when renaming and moving files via the files pane in the project. Changes to the examples and template was performed by the scripting utilities. I then hand modified the Keysight example to fix broken references to malleable VIs. I couldn't figure out a way to get the original name/path of a missing malleable VI.

Promotes greater consistency of file names/paths.

Manually opened all example projects and verified Run Service.vi and the TestStand code module VIs aren't broken.


Source/Runtime/Measurement Server.lvproj
pbirkhol-ni, line 1: (This comment is here just for threading purposes).

This looks good. A few comments:
-  Do you think we should move the Connection Map and IReservationObserver classes out of the library? They seem like they are pretty coupled to the session management service so having them in the library probably makes sense. But I can imagine a scenario where we would need to rev the Session Reservation class but not the Connection Map class. In that case, would we just be creating a copy of the Connection Map?
- Should we rename `IReservationObserver` to be `IReservation Observer`? This can be done in another change. It's also private, so we could just not do it.
- Do we need to version `Measure Call Context` as well since it references the session reservation class?

PR #451, opened 2023-12-11
Update grpc-labview dependency for the VIPBs

pbirkhol-ni:

We recently updated the grpc-labview version to 1.0.1.1. This PR updates the dependency in our VIPB files.

Note that I made the dependency `>=` instead of `==`. I really wish VIPM supported more options so that we could properly implement semantic versioning. I'd like to do something like `1.0.0.0 <= grc-labview <= 2.0.0.0`. However, that is not possible in VIPM.


Source/Build Specs/MeasurementLink Generator.vipb
jasonmreding, line 1: Did we update the generated code for xydata.proto? Should we? I don't know if the rich/flat wrapper VIs are required or not with the newest runtime.

Should `ni_protobuf_types.vipb` have a dependency on grpc-labview as well? It looks like Joel deleted the generated code for message registration so it's probably not strictly needed.

jasonmreding (approved):

Source/Build Specs/MeasurementLink Generator.vipb
pbirkhol-ni: I looked into adding a dependency on grpc-labview but didn't see a reason why it was needed.

As far as regenerating the code goes, I'm not sure. The Game of Life example works without regenerating it. However, I can regenerate for consistency if you think it's necessary. I'm not sure there would be any difference, though.

PR #452, opened 2023-12-11
Fix Fake Measurement Plugin.lvclass

jasonmreding:

The recent case only renames to some files broke the `Fake Measurement Plugin.lvclass`. I'm not sure why the PR didn't catch this, but this change fixes the class so that it contains dynamic dispatch members with the correct name again.

Fixes lvclass file.

Class is no longer broken when loaded in the editor.


pbirkhol-ni (approved):
> I'm not sure why the PR didn't catch this

It's because I forgot to name those tests correctly. Can you prepend `test_` to the test names in `Measurement Plugin Service Tests.lvlib`?

PR #459, opened 2023-12-14
Example on how LV can call MeasLink Server. A demo Server is included.

yardov:
- [ ] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-labview/blob/main/CONTRIBUTING.md). _(Required)_

An example in LV 2023 that shows how to call into a Measurement Link Server.
The example shows how to call the discovery service to get all the services currently running. Then it takes the one selected by the user and opens the measurement service. Sends the configuration and receives the results. These need to PERFECTLY match the service you are about to use or else you get an error.

To change the data, there are two typedefs; Configuration and Results.

Shows customers how to use LV to call Measurement Link

Testing with the demoed service only.


Source/Example Measurements/LV CallingMeasLink Measurement/Demo LV Measurement Server/DemoLVService.lvproj
pbirkhol-ni, line 1: All of our LabVIEW code is still in LabVIEW 2020. Even if we wanted to submit this code as is, we couldn't because our PR build is almost certainly going to fail.

pbirkhol-ni: It looks like the PR build passed. I think that is because we don't do anything with the examples as part of the PR build. Nevertheless, I'd prefer not to submit LV 2023 code.

yardov: I was afraid of this process but it went smooth. I was able to update all the files to LV2020 and tested them.
However, my instrument Studio is not able to instantiate the measlink measurement anymore:
image

pbirkhol-ni: I did not see this issue. I also rewrote a lot of the example. You should try again.

pbirkhol-ni:
I just made a lot of changes to this example. I want Gerardo to give his approval before submitting.

dixonjoel (approved):
@pbirkhol-ni You can check this in or wait until after the new year at your discretion. We can exclude it from the zipped examples if it's not quite ready to be released.

pbirkhol-ni:
> @pbirkhol-ni You can check this in or wait until after the new year at your discretion. We can exclude it from the zipped examples if it's not quite ready to be released.

I'm just about to push the example changes, which are the last of the changes. I used the script to update the examples and sanity tested a few of them. 

Thanks for the review.  :)

Source/Example Measurements/LV CallingMeasLink Measurement/Demo LV Measurement Server/DemoLVService.lvproj
yardov: I tested the latest changes and no issues:
- Tested Server in IS.
- Tested Server called by Client
- Tested in LV 2020

Thank you for all the changes, it's a much better example.

yardov:
I actually might have a comment on this @pbirkhol-ni :

image

If you do not include a package name, when scaling this into calling multiple servers, the messages can be mixed if they are all called 'measurement configuration' and 'measurement results'. We might want to add a package name.

pbirkhol-ni:
> I actually might have a comment on this @pbirkhol-ni :
> 
> image
> 
> If you do not include a package name, when scaling this into calling multiple servers, the messages can be mixed if they are all called 'measurement configuration' and 'measurement results'. We might want to add a package name.

I did remove the package name. Honestly, I didn't put much thought into it when I removed it, but I'm not sure I understand the issue that you are worried about. A client talks to a single server. If you want to talk to multiple servers then you should use multiple clients. In that case, I believe each client can have similarly named message metadata without issues.

Are you talking about the case where a single server has multiple services? That seems like a situation where you would have a lot of issues. If a server has multiple, unique services then that seems fine. But I don't know what would happen if you ran multiple measurement services on the same server. It's certainly something I have never tried.

As far as I know, the only reason for registering the `measurement configuration` and `measurement results` clusters is to serialize and deserialize from Any. So if you had multiple configurations or multiple results, that could be an issue, but I don't know how that could happen.

PR #479, opened 2024-01-25
Add a link to the supported datatypes documentation

dixonjoel:

Update the error message for an invalid data type to point to the online help page enumerating the allowed types.
image

Clarify the user's options when they use an invalid data type.

None


Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
pbirkhol-ni, line 1: How difficult would it be to add the data type next to the control/indicator name? For things like rings and enums, it may not be obvious what we mean when we say that the data type isn't supported. Is it that the enum itself isn't supported? The underlying type of the enum? If we added the data type, it might be clearer. We could also just add a blurb like : `For enums and rings, the underlying representation must be a supported data type.`

pbirkhol-ni, line 1: I talked to Nathan Stokes and he suggested we use a redirect URL. https://www.ni.com/r/measurementlinkdatatypes

Nathan has created the URL and expects it to go live within 24 hours.

pbirkhol-ni (approved):

Source/Runtime/MeasurementLink Measurement Server/Helpers/Error Strings Global Internal.vi
dixonjoel: We could use 'Get Type Information.vi' to easily get the 'type', but out of the box it just gives you the top-level type. E.g. 'Array'. If that's as far as we want to go, that's easy. However, if we want to get a user-consumable text equivalent of what you get when you hover over a wire in LabVIEW, that's harder. I imagine we'd only want to go as far as, e.g. 'Array of I32' or 'Cluster'. The enum case would say 'Enum 16' or similar. I'll take a stab at including the current type in the error message.
image

dixonjoel: How about this?
image

(Scroll down in the dialog)...

image

dixonjoel: Maybe 'not supported by the Measurement Service' -> 'not support for a measurement service'.

dixonjoel: Oh, also, previously, there was simply an error out indicator on the Run Service.vi. I like the experience better of showing the Simple Error Handler.vi for this case so it's right in their face. So I'd change the template for Run Service to have the Simple Error Handler in it.

pbirkhol-ni: I like the paragraph explaining the error. I don't understand what this means:
`enum,my cluster`

What configuration were you using?

dixonjoel: That's just the labels. Should we make it more clear what those refer to?
image

pbirkhol-ni: We used to have a simple error handler, but we removed it because it pops a modal dialog that can hang automated tests. This wasn't just a hypothetical problem; it was hanging our TestStand ATS.

dixonjoel: We could say 'Measurement Configurations (labels)' or similar?

dixonjoel: It would only hang if the service were invalid, right? I guess that can happen during development, so maybe that's a non-starter. Or we find a way to disable the error handler during test runs.

pbirkhol-ni: > We could say 'Measurement Configurations (labels)' or similar?

No, I think that is fine. Without context, I wasn't sure if `enum` was supposed to represent a data type or the name of a control. It makes more sense with the context.

dixonjoel: Or if no error handler, at least we need to make the default size of the error out indicator's text field significantly larger.

pbirkhol-ni: > It would only hang if the service were invalid, right? I guess that can happen during development, so maybe that's a non-starter. Or we find a way to disable the error handler during test runs.

When you run a measurement and it errors, the front panel pops into focus with an error in the error indicator. That has always been obvious to me.

dixonjoel: Ok. Based on offline discussion with @pbirkhol-ni, I won't change the Run Service experience or add a simple error handler. Current iteration:
image

dixonjoel: You'll actually have to click on 'Explain Error' on the error cluster and then you'll get this:
image

dixonjoel: This redirect is working now.

jasonmreding: > We could use 'Get Type Information.vi' to easily get the 'type', but out of the box it just gives you the top-level type. E.g. 'Array'. If that's as far as we want to go, that's easy. However, if we want to get a user-consumable text equivalent of what you get when you hover over a wire in LabVIEW, that's harder.

I know how to get an image of the context help for the data type of the control. Sadly, I can't figure out a way to get the textual description displayed in that image.

PR #483, opened 2024-02-14
stubs: Update session management service proto

Jayaseelan-James:
- [x] This contribution adheres to [CONTRIBUTING.md](https://github.com/ni/measurement-services-labview/blob/main/CONTRIBUTING.md). _(Required)_
- [ ]  Automatically post PR comments with images for G code changes? _(Recommended for small changes)_

### What does this Pull Request accomplish?
The `session_management_service.proto` was updated to accommodate the multiplexer-related APIs.
[AB#2645217](https://dev.azure.com/ni/DevCentral/_workitems/edit/2645217)

### Why should this Pull Request be merged?
To keep the proto files and stubs in sync with their copies in ni-apis. This PR only updates the stubs. Any wrapper(s) and related VI changes should be performed once we plan to include multiplexer support in MeasurementLink LabVIEW. Having said that, I'm guessing that the generated stubs with the new APIs shouldn't impact any of the existing workflow.

### What testing has been done?
Existing tests should pass.

PR #487, opened 2024-02-27
fix: update VIs and library dependencies

Jayaseelan-James:

Reroutes the dependencies from `<vi.lib>` to repository VIs and libraries.

When we try building the `ni_measurementlink_service.vipb` without having its prior version installed. The following dependency errors occur.
image

image

- Verified that the build doesn't produce such dependency errors when `ni_measurementlink_service` is not installed priorly in the machine.
- Existing tests should pass.


jasonmreding:
Do we know what change broke this? I wonder how hard it would be to write a PR test that would prevent this sort of thing in the future?

Jayaseelan-James:
> Do we know what change broke this?

@jasonmreding The GitHub history leads to the #483 PR when updating the dependencies to fix pipeline failures. Here, LabVIEW took the vi.lib dependencies instead of the local ones.

> I wonder how hard it would be to write a PR test that would prevent this sort of thing in the future?

This sounds to be a good thing to do but I'm unaware of the amount of work and complexity it would involve at the moment. Shall I create a tech debt for it? So that we keep track of it and could possibly take it up in the near future.

Source/Runtime/Measurement Server.lvproj
```diff
@@ -141,6 +141,17 @@
 				<Item Name="niDMM Self-Test.vi" Type="VI" URL="/&lt;instrlib&gt;/niDMM/nidmm.llb/niDMM Self-Test.vi"/>
 				<Item Name="niDigital Self Test.vi" Type="VI" URL="/&lt;instrlib&gt;/niDigital/niDigital.llb/niDigital Self Test.vi"/>
 				<Item Name="niDCPower Self Test.vi" Type="VI" URL="/&lt;instrlib&gt;/niDCPower/nidcpower.llb/niDCPower Self Test.vi"/>
+				<Item Name="niDigital Get Session Reference.vi" Type="VI" URL="/&lt;instrlib&gt;/niDigital/niDigital.llb/niDigital Get Session Reference.vi"/>
```
Jayaseelan-James, line 144: @jasonmreding @dixonjoel The changes in this file are always added whenever I open `Measurement Server.lvproj`. Therefore, I'm tempted to ask if we can check-in these changes.

jasonmreding: The LV project persists the dependencies of the project that come from outside of the project (e.g. vilib), which stinks for SCC. So if a particular library is missing (because you don't have the driver installed or what not), it tends to result in a change to the list of dependencies that get persisted in the project. I personally try to just not commit project file changes if I know I haven't actually made meaningful edits to the project in order to reduce PR noise. Otherwise, you end up playing Whack-A-Mole with the project file since what each developer has installed on their development machine is frequently different. For example, I don't have every single driver installed on my dev machine so the project will always have some missing dependencies when it loads. This means the project almost always loads with a dirty dot for me. With that said, I recognize this is a difficult thing to stay on top of so I've mostly learned to ignore the noise.

Jayaseelan-James: Sure. I'll revert these changes then.

Jayaseelan-James:
@jasonmreding Please let me know if there are any more concerns to be addressed.

PR #490, opened 2024-03-04
Add most data types to the template for a measurement service

dixonjoel:

Modifies the template to showcase using most of the data types we support including XY data and arrays of XY data.
image

Some internal customers asked for clarification on how XY data can be used, so we decided to showcase it on the template / starting point for measurements.

Tested with the template locally to ensure it ran correctly and gave the correct (loopback) behavior for all inputs and outputs.


Source/Generator/Measurement Service Template.lvproj
jasonmreding, line 1: Anchoring comments to this file for threaded conversations:

Should we also include examples of path and pin controls? On one hand, I'm somewhat reluctant to promote paths as that gets weird if we ever support remote panels. On the other hand, we added the type for a reason so why not advertise it?

Adding a pin control would also require modifying `Get Type Specializations.vi`, and I'm not sure how much effort we want to make the user go through to remove stuff they aren't using. At least I presume they will get an error if they have an annotation for a parameter name that doesn't exist, and it might not be obvious why they are getting that error if they didn't specifically add the annotation. I forget if the same thing is required for paths or not. I think we might automatically annotate this as part of the framework, but I can't remember now. If not, we might want to avoid it for similar reasons.

jasonmreding, line 1: Should there be a "#MeasurementToDo" comment on the front panel of the `Measurement Configuration.ctl` and `Measurement Results.ctl` controls. The comment could explain that all supported data type are being shown in the default template and that users should add/remove controls to fit the needs of their measurement.

Similarly, for the `Measurement UI.vi`, we might want to update the todo comment to explain that the template is showing use of controls/indicators for all supported data types. Users should add/remove controls as needed for their desired measurement UI. Also, the todo comment could use a little cleaning up as it references concepts like gplugin for which the user has no knowledge of and really aren't public.

jasonmreding (approved):

Source/Generator/Measurement Service Template.lvproj
dixonjoel: Yeah because of the reasons you mentioned, I'd rather not include those types. We do have other examples, of course, at least for Pins.

PR #493, opened 2024-03-08
Add MeasurementLink top-level package and an examples package.

sreekantham:

Adds a MeasurementLink top-level package.  Adds an examples package.  Cleans up license info and icons in all packages.

Updated minimum LabVIEW version to 2021 per internal decision document (https://dev.azure.com/ni/Decision%20Registers/_workitems/edit/2721249)

**Please note: The new packages only work when built with LV 2021 from VIPM.  It doesn't work when using LV 2020 because our drivers don't install LV 2020 support, thus the examples are broken.  Further it doesn't work with LV 2023 (and perhaps 2022 as well) because the VI's moved to their version agnostic location.  We would need to resave our examples in the given version when the time comes.**  CC @cfuentes-ni 

Easy single package to install MeasurementLink.

Manually tested with LV 2021 and 2023.


Source/Build Specs/MeasurementLink Service.vipb
```diff
@@ -31,11 +31,11 @@
     </Custom_Action_VIs>
     <VI_Package_Configuration_File>MeasurementLink Service.vipc</VI_Package_Configuration_File>
     <Description>
-      <One_Line_Description_Summary/>
-      <Description>This is the MeasurementLink Service runtime support package.</Description>
+      <One_Line_Description_Summary>Contains the VIs needed to development measurements</One_Line_Description_Summary>
```
jasonmreding, line 34: Some grammar issues here. Maybe something like the following instead?

<One_Line_Description_Summary>Provides runtime libraries needed to edit and run MeasurementLink measurements.</One_Line_Description_Summary>

\<Description>This package provides libraries and APIs that simplify implementing and running MeasurementLink measurements. The APIs can be found under the "Measurement I/O -> MeasurementLink" category on the functions palette.\</Description>

Source/Build Specs/MeasurementLink Generator.vipb
```diff
@@ -30,11 +30,11 @@
     </Custom_Action_VIs>
     <VI_Package_Configuration_File>MeasurementLink Generator.vipc</VI_Package_Configuration_File>
     <Description>
-      <One_Line_Description_Summary/>
-      <Description>This is the MeasurementLink Plug-in Generator package.</Description>
+      <One_Line_Description_Summary>Contains a tool which generates a plugin template</One_Line_Description_Summary>
+      <Description>Contains a tool which generates a measurementlink plugin template.  Accessable from the LabVIEW tools menu.</Description>
```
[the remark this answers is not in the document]
jasonmreding: Did I mention I really dislike the MeasurementLink branded VI icon that is shown for the package. 😃

Source/Build Specs/MeasurementLink Examples.vipb
jasonmreding, line 1: image

Contains examples of MeasurementLink measurements.

image

This package contains examples that illustrate various aspects of MeasurementLink measurements. Some examples utilize various measurement hardware which requires separate installation of the device driver in order to run the example. The examples are installed to the "examples\MeasurementLink" folder underneath the LabVIEW application directory.

jasonmreding, line 1: image

This appears to be generating palettes for each example, which I don't think we want. I would uncheck both palette check boxes. I think the only presence we want on disk is the examples folder.

Source/Build Specs/MeasurementLink Service.vipb
sreekantham: Updated

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: Updated

Source/Build Specs/MeasurementLink Generator.vipb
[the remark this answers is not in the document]
sreekantham: Updated summary.  Do you have an alternate suggestion for the icon @jasonmreding ?

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: Removed

jasonmreding, line 1: image

Maybe it's not worth worrying about, but I'm not sure we should be installing this README file. Each example project has its own README that's more tailored to it. However, this README isn't included in the project files, which is maybe something we should fix.

It also brings up questions as to whether we should be trimming the files under each example project. For instance, each example folder currently includes .sfp, .instudioproj, .seq, and .pinmap files which have nothing to do with LabVIEW. I don't feel that strongly about this, but I thought it was worth a mention.

dixonjoel (approved):
I think @jasonmreding has already given this one a thorough review. So approving pending his approval.

Source/Build Specs/MeasurementLink.vipb
jasonmreding, line 1: image

I don't think we want these palettes.

Source/Build Specs/MeasurementLink Generator.vipb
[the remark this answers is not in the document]
sreekantham: Resolved

Source/Build Specs/MeasurementLink.vipb
sreekantham: VIPM seems to make a lot of assumptions I wasn't aware of.  Thanks for catching this stuff.

sreekantham: Removed

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: FYI, when you use the LV example finder it only shows projects and VIs.  So it doesn't show the readme.

Source/Build Specs/MeasurementLink.vipb
jasonmreding, line 1: image

I'm not sure where these package names are defined, but can we make the examples package name be more consistent with the other package names?

jasonmreding, line 1: Is the plan to only publish this top level package or to publish all of them? I guess there probably isn't a way for the dependency packages to this package to not show up in the store?

sreekantham: image
Looks correct to me :(

sreekantham: It should come from the Display Information > Product Name

sreekantham: We need to publish all of them.  But the customer only has to install the top-level one to get all of them.  Alternatively, they can pick and choose if they want to.

sreekantham: See also:
image

jasonmreding, line 1: image

Icon is inconsistent. While I don't like the icon from the other packages, we might as well be consistent about its usage under the umbrella of MeasurementLink.

jasonmreding, line 1: While we're at it, maybe update the `ni_protobuf_types.vipb` package for consistency? Icon, description, licensing, etc. It shouldn't reuse the MeasurementLink icon, but the current icon also isn't very good.

sreekantham: strange, the icon is there for me:
image

sreekantham: That's next on my list.

Source/Build Specs/MeasurementLink Examples.vipb
jasonmreding, line 1: Posting here so we can have a threaded conversation.

> Further it doesn't work with LV 2023 (and perhaps 2022 as well) because the VI's moved to their version agnostic location.

Does LV not handle search paths for dependencies when moving from the version specific <vilib> to version agnostic location of <vilib>? Or is this just a VIPM thing where it doesn't handle that transition for dependencies seamlessly? Otherwise, we should probably come up with a better transition plan for this sort of thing.

jasonmreding: Yeah, I didn't figure it did. It's just putting more stuff on disk than needs to be there. Same goes for the TS and IS files. It's not a big deal as it's not a lot of disk space, but technically it doesn't need to be there unless we intend for this to be a one stop shop for finding examples for LV, TS, and IS. However, I have no idea how anyone would ever think to look in this folder for IS or TS examples. We don't really have a good solution for an example folder for a cross application work flow. That particular README file strikes me as more a repo readme file than an example readme file. However, it's unlikely anyone will ever notice it.

Source/Build Specs/MeasurementLink.vipb
jasonmreding: It's there for me now as well. Not sure if I accidentally crated a local change that messed it up or what? Anyway, I'm resolving this.

jasonmreding: That's weird, it still shows up like below for me after pulling the latest changes to this branch.

image

Oh well... feel free to resolve. I couldn't find anywhere where that name might be coming from either which is why I was confused. I was hoping you might know, but it sounds like there might be some gremlins in the tools.

Source/Build Specs/MeasurementLink Generator.vipb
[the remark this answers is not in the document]
jasonmreding: Not really, at least not anything I have a ready made icon I could give you. For the client side libraries, this is rough iconography for the public APIs.

image

We could reuse one of those banners, but I'm not sure what icon to put in the main part unless you like just leaving it a field of white space.

Otherwise, I would almost rather use a generic NI icon/logo like you get with the task bar icon with the UI editor. However, I don't know how easy that would be to convert into a 32x32 icon. Just browsing through the [NI packages online](https://www.vipm.io/publisher/ni/), it doesn't look like we've put a whole lot of effort into this in the past. Maybe just using the MLink banner as shown above with some basic icon/text would be sufficient? Here's a shot at some alternatives using text and basic glyphs that already exist in the default icon library used by LV. If you like any of these options better, feel free to steal them.

image

Source/Build Specs/MeasurementLink Service.vipb
jasonmreding, line 1: I meant to add this earlier and forgot, but I like the package name "MeasurementLink Runtime" better than "MeasurementLink Service" as that reads better and intuitively makes more sense to me as to what the package is upon first read. The fact that it includes code from the "Source\Runtime" folder kind of proves my point. However, it's not a deal breaker to me.

Source/Build Specs/MeasurementLink Generator.vipb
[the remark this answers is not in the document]
sreekantham: @jasonmreding I'm inclined to leave it as-is.  None of the other icons seem unique enough to gather the purpose at a glance.  Technically that is the official icon from our marketing folks.

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: While this is a cross application workflow, it is first and foremost a LV example, so I think it makes sense to leave it here.  The additional TS and IS files are good to include because they show out to integrate this LV measurement/example into those other applications.  We will still have the github repo with the examples too.
I'm going to leave those extra files as the additional disk footprint is minimal and it does make it consistent with what one can get from github.

Source/Build Specs/MeasurementLink.vipb
sreekantham: I'm going to see if someone else can reproduce this issue.

Source/Build Specs/MeasurementLink Service.vipb
sreekantham: @jasonmreding doesn't the Runtime/Service package include code needed for development too?  Like the session management APIs and such?  I know with LabVIEW that line is kinda blurred.

I'm going to keep the word Service for this PR.  We can consider it as a follow-up.  But if we make a name change, I think we should make the change consistent with the Python repo as well (which also currently uses services as I understand it)

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: When you load it in LabVIEW it does yes.  However, when that change it made it also modifies the source files accordingly and makes it "dirty".

Putting myself in VIPM's shoes, I wouldn't want to make any changes to the source files while building the package so I bet it chooses the safe thing and errors out.

Whenever we cross that LV version change where the folder's are updated we just have to resave our VIs and it will work fine.

jasonmreding: OK, does that mean the effective driver version required for each driver to build the vipb files has also been increased? I don't know if we actually have anything documented for that or not. All of the source code is still saved in LV 2020, and we've intentionally avoided requiring anything newer to date. However, I know that creates tension with the drivers and their 3 year sliding window of support. Once drivers move to the addon model, I suspect this 3 year sliding window will get relaxed since it no longer requires installer changes on their part, but I'm not certain.

Source/Build Specs/MeasurementLink Service.vipb
jasonmreding: > doesn't the Runtime/Service package include code needed for development too?

Yes, I was hoping the term runtime implied that as well. There is probably no great answer for this. The bulk of the current code that is distributed is for the measurement service, which then isn't needed once built into an exe since it is self contained at that point. However, a lot of the APIs are really client libraries. It's just that the measurement service is really a client to a number of other services. However, those APIs can also be called from outside a measurement. We also have TestStand specific libraries which don't really fall under the service umbrella. Similarly, there has been quite a bit of talk recently about how to call measurements from outside InstrumentStudio and TestStand. If we ever added libraries here to help make this easier, using the term service in the package name becomes even muddier. However, maybe that's because I'm reading it as the libraries required to implement a service and not call a service. If you frame it as libraries related to measurement services, then it still kind of works.

sreekantham:
We are considering holding off submitting this until after the Q2 release.  Since this change requires bumping the minimum LV version to 2021 we are hoping to align this change with the Q3 release.
CC @cfuentes-ni

Source/Build Specs/MeasurementLink Examples.vipb
```diff
@@ -0,0 +1,186 @@
...
+  <Library_General_Settings>
+    <Package_File_Name>NI_MeasurementLink_Examples</Package_File_Name>
+    <Library_Version>2.0.0.2</Library_Version>
+    <Auto_Increment_Version>false</Auto_Increment_Version>
+    <Library_Source_Folder>..\Example Measurements</Library_Source_Folder>
+    <Library_Output_Folder>..\..\Build Output</Library_Output_Folder>
+    <Company_Name>NI</Company_Name>
+    <Library_Summary/>
+    <Library_License>MIT</Library_License>
+    <Package_LabVIEW_Version>21.0 (64-bit)</Package_LabVIEW_Version>
```
sreekantham, line 11: We are considering holding off submitting this until after the Q2 release. Since this change requires bumping the minimum LV version to 2021 we are hoping to align this change with the Q3 release.
CC @cfuentes-ni

Source/Build Specs/MeasurementLink.vipb
aepete: A workaround, locally, would be to build & install the ni_measurement_examples .vip, then reload MeasurementLink.vipb in Pacakge Builder.

I was able to reproduce the same behavior in regard to MeasurementLink Service .vipb's dependency upon the grpc_types vip, and the described fix caused the preferred cosmetic name to appear in the service package's dependencies list.

jasonmreding (approved):

Source/Build Specs/MeasurementLink Examples.vipb
sreekantham: per internal decision document, we will support lv 2021+
https://dev.azure.com/ni/Decision%20Registers/_workitems/edit/2721249

aepete (approved):
Could you please provide brief rationale for having updated the `Package_LabVIEW_Version` to 2021 in the vipb files?

edit: No response needed....  The answer had already been provided within the updated PR description.

cfuentes-ni (approved):

PR #501, opened 2024-04-05
Add generated code and .proto for ni.protobuf.types.DoubleAnalogWaveform

ccaltagi:

To support LabVIEW services returning `DoubleAnalogWaveform`s, we need to have a canonical definition of the cluster for `ni.protobuf.types.DoubleAnalogWaveform` and its dependencies.
This PR adds the new controls to the `ni.protobuf.types` library and the corresponding proto files.

Precursor to supporting `DoubleAnalogWaveform` in LabVIEW services.

Build and installed the Common `gRPC Types.vipb` and ensured that the `DoubleAnalogWaveform` types are in the correct place in vi.lib.


jasonmreding (approved):

Source/gRPC/Generated APIs/ni/protobuf/types/ni.protobuf.types.lvlib
dixonjoel, line 1: I would expect these to be under a different folder, e.g. 'ni_protobuf_types_DoubleAnalogWaveform'. Was this the folder they were generated in?
image

Source/gRPC/Generated APIs/ni/protobuf/types/RPC Messages/ni_protobuf_types_DoubleAnalogWaveform.ctl
dixonjoel, line 1: The front panels of all of these .ctls open on the far right of my fairly wide monitor. Please re-save them so they open near the top-left of the screen.

Source/gRPC/Generated APIs/ni/protobuf/types/ni.protobuf.types.lvlib
dixonjoel, line 1: If you haven't, you should build https://github.com/ni/measurementlink-labview/blob/main/Source/Build%20Specs/ni_protobuf_types.vipb and install it with VI Package Manager to ensure it puts the right stuff in vi.lib. I think we'll want to change the version from 1.0.0.1 to 1.1.0.1 whenever we make our next release. But we can take care of that during the release.

dixonjoel (approved):
Approved with suggestions.

Source/gRPC/Generated APIs/ni/protobuf/types/ni.protobuf.types.lvlib
jasonmreding: If you didn't already, you will probably want to generate/regenerate all proto files defined in the package at the same time. Otherwise, you are likely going to have to make a lot of changes by hand after the fact.

ccaltagi: @dixonjoel, I didn't notice it at first, but the `ni_protobuf_types_DoubleXYData` folder is actually a virtual folder in the lvlib and doesn't match the folder architecture on disk (the typedefs are created in `Source/gRPC/Generated APIs/ni/protobuf/types/RPC Messages`).
Do you want me to follow that same model?
I also have a question on the naming of my files: I decided to only expose the Flat version of the typedefs (as opposed to the non flat that embeds classes) and renamed ni_protobuf_types_DoubleAnalogWaveform_Flat.ctl => ni_protobuf_types_DoubleAnalogWaveform.ctl
Would you recommend using the original name instead (ni_protobuf_types_DoubleAnalogWaveform_Flat.ctl)?

@jasonmreding, are you suggesting I regenerate the ni_protobuf_types_DoubleXYData.ctl? That is currently the only  typedef in `Source/gRPC/Generated APIs/ni/protobuf/types/RPC Messages`.
It seems we should only do that if the protofile has changed, right?

ccaltagi: @jasonmreding, thanks for the context. I wasn't aware the flat controls were generated to support `oneof` fields.
In my case, the `WaveformAttributeValue` message has a `oneof`, hence 'normal' and flat typedefs have been generated for the `WaveformAttributeValue` and containing `DoubleAnalogWaveform`.
Do I need to submit both? I thought only the flat one would work well with the gRPC API.
Also, do I need to submit the flat to rich and rich to flat conversion methods?
Thanks.

ccaltagi: I built the package and verified the files are installed at the correct location in vi.lib.

jasonmreding: As I said previously, my preference would be to submit everything that is generated by grpc-labview as is and not worry about this sort of thing. The flat -> rich and rich -> flat are used when reading request messages from grpc containing the oneof and writing response message to grpc containing the oneof (I forget what direction corresponds to which). Yes, all 4 VIs are needed for fields/messages where oneof is used. If the generated VIs contain meaningful conversion implementation, then they are required. Otherwise, they could technically be deleted if you manually delete and/or rename things. I would have to look at what the generator produces again to be certain, which I don't want to do because I don't think we should be messing with this stuff. I would rather just fix the generator so that it generates the same number of VIs it did previously for each message if it doesn't need oneof support. The same code generation would apply to `optional` fields as well if/when grpc-labview supports that keyword.

ccaltagi: @jasonmreding OK, I'll submit the generated code the tool produced as is.
One last question: the tool prompts to generate the client, server or both. The types and conversion VIs are the same for both, but the server also has setter VIs that the client doesn't have. Do I need to submit those as well?
In other words, should the RPC messages that we submit to the measurementlink-labview repo be the client or the server?
Also should I submit `Register gRPC Messages.vi`?

dixonjoel: These .protos don't have any 'service' elements, only 'messages'. So I think just the client.

ccaltagi: Committed all auto-generated types and conversion VIs.

PR #504, opened 2024-04-11
Improve DC Power Example usability

### LINKED ISSUE #486: Improve usability of DC-Power example
# Bug Report

The DC-Power example should have an optional 'Device Reset' once complete to return the instrument back to a known state. Measurements are available to be used in TestStand, and should support the workflow of multiple measurements using the same session.

## Repro or Code Sample

Example:
image

## Expected Behavior

I'd like some documentation about when and why you would perform a device reset.

[AB#2661919](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2661919)

nathan-murphy:

Improves usability by illustrating a device reset at the end of a measurement.

When our team was working with this example as a starting point, we got tripped up by needing to reconfigure the instrument and not realizing that it wasn't being handled by MeasurementLink directly. A note like this would have helped save us some development/debugging time from the start.  

Code added is exclusively in a diagram disable structure, but we have tested that the functionality works on real hardware. Original example continues to work as normal. 

Fixes #486


nathan-murphy:
Wrong LV version was used. Will update with correct version.

bkeryan:
@nathan-murphy @dixonjoel Should the Python example be updated too? https://github.com/ni/measurementlink-python/blob/main/examples/nidcpower_source_dc_voltage/measurement.py

nathan-murphy:
@bkeryan If we started with the python measurement example it's likely that we would have run into the same issue, so yes, I think the python one should also be updated. As long as there's no dissent from @dixonjoel I'll open another issue on that repo.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
dixonjoel, line 1: I'll wait for the updated LV version change.

dixonjoel (changes requested):

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
nathan-murphy: The second commit on this PR has the VI saved for LabVIEW 2020.

dixonjoel: I was going to ask the same thing. What are the implications or are there any drawbacks if we just always reset / abort it?

nathan-murphy: It's not clear if Abort will reset all properties of a channel to default, so if you call different measurements with the same channel and abort between them I'm concerned that settings will bleed over from the first measurement into the second if the user doesn't explicitly set all values. 

I assume Reset takes longer than Abort, but I don't have HW available at the moment to benchmark this.

dixonjoel (approved):
Approved with suggestions.

Source/Example Measurements/NI-DCPower Source DC Voltage/NI-DCPower Measurement/subVIs/NI-DCPower Specific Logic.vi
nathan-murphy: Point of information:
The driver example referenced in this Measurement example uses Reset. I'll stick with Reset for consistency.

PR #506, opened 2024-04-15
example: updates niscope measurement to run when trigger source isn't present in measure pins

Jayaseelan-James:

- The `Trigger Source` configuration value is reserved along with `Pin Names`.
- Renamed the input configuration name from `Pin Names` to `Measurement Pins` and `Trigger Source` to `Trigger Pin` for improved clarity.

[AB#2683394](https://dev.azure.com/ni/DevCentral/_workitems/edit/2683394)

The existing NI-SCOPE example holds a software constraint that the trigger pin must always be present within the pin names (measure pins) because only the pins in the `pin_names` are reserved and not the `trigger_source`. However, there could be use cases where the trigger pin need not be within the measure pins.

Python PR containing similar update: https://github.com/ni/measurementlink-python/pull/694

Manually verified the example execution.


jasonmreding:
"Measure Pins" reads weird to me as "measure" is a verb and not an adjective. I think "Measurement Pins" or "Acquisition Pins" makes more sense in this context. For consistency, I would kind of expect the trigger input to also have "pin" in the name if it only accepts pins and not other signal names. So maybe "Trigger Pin" and "Measurement Pins" or "Trigger Source" and "Measurement Source".

jasonmreding (approved):

bkeryan:
> "Measure Pins" reads weird to me as "measure" is a verb and not an adjective. I think "Measurement Pins" or "Acquisition Pins" makes more sense in this context. For consistency, I would kind of expect the trigger input to also have "pin" in the name if it only accepts pins and not other signal names. So maybe "Trigger Pin" and "Measurement Pins" or "Trigger Source" and "Measurement Source".

If you rename it based on this feedback, please rename it for Python too

Jayaseelan-James:
> "Measure Pins" reads weird to me as "measure" is a verb and not an adjective. I think "Measurement Pins" or "Acquisition Pins" makes more sense in this context. For consistency, I would kind of expect the trigger input to also have "pin" in the name if it only accepts pins and not other signal names. So maybe "Trigger Pin" and "Measurement Pins" or "Trigger Source" and "Measurement Source".

I've updated the parameter names to "Measurement Pins" and "Trigger Pin" respectively.

PR #511, opened 2024-04-25
Modify measurement template to have a PPL for the Measurement UI

jarnoldNI:
Changed the template to be better suited for UIs which use anything other than LabVIEW primitives. This template separates the UI VI into a different lvlib. The generator adds a PPL build spec, builds the PPL, and updates the Get UI Details.vi with the appropriate path. 

Here's what the output of the new template is:
image

Adds a template with a UI users can use with subVIs, type defs, etc. as opposed to the original template UI which only works with LabVIEW primitives.

Built into a local VI package, installed and tested creating templates with and without the new flag toggled. Ran the new template in the development environment. Also, built the exe and deployed as well. Tested that the plug-in works in InstrumentStudio.


jasonmreding (changes requested):
I haven't done a detailed review yet since viewing diffs is very difficult with the current structure of the change. Rather than making a copy of the current template library in source and adding a new library for the UI, I would prefer to just maintain the existing template as is. Then as part of the scripting when creating the service, remove the UI VI from the template project and add it to a new UI library that is created on the fly. That should eliminate having to maintain two copies of what appear to be the exact same files in source. As long as you have existing VIs to work from, I don't think the scripting to create new libraries and modifying the copy of the template within the user's project should be too bad.

Once that structure is in place, the real changes should be more apparent, and I may have additional feedback at that time. 

Another thing you could consider while doing this is simply adding a second scripting pass over what gets created from the existing template. That might allow for adding a second command into the tools menu for creating a UI template stand alone which might be nice for cases where the measurement is written in another language. Or it could be used to upgrade an existing measurement from a stand alone UI VI to an advanced UI with its own build. The logic for these commands would be identical to the logic you are adding here for creating the measurement with the advanced UI checkbox checked. They would just need to be wrapped into a UI that is specific to the needs of that command.

jarnoldNI:
Another option which would get rid of having to maintain two templates would be to just keep one and change it to be the 'advanced' one that I've added here. No more Meas UI in the same lvlib and I could get rid of the checkbox option on the UI. Just do a full switch. I added the switch in case someone really wanted to the old behavior.

jasonmreding:
> Another option which...

I would prefer that over maintaining two templates as is the case in the current state of the PR. I'm not sure how clients would feel about that though. If scripting the changes needed to support both use cases from a single template is a lot of work, we can discuss this further. We should probably run it by some of our internal clients first though to gauge their reaction.

jarnoldNI:
> > Another option which...
> 
> I would prefer that over maintaining...

The scripting work feels like a large lift and potentially error prone compared to the value of keeping the original template around. I'll survey internally then we can gauge the response and determine the next steps.

jarnoldNI:
Latest commit removes the additional template and updates the original one to the new pattern shown above which better supports UI development.

dixonjoel:
Please update the PR description since we don't have the check box anymore.

Source/Generator/Create Measurement Service.vi
jasonmreding, line 1: Posting comment here so I can get a threaded conversation.

This PR still includes duplicates of all of the files on disk. Please delete the files that are no longer being used/referenced. At this point, it might be easier to abandon and start from scratch (not sure). That should reduce the noise in this PR considerably and make it easier to see what truly changed since hopefully the files will show up as renames/edits instead of adds.

Source/Generator/MeasurementLink Measurement Template/Measurement/Advanced/Build Assets/Post-Build Action.vi
jasonmreding, line 1: This PR still includes duplicates of all of the files on disk. Please delete the files that are no longer being used/referenced. At this point, it might be easier to abandon and start from scratch (not sure). That should reduce the noise in this PR considerably and make it easier to see what truly changed since hopefully the files will show up as renames/edits instead of as adds.

Source/Generator/MeasurementLink Measurement Template/Run Service.vi
jasonmreding, line 1: I believe the file at this location is no longer needed.

Source/Generator/Create Measurement Service.vi
jasonmreding, line 1: You effectively have dead/unreachable code here. I would prefer to see the code removed as I don't think there is any value in keeping it around. Once the code is removed, any sub VIs that are no longer referenced should be deleted.

Source/Generator/_Editor/Create PPL Build Spec.vi
jasonmreding, line 1: There are a couple forks of files here. I would either retain the original file name with the updated logic (which makes diffing easier) and then revert the add of the new file or delete the old file.

Source/Generator/_Editor/Save As Measurement Service template [Adv UI].vi
Comment (jasonmreding) [inline Source/Generator/_Editor/Save As Measurement Service template [Adv UI].vi:1]: Same thing here.

Source/Generator/Create Measurement Service.vi
jarnoldNI: fixed

Source/Generator/_Editor/Save As Measurement Service template [Adv UI].vi
Comment (jarnoldNI) [reply Source/Generator/_Editor/Save As Measurement Service template [Adv UI].vi:1]: fixed....removed unused VIs

Source/Generator/_Editor/Create PPL Build Spec.vi
jarnoldNI: That was my bad. Somehow those old files stayed on disk and were included in the commit. I have deleted all of the old unused VIs. Hopefully it's easier to see the changes now that the dups are gone.

Source/Generator/MeasurementLink Measurement Template/Measurement/Measurement UI.vi
jasonmreding, line 1: I believe this VI should be deleted from this library since you copied/moved it to the UI library.

Source/Generator/_Editor/Create PPL Build Spec.vi
jasonmreding, line 1: - All inputs aside from error in should be required.
- Update the VI properties so that the compiled code isn't saved with the source code.

Source/Generator/_Editor/Update UI Details.vi
jasonmreding, line 1: I would recommend updating the label name of the array constant from "appended array" to something that is more meaningful/readable, like "user interface details" or "user interface information".

image

Source/Generator/_Editor/Create PPL Build Spec.vi
jasonmreding, line 1: Building the ppl for the UI as part of template creation seems a bit weird to me. Is there a reason for doing that? It seems like you have all the information available at hand without having to build anything. I'm not sure if that is what is contributing to what feels like longer generation times than what I am used to, but I would prefer not to do this unless there is a good reason.

jasonmreding, line 1: I personally don't like putting build files in the same directory as the source code such that files from the build or co-mingled with source files. This is a PITA if you are using source code control. However, I realize the existing build spec for the exe is doing the same thing. So while I won't require you to fix this as part of this PR, I would certainly support it if you wanted to make that change.

Source/Generator/MeasurementLink Measurement Template/Measurement/Measurement UI.vi
jarnoldNI: done

Source/Generator/MeasurementLink Measurement Template/UI/Measurement Template UI.lvlib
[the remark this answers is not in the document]
jarnoldNI: I added the libraries to the project. The libraries are already named consistently. I removed the MeasurementLink prefix previously to remove some naming redundancy. I also deleted the test which is no longer valid with the new UI structure.
image

Source/Generator/_Editor/Create PPL Build Spec.vi
jarnoldNI: done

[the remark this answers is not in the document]
jarnoldNI: Done. 
image

Source/Generator/_Editor/Update UI Details.vi
jarnoldNI: Done.
image
image

Source/Generator/_Editor/Create PPL Build Spec.vi
jarnoldNI: I added it because it removes one step for the user right after generation for running their code plus I didn't notice any longer generation times so didn't see a reason not to do it. I can remove it if you feel strongly about it.

jarnoldNI: I agree in principle. Reasons not to do that though:

- Everyone will have slightly different folder structures and no matter what we won't always get what the user wants in that regard. Personally, I always change the build destination for the exe to match my needed project folder structure. Having the PPL in the source eliminates having to manage a second build destination.
- Having the PPL be relative to the source makes it possible to fill in the UI path in UI Getails.vi and keep it 'portable' with the code. If we put it outside of the source folder and we expect that user will change their build folder then we would be forcing them to also update the Get UI Details.vi. It's just another step that I don't think adds that much value.

Source/Generator/_Editor/Update UI Details.vi
jasonmreding: While you are here, can you set "Size to Text" to true on the string constant inside the array constant? I think this generally results in a better user experience than having to hover or grow the constant to read what is in it.

Source/Generator/MeasurementLink Measurement Template/UI/Measurement Template UI.lvlib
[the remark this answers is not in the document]
jasonmreding: nit: I would rename `Measurement Template UI` to `Measurement UI Template`. The former sounds like a user interface for a template rather than a template that is a user interface.

Source/Generator/_Editor/Create PPL Build Spec.vi
jasonmreding: I don't know if I agree with all of this, but the conversation did remind me of the primary reason it is this way currently. The reason is we thought it was desirable for the measurement logic to work as is when running the service manually and debugging through the IDE and when running as a built application. For that to work, the relative path to the ppl from `Measurement Logic.vi`has to be the same when running `Run Service.vi` directly in the IDE and when launching the exe within the runtime engine. There are likely other things we could do here that might make things a little better, but I would consider them out of scope for this PR. Since you are one of the primary consumers of this portion of the work flow, I would tend to defer to your feedback as to whether maintaining this dev time convenience is preferable to any negatives it introduces around source code control.

jasonmreding: My preference would be to remove it. The default UI isn't going to link up to any measurement parameters so I don't really see any value in building it. Also, if it's common for the user to update the output directories in their build specs to better suit their project layout, then it is just another file they have to track down and clean up.

Source/Generator/_Editor/Update UI Details.vi
jarnoldNI: done

jasonmreding (approved):

Source/Generator/_Editor/Update UI Details.vi
jasonmreding: I didn't really see any difference here, but it might not be worth messing around with too much anyway. After playing around with this a bit, the size to text behavior within an array constant seems a little buggy anyway as opposed to a stand alone scalar constant. If you were going do anything further with this, I might just change the default size of the array constant  inside the template so that at least some portion of the string constant is readable.

jarnoldNI: Agree. I did resize in the template, but it does not resize after scripting in a new path so I just increased the size of the string like you said.

Source/Generator/MeasurementLink Measurement Template/UI/Measurement Template UI.lvlib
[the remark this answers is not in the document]
jarnoldNI: renamed as recommended

Source/Generator/_Editor/Create PPL Build Spec.vi
jarnoldNI: removed the building of the PPL from the generator

PR #516, opened 2024-05-15
Condition reading the MechAction property to not be read at runtime

dixonjoel:

Fixes [AB#2738336](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2738336)

When a measurement was built into an executable with the 2.2.0.1 prerelease, it would launch briefly and then the .exe would go away. The error turned out to be attempting to read the 'MechAction' property to see if the user was using any latched booleans. We already check this at edit time with Run Service.vi and at build time of the .exe. So I'm disabling this check in the run time engine.

Allows measurement services to be built into .exes, a basic requirement of the framework.

I will build the .vip files locally in preparation for a prerelease to be made soon and test the built .exe case.


Source/Runtime/MeasurementLink Measurement Server/Helpers/Is Latched Boolean.vi
jasonmreding, line 1: While it doesn't make a functional difference with the current state of the code, I think maintenance/readability would be improved if you wire the error cluster through all cases of the conditional disable structure.

image

jasonmreding (approved):

dixonjoel:
> 0 of 1 files processed successfully. Processed `Is Latched Boolean.vi`: ERROR 1 Error 1 occurred at Check Data Size.vi
> 
> Possible reason(s):
> 
> # LabVIEW: (Hex 0x1) An input parameter is invalid. For example if the input is a path, the path might contain a character not allowed by the OS such as ? or @.
> Command requires GPIB Controller to be Controller-In-Charge.
> 
> Complete call chain:      Check Data Size.vi      Write PNG File.vi      Graphical Diff.lvlib:Merge Diff Images.vi      Graphical Diff.lvlib:Process Diff Images.vi      Graphical Diff.lvlib:run_diff.vi      RunVI.lvclass:RunOperation.vi:5910001      CoreOperation.lvclass:CallRunOperation.vi:4230002      ExecuteRunOperation.vi:5250001      ExecuteOperation.vi:6900002      ExecuteOperation.vi.ProxyCaller Diff images generated for `Source/Runtime/MeasurementLink Measurement Server/Helpers/Is Latched Boolean.vi`. Attribute Changes      None

@aepete any idea why this is not working? It's not a big deal for me in this case, but maybe you could take a look.

PR #522, opened 2024-05-27
Fix tests causing Labview hang - Add Cluster Id support for tests

MounikaBattu17:

Fixes the tests that causes LabVIEW hang by implementing cluster Id support for Starting Discovery Service in tests.

- Removed the 6 issues tests from skip tests list.
- Updated new cluster Id arg whenever there is a Kill Discovery Service in tests such that we can ensure that current Discovery service instance is killed and new service will be launched in a unique port.
- Destroyed grpc clients when killing Discovery or Session Management Services.
- Propagated cluster id to required VIs.
- Changed signature of Create Client and Recreate Client VIs for all services. 
- These [tests ](https://github.com/ni/measurementlink-labview/tree/main/Source/Tests/Tests.Runtime/Discovery%20Client)expect Discovery service to be running prior starting the tests. Thus started and killed the Discovery service at start and end of `run_tests.vi`.

Note: cluster Id is provided as **optional** terminal in connector pane for all VIs.

.PR checks.


Source/Runtime/MeasurementLink Discovery V1/Create Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Runtime/MeasurementLink Discovery V1/Ensure Discovery Service Started.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Runtime/MeasurementLink Discovery V1/Get Discovery Service Address.vi
MounikaBattu17, line 1: **Diff images:**
image

Source/Runtime/MeasurementLink Discovery V1/Recreate Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Runtime/MeasurementLink Discovery V1/Start Discovery Service.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Runtime/MeasurementLink Discovery V1/TypeDefs/gRPC Client Cache FGV Action.ctl
MounikaBattu17, line 1: Add new these values:
1) Update ClusterId
2) Destroy Discovery Client
3) Destroy Session Management Client

Source/Runtime/MeasurementLink Session Management V1/Session Reservation/Create Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Runtime/MeasurementLink Session Management V1/Session Reservation/Session Client FGV.vi
MounikaBattu17, line 1: No changes. But loaded the `Create Client.vi` and `Recreate Client.vi` references in this file.

Source/Tests/Tests.Runtime/Measurement Server/Discovery/Fake Client Cache FGV.vi
MounikaBattu17, line 1: No changes. But loaded the `Fake Create Client.vi` and `Fake Recreate Client.vi` references in this file.

Source/Tests/Tests.Runtime/Measurement Server/Discovery/Fake Create Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Tests/Tests.Runtime/Measurement Server/Discovery/Fake Recreate Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Tests/Tests.Runtime/Measurement Server/Discovery/test_DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi
MounikaBattu17, line 1: **Diff images:**
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Create Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Discovery Service.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Management Service.vi
MounikaBattu17, line 1: **Diff images:**
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Pin Map Client FGV.vi
MounikaBattu17, line 1: No changes. But loaded the `Create Client.vi` and `Recreate Client.vi` references in this file.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Recreate Client.vi
MounikaBattu17, line 1: **Diff images:**
image
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Start Discovery Service.vi
MounikaBattu17, line 1: This file is a copy of `Start Discovery Service.vi` from `Measurement Server.lvproj`.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Wait For Key File.vi
MounikaBattu17, line 1: This file is a copy of `Start Discovery Service.vi` from `Measurement Server.lvproj`.

Source/Tests/run_tests.vi
MounikaBattu17, line 1: image
image

Source/Runtime/MeasurementLink Discovery V1/gRPC Client Cache FGV Impl.vi
MounikaBattu17, line 1: **Update clusterId case:**
image

**Destroy Discovery Client:**
image

**Destroy Session Management Client:**
image

Source/Runtime/MeasurementLink Discovery V1/TypeDefs/gRPC Client Cache VI References.ctl
MounikaBattu17, line 1: Updates the vi references controls.

Source/Runtime/MeasurementLink Session Management V1/Session Reservation/Create Client.vi
jasonmreding: Why is there a new control added to the connector pane whose value is never used? I don't think we need to modify the service specific Create Client APIs outside of the one for the discovery service.

Source/Runtime/MeasurementLink Session Management V1/Session Reservation/Recreate Client.vi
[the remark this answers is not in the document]
jasonmreding: Same thing, I think you can revert this change.

Source/Runtime/MeasurementLink Discovery V1/TypeDefs/gRPC Client Cache FGV Action.ctl
jasonmreding: This should not have service specific enum entries.

Source/Runtime/MeasurementLink Session Management V1/Session Reservation/Create Client.vi
jasonmreding: Never mind. I guess this is needed to fulfill the connector pane contract? Might be worth a comment.

Source/Runtime/MeasurementLink Discovery V1/gRPC Client Cache FGV Impl.vi
jasonmreding: Also, I don't think we should ever be resetting the `unique id` back to zero as that technically creates race conditions (although very unlikely). Instead, any time we change the `client id`, I think we should stick with the pattern of always incrementing the `unique id` by one.

jasonmreding: If you wanted, I think you could collapse the "Update ClusterId" and "Destroy Client" into a single action case, but I don't feel that strongly about it.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Management Service.vi
jasonmreding: Is there any reason you discovered why killing and restarting the session management service doesn't seem to be problematic while the discovery service is? This would seem to indicate the problem/race condition was with the key file for the discovery service. Was the call to destroy the client even necessary here for things to run reliably or did you just do it for consistency with the discovery service?

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Start Discovery Service.vi
jasonmreding: I assume you copied this file and `Wait For Key File.vi` because you didn't want to make them public? I would prefer to make them community scope and add new friend libraries instead. If that means we have to wrap some of the existing tests or test runners in a library, I think that is preferable.

Source/Tests/run_tests.vi
jasonmreding: My preference would be to insert a call to `Ensure Discovery Service Started.vi` to each test which has this requirement rather than making the test runner responsible for this. That allows the VI to be runnable stand alone outside the runner which to my knowledge is true of all of our other tests.

I believe this is only an issue for the 3 tests which are referenced from the `Discovery Client.lvproj`. If that's the case, I would also be OK with:
- Making these manual tests which are only run when regenerating the client stubs. A lot of the validation seems to be related to server side error codes which isn't that useful in this context. That sort of testing is better left as unit/integration tests with the server code itself. Otherwise, these tests mostly serve as sanity checks that the generated client stubs are able to perform basic communication with the server.
- Move the tests into `Measurement Server.lvproj` with all of the other tests and use the higher level APIs and test utilities to test/validate against.

Looking closer, there are similar tests in the "Pin Map Client"  and "Session Management Client" folders which have similar requirements that certain services already be started for the tests to pass. So if these tests are passing, it seems like we are getting lucky there as well. I haven't done a deep dive, but my guess is that we already have overlapping test coverage from the tests in the "Measurement Server" folder, and these tests are redundant from an automated regression stand point. Assuming this is true, I would lean towards just deleting "test_" from the test VIs and make them manual, sanity tests that can be run when regenerating the client APIs with new versions of grpc-labview.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Discovery Service.vi
jasonmreding: I would recommend renaming `cluster id` to `new cluster id` or something like that to more clearly indicate the returned id is meant to be a unique id to be used with subsequent calls and not the current cluster id for the service instance that was just killed.

Source/Tests/Tests.Runtime/Measurement Server/Discovery/test_DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi
jasonmreding: This was a bit confusing and took me a while to understand. There are currently three callers to `Kill Discovery Service.vi`:
1. This test.
2. `test_DiscoveryWrapper_StopDiscoveryService_WrapperMethodsWorkWithoutError.vi`
3. `Restart Discovery Service.vi`

However, this was the only test VI that was modified to follow this pattern. The second test above doesn't need to adopt this pattern because it is using the higher level APIs which wrap `Start Discovery Service.vi`, and it isn't calling it directly. This means the call to `Kill Discovery Service` will update the FGV and the subsequent VIs will also use the updated cluster ID from the FGV. Conversely, since this test is calling `Start Discovery Service.vi` directly, it needs to explicitly chain the new cluster id since it is not using the FGV to start the service.

So if we look at current callers of `Start Discovery Service.vi` outside the FGV code path, we'll see two callers:
1. `test_DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi`
2. `Restart Discovery Service.vi`

However, the implementation of `Restart Discovery Service.vi` does not follow this same pattern and seems like it would be susceptible to the same intermittent hang you are attempting to fix with this PR.

image

It would be nice to centralize and hide more of this code so this type of mistake is more difficult to make in the future. Perhaps the best thing to do is:
- Make `Start Discovery Service.vi` private and not community scope.
- Update existing calls from test code (not the FGV code path) to call `Discovery Client FGV.vi` instead with the `Get Client` action.
- The call to `Destroy Client.vi` from `Restart Discovery Service.vi` is now redundant with changes made to `Kill Discovery Service.vi`. Delete it.

jasonmreding (changes requested):
At a minimum, I think we need to revisit implementation of `Restart Discovery Service.vi`.

Source/Runtime/MeasurementLink Discovery V1/Ensure Discovery Service Started.vi
MounikaBattu17: Changed the access scope to "Community" because it is needed in tests.

Source/Runtime/MeasurementLink Discovery V1/Start Discovery Service.vi
MounikaBattu17: Changed its access scope "private" since it's need is removed from tests.

Source/Runtime/MeasurementLink Discovery V1/TypeDefs/gRPC Client Cache FGV Action.ctl
MounikaBattu17: Removed service specific enum entries. Now the updated entries are
1. Get Client
2. Recreate Client
3. Destroy Client
4. Update ClusterId

Source/Runtime/MeasurementLink Discovery V1/gRPC Client Cache FGV Impl.vi
MounikaBattu17: Updated the following based on the suggestions
1) Added new cases `Destroy Client` and removed service specific cases.
2) Used common `Close Client.vi` implementation instead of having separate cases for different services.
3) Not resetting unique id to zero in destroy client case.
4) > If you wanted, I think you could collapse the "Update ClusterId" and "Destroy Client" into a single action case, but I don't feel that strongly about it.

This Destroy Client case is used for other services like Session management, PinMap, grpc device. Thus, combing these cases will update the cluster id to new value whenever there is destroy call from other services which is not needed. So maintaining separate cases.

The updated case looks like;
image

Source/Tests/Tests.Runtime/Measurement Server/Discovery/test_DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi
MounikaBattu17: Thanks for pointing this out. Surely `Restart Discovery service.vi` is a miss from me. I have updated the `Restart Discovery service.vi` as per your suggestions and the updated vi looks like
image

`test_DiscoveryServiceNotRunning_LaunchDiscoveryService_DiscoveryServiceStarted.vi`
image

`Kill Discovery Service.vi`
image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill gRPC Device Service.vi
MounikaBattu17, line 1: image

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Start Discovery Service.vi
MounikaBattu17: As per your previous comment, I have removed the logic of starting Discovery service in run_tests.vi. Thus there is no need for this and `Wait For Key File.vi` file. Hence removed these files.

Source/Tests/run_tests.vi
MounikaBattu17: Thanks for the suggestion. Since these tests are very much needed for automated regression. I have done the following
- Removed the `test_` from these test names. So these tests won't be a part of automated regression.
- Moved these tests to `Integration folder` in `MeasurementServerTests.lvlib`. Since these can't be considered as unit tests. I have named them under Integration tests. Please share your thoughts on this
- Now we have made these tests as manual run tests, should we somehow let the user know that these tests won't be part of regression?

please share your thoughts on this

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Management Service.vi
MounikaBattu17: I observed that using the same grpc client though the service is killed, the hang is there. Thus, I've added case to destroy grpc clients whenever services are killed. I have not considered pin map and gRPC device services because as you mentioned they didn't have any failing tests. But now, to maintain uniformity I have destroyed grpc clients for all services when killed. Thus updated `Kill PinMap Service.vi` & `Kill gRPC Device Service.vi` also.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Pin Map Service.vi
jasonmreding, line 1: This is destroying the client for the discovery service FGV and not the pin map service.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill gRPC Device Service.vi
jasonmreding, line 1: This is destroying the client for the discovery service FGV and not the grpc device activation service.

Source/Runtime/MeasurementLink Discovery V1/MeasurementLink Discovery V1.lvlib
```diff
@@ -29,7 +29,7 @@
 	</Item>
 	<Item Name="Discovery Service Helpers" Type="Folder">
 		<Item Name="Ensure Discovery Service Started.vi" Type="VI" URL="../Ensure Discovery Service Started.vi">
-			<Property Name="NI.LibItem.Scope" Type="Int">2</Property>
+			<Property Name="NI.LibItem.Scope" Type="Int">4</Property>
```
jasonmreding, line 32: This doesn't appear to be called from outside the library. I think it should remain private. Also, `Get Discovery Service Key File Path.vi` also doesn't appear to be called from outside the library. I would make it private as well.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill gRPC Device Service.vi
jasonmreding: It doesn't look like we created a FGV for this client so I guess you can revert changes to this VI.

Source/Tests/run_tests.vi
jasonmreding: > Added Get Discovery Service Key File Path. vi and Ensure Discovery Service Started.vi in each of these tests to start Discovery Service. Changed scope of Ensure Discovery Service Started.vi to community.

This breaks the VIs since those APIs are private/community scope. You would either need to:
- Make the APIs public. I don't think we should do this.
- Create a new library in each project that the tests are added to. Then update the friends list for  `MeasurementLink Discovery V1.lvlib`.
- Just revert the addition of these VIs to the tests and rely on the user manually starting things. The front panel already indicates this requirement to the user. This is probably what I would do for now. I kind of doubt anyone will ever run these tests. You could update these [instructions](https://github.com/ni/measurementlink-labview/blob/main/docs/Regenerating%20Server%20and%20Client.md) for regenerating code from proto files if you want to promote the existence of these manual tests which can be used as sanity tests. The tests currently only rely on grpc-labview and code generated from grpc-labview which I think was the point of the tests.

If you wanted to make it clearer these tests are meant to be manual tests, you could place the files in new top level folder called "Tests.Manual" or something like that. I'm certainly not opposed to you doing something like that, but I'm also not going to require you to do the work now to move the files and to fix up all of the projects for the renames/moves.

> I believe it is worth killing the started Discovery service instance after test run in this file, hence I have not removed the kill check alone. Please share your thoughts.

I can live with this, but I think we need to be careful not to put too much responsibility directly on the test runner and coupling it to test specific behavior. It would be nice to not have to duplicate logic for killing the service. It's possible we might end up renaming the service executables soon so having to find all of the hard coded spots in code for the exe names will be a bit of a pain (especially if it is not guaranteed to cause a failure). We could just call the existing `Kill Discovery Service.vi` from the test suite. However, that again couples the runner to our test utilities, but maybe that is the least bad option.

Source/Runtime/MeasurementLink Discovery V1/MeasurementLink Discovery V1.lvlib
MounikaBattu17: Changed the access of mentioned VIs to private

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill gRPC Device Service.vi
MounikaBattu17: Reverted.

PR #565, opened 2024-07-23
Feat: Add description URL to measurement details

MounikaBattu17:

Labview measurements include description URL while registering to the discovery service.

- Added a new class member `Get Service Descriptor.vi` to `Measurement PlugIn Service.lvclass` that includes `description URL` along with `measurement details`, `service class` and `measurement annotations` obtained from `Get Measurement Details.vi`.
- This VI is exposed as public as this is a dynamic dispatch VI.
- Replaced `Get Measurement Details.vi` with this new VI in `Register to Discovery Service.vi` (to include description URL while registering to discovery service) and in `Post-Build Action.vi` (to include description URL while creating service config file).

For testing purpose, I tweaked `Get Measurement Details.vi` in NI-DCPower example to include `description URL` and verified expected output with and without description URL.
No new tests are added. Verified existing tests pass.


MounikaBattu17:
Diff:
**Register to Discovery Service.vi**
Before:
image

After:
image


**Post-Build Action.vi:**
Before:
image

After:
image

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
dixonjoel, line 1: I don't understand why we're making this new VI instead of just modifying the implementation of Get Measurement Details. Get Measurement Details.vi is already doing the same thing of opening up the get measurement details vi path and running it so it can get the indicator values. Why not just add the 'descriptionUrl' into that VI? If we want that particular one to be optional, we can handle the error if the indicator is not there.
image

dixonjoel (changes requested):
Want to see if we can get away without a new VI in the class.

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
jasonmreding: See this [conversation](https://dev.azure.com/ni/DevCentral/_git/ASW/pullRequest/718752#1716507319) from the spec.

jasonmreding, line 1: Make this a static dispatch VI instead of a dynamic dispatch VI. We don't expect derived classes to override it, and making it static avoids backward compatibility issues if we want to change the connector pane in the future for new outputs.

jasonmreding, line 1: I would prefer to keep all of the connector panes consistent at 4x2 rather than have this be the only one that is 5x3.

jasonmreding, line 1: While it is a bit more work, I think things will be conceptually cleaner if this VI uses a typedef called something like `Service Descriptor` or `Service Description`. This would replace all of the current indicators on the front panel and consolidate them into one. This would more closely map to the [ServiceDescriptor](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/discovery/v1/discovery_service.proto#L67) message on the discovery service. With this approach, any future change where we want to add something new to the descriptor won't require a change to the template for existing measurements. They will simply get the new fields with they upgrade and get the new changes for the typedef. I think this is a better place to be in long term rather than trying to figure out how to add new template VIs or modify logic of existing template VIs for measurements that have been created from an older version of our generator/template.

As part of this, I would mark the existing `Get Measurement Details.vi` as deprecated in somewhere and indicate it only exists for backward compatibility. At a minimum I would add some comment to the front panel stating this.

jasonmreding:
There are no changes to the template/generator as part of this PR. Were you planning to make those changes in a follow up PR?

jasonmreding (changes requested):
> Waiting to see a version that uses a typedef instead of a collection of loose controls.

I've created a measurement with new template changes. The `Get Measurement Details.vi` template look like this
image

Source/Runtime/Measurements/Helpers/Register to Discovery Service.vi
MounikaBattu17, line 1: Diff images:
image

Source/Runtime/Measurements/Build Assets/Post-Build Action.vi
MounikaBattu17, line 1: Diff Images:
image

Source/Runtime/Measurements/IMeasurement Service/Create Metadata Response.vi
MounikaBattu17, line 1: Diff Images:
image

Source/Runtime/Measurements/Measurement Plugin Service/Get Measurement Details.vi
MounikaBattu17, line 1: Added a deprecation comment.
image

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
MounikaBattu17: Updated to static type.

Source/Runtime/Measurements/Run Service.vi
MounikaBattu17, line 1: Diff Images:
image

Source/Generator/Measurement Plug-In Template/Measurement/Get Measurement Details.vi
MounikaBattu17, line 1: Updated this template VI to use a cluster of fields instead of individual fields.

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
MounikaBattu17: Since this VI is exposed as public, we didn't want to break backwards compatibility. Thus, introduced a new VI to replace this. I added a comment in the front panel stating this VI as deprecated, prompting the users to use the new VI.

MounikaBattu17: Thanks for the suggestion.

- I've created a typedef `Service Descriptor.ctl` which is a cluster of measurement details, description URL, service class, and measurement annotations.
- Updated `Get Measurement Details.vi` template to use `Service Descriptor.ctl` instead of individual controls.
- Modified `Get Service Descriptor.vi` will update `service descriptor` cluster from the template. If not found, it will fall back to use `Get Measurement Details.vi` and update the cluster.
- Replaced `Get Measurement Details.vi` callers with `Get Service Descriptor.vi`.
- Added a comment to `Get Measurement Details.vi` stating it as deprecated.

MounikaBattu17:
> There are no changes to the template/generator as part of this PR. Were you planning to make those changes in a follow up PR?

Initially, I thought of follow-up PR for the template. But with the new proposed changes, I've included generator changes in this PR itself.

Source/Runtime/Measurements/Build Assets/Post-Build Action.vi
jasonmreding, line 1: Consider updating `Generate Service Config JSON.vi` so that it takes `Service Descriptor.ctl` as the input rather than all of the cluster fields individually.

Source/Runtime/Measurements/Measurement Plugin Service/Service Descriptor.ctl
jasonmreding, line 1: I don't really know why we did things this way (I'm guessing it was lack of attention to detail), but the majority of the public typedefs are defined by the `NI Measurement Plugin-in SDK` library and not the plugin class library. For consistency, I would move this typedef there as well.

image

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
jasonmreding, line 1: Consider updating the `Plugin Paths` typedef by renaming `get measurement details VI path` to `get service descriptor VI path`.

Source/Runtime/Measurements/Measurement Plugin Service/Service Descriptor.ctl
jasonmreding, line 1: Consider removing reference to `Measurement Details` typedef and just including those fields as part of the `Service Descriptor` typedef. I don't feel strongly about this. I think it is conceptually cleaner, but it makes the internal implementation a little more tedious.

Source/Generator/Measurement Plug-In Template/Measurement/Get Measurement Details.vi
jasonmreding, line 1: Consider renaming this VI to `Get Service Descriptor.vi`.

jasonmreding (approved):

Source/Runtime/Measurements/Build Assets/Post-Build Action.vi
MounikaBattu17: Updated.

Source/Runtime/Measurements/Measurement Plugin Service/Service Descriptor.ctl
MounikaBattu17: Moved `Service Descriptor.ctl` to public helpers.

MounikaBattu17: `Measurement Details` is used while creating measurement in template `Get Service Descriptor.vi`. I feel to leave this typedef as it is, as we also have `Measurement Annotations` typedef.

Source/Generator/Measurement Plug-In Template/Measurement/Get Measurement Details.vi
MounikaBattu17: Renamed.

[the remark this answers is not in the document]
MounikaBattu17: I've run existing measurements, created new measurements, and verified these changes work as expected.

Source/Runtime/Measurements/Measurement Plugin Service/Get Service Descriptor.vi
MounikaBattu17: Updated.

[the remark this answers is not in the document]
MounikaBattu17: Marked `measurement plugin service in` as required terminal.

dixonjoel (approved):

PR #571, opened 2024-08-12
Measurement Client Generator | List available V2 measurement services

varshini-senthil:

[User Story 2738655](https://dev.azure.com/ni/DevCentral/_workitems/edit/2738655): [LabVIEW] As a user, I want to view the list of active V2 measurement services
As a part of Measurement Client Generator, we need to list the available measurement services for the user to choose a measurement service for client creation. Thus, 
- The access scope of the `Enumerate Services` API has been updated from community to public. Additionally, the gRPC ID of the Discovery client instance, created by the `Enumerate Services` API is now exposed, allowing for the client to be destroyed after retrieving the list of available V2 measurement services.
             - This change in scope is intended to decouple the Generator and Service packages, as maintaining a community scope API dependency across packages is complex.

- Renamed the `Measurement Service Scripting.lvproj` to `Measurement Generator Scripting.lvproj` to accommodate all generator-dependent libraries.
             - `MeasurementService Editor.lvlib`: An existing library for generating measurement services.
             - `MeasurementClient Editor.lvlib`: A new library that handles logic for creating the measurement plug-in client. In this PR, the client generator VI will only display the list of available V2 measurement services to users; other functionalities will be addressed in upcoming PRs.
             - `Generator Helpers.lvlib`: A newly introduced library to manage common dependencies between both generators.

- Modified the `Measurement Plug-In SDK Generator.vipb` to install client generator dependencies as well.

- Users can now edit client names for available V2 measurement services. Validation will identify and highlight any invalid names. However, users are permitted to proceed with client creation if the selected measurement service client names are valid

- To update the scope of `Enumerate Services` API to public, as it is a dependency for the Measurement Client Generator. 
- To list the available measurements for the user, so that they can choose the appropriate measurement for client creation.
- To edit the client names of the available measurement services, if required.

- Generated package for Measurement Plug-In SDK Service and Generator for the new changes and tested 
- Automated test for major functions

P.S. The PR includes an RFC tag to gather feedback from reviewers regarding the change in scope of the `Enumerate Services` API. If the approach is accepted, the tag will be removed, and the PR can be considered for merging.


varshini-senthil:
`Enumerate Services` API
Diff image: 
image
image

varshini-senthil:
`Measurement Service Scripting.lvproj` -> `Measurement Generator Scripting.lvproj`
Diff image:
image
image

varshini-senthil:
`Measurement Client Generator` Workflow

1. The available V2 measurement services are listed upon execution of the tool.
image

2. `Create Measurement Plug-in Client(s)` option is enabled once an available measurement service is selected.
image

3. The tool highlights duplicate and invalid client names, during validation of user input.
image

4. Client creation is permitted, as long as the selected measurement service client names are valid.
image

Source/Generator/_MeasurementClient Utility/MeasurementClient Editor.lvlib
dixonjoel, line 1: We should make all the sub VIs private unless we need to expose them for some reason.

Source/Generator/Create Measurement Client.vi
dixonjoel, line 1: Should this pass a 'True' out to stop the VI / dialog? With it as is, clicking create client doesn't close the dialog, I think.
image

Source/Generator/_MeasurementClient Utility/SubVIs/Generate Mouse Click.vi
dixonjoel, line 1: I haven't seen this technique before. Did you find a pattern for this somewhere? I think it's fine, mostly just curious how you discovered the need for this.

dixonjoel (approved):
Approve with suggestions.

Source/Generator/_Editor/Get Reserved Names.vi
jasonmreding, line 1: Most of the names here are stale. Ideally our library names are sufficiently unique that this really shouldn't be an issue. That probably means all of our library names should begin with NI at the very least in order to uniquify them. I believe that is true of all of our run time library. For the edit time libraries for the generators, we should really be loading everything in a unique context anyway to avoid these sorts of issues. I thought LV did some of that for us for things in the tools menu, but I'd need to research that a little more. Otherwise, if our libraries are getting added as dependencies of the project we are scripting into then we should be doing a better job of managing our application contexts so that doesn't happen.

All of this is to say I don't think this change should really be necessary. If it is necessary for some reason, it is probably covering up other bugs. I would actually be OK with just getting rid of all reserved checks, but I don't necessarily expect you to make that change if you are not comfortable making it.

Source/Generator/_MeasurementClient Utility/SubVIs/Generate Mouse Click.vi
bkeryan: Wow, is this calling user32.dll?
image

Yes, it is.

What uses this?
image

Is it possible to configure the control to have the desired behavior? Generating user input seems like a hack.

Cc: @jasonmreding

Source/Runtime/Clients/Discovery V1/Enumerate Services.vi
jasonmreding, line 1: This should not expose the grpc id as an output. The public APIs we have exposed around this don't require or allow any user control over the client connection. That is all managed under the hood by our client pool. Allowing the user access to this ID so they can close the connection defeats the purpose of the client pool. It also leaves the client pool holding a stale ID. In theory this should be OK since we have retry wrappers around each gRPC call, but I don't know how grpc-labview is managing these IDs under the hood. If they end up recycling the ID, then we could end up talking to the wrong thing. Anyway, unless there is a specific problem you ran into that led you to expose this, I think the changes to this VI should be reverted.

Source/Generator/_Editor/Generator Helpers.lvlib
jasonmreding, line 1: Now that this is a top level library that acts as a sibling to the service/client scripting libraries, move it to a new folder that is a sibling to those library folders. A convention that we often use is to create folder with the same name as the owning library (minus file extension). Then place the library file and all VIs owned by that library flat underneath that folder.

Source/Generator/_MeasurementClient Utility/MeasurementClient Editor.lvlib
jasonmreding, line 1: Typically, we don't have folders like "SubVI". Those are only organizational constructs in the project file. If you implement Joel's suggestion above, I would recommend just creating a folder called "Private" or "Private Helpers". It's a lot easier to manage access scope in a batch fashion by using project folders in this way.

From a disk layout perspective, all of the VIs in the library and the library itself are typically contained flat underneath a single folder. The folder is typically has the same name as the library minus the file extension.

Source/Generator/_Editor/MeasurementService Editor.lvlib
jasonmreding, line 1: Consider updating the "_Editor" folder to mirror the naming for the measurement client code.

Source/Generator/_MeasurementClient Utility/SubVIs/Generate Mouse Click.vi
jasonmreding: See this [article](https://forums.ni.com/t5/Example-Code/Edit-Type-in-Multicolumn-List-Box-During-RunTime/ta-p/3681771). This basically eliminates the delay for editing with a single click. I don't know what the typical single click behavior is relative to highlighting and what not. I played around with it a little bit, but it wasn't clear to me if there is an expectation on standard UX behavior for this type of control. There is also an event for double click where you can insert similar logic if you want to require a double click for editing instead. You may also want to consider disabling double click editing behavior for the first column if that is read only data. Either that or you could always direct editing to the second column. I don't know if there is an easy way to disable editing of certain columns of data all together. I imagine you would have to intercept both mouse down and double click events to completely circumvent that.

Source/Generator/_MeasurementClient Utility/SubVIs/Validate Selected Client Names.vi
jasonmreding, line 1: Consider creating a helper VI around this logic so you have a single centralized point for controlling which font colors are used for error cases.

image

Source/Tests/Tests.Generator/Dialog/test_DuplicateClientNames_CheckClientNameIsNotRepeated_NameInvalid.vi
jasonmreding, line 1: Consider adding all of the test VIs for scripting to a library and then adding that library to the `Measurement Generator Scripting.lvproj`. In other words, follow the pattern used for `Measurement Server.lvproj` so that it is easy to find tests for the code that you are modifying without having to go spelunking on disk.

jasonmreding (changes requested):
From an RFC perspective, I don't see anything that warrants any huge change in direction. However, I would like to review the requested changes again before approving.

Source/Runtime/Clients/Discovery V1/Enumerate Services.vi
jasonmreding: I have been looking into this more closely, and this is a bit of a mess. The gRPC client/channel pool concept was created mostly with the server side in mind where the lifetime of the exe and the grpc-labview runtime library are the same. In that case, the fact that we never close the client session unless its usage generates an error doesn't really matter since there is no effective leak. However, when manually running the server from the IDE, this isn't the case since the lifetime of the server is shorter than the lifetime of the app. The same will almost certainly be true for client side code which you will be generating.

Furthermore, I discovered today that grpc-labview actually registers cleanup procs with all of its APIs in the C++ code. So while the U64 `ClientId` doesn't look like a refnum from a LV perspective, it is effectively behaving like one. In other words, even if you fail to call `Close Client.vi`, the client session will still get cleaned up when the VI hierarchy in which the session was created goes idle. 

However, when debugging into that implementation, it appears there is a bug where we are leaking a shared pointer (at least I think this is a bug) so you will get a very slow growing memory leak. This is actually a problem for our channel pool implementation because the measurement logic for the plugin is called by reference. That means, it effectively forms its own VI hierarchy which immediately goes idle upon completion of each `Measure` call which then triggers any cleanup procs which were registered as part of execution of that VI hierarchy. This includes clean up procs which were registered as part of our channel pool implementation which completely defeats the purpose of the channel pool. However, because of the bug referenced earlier, we never actually clean up the connections so our channel pool still kind of functions as desired even though it is slowly leaking pointers/connections.

I'm not really sure what to do about this at the moment. My immediate thoughts are either:
- Don't use the channel pool for the discovery service in your code. You are immediately closing the client session after using it so you aren't getting any of the benefits of the channel pool anyway. So just use the lower level APIs to get the discovery service address and then create a client session using the grpc-labview generated APIs. 
- Create a new public API that wraps the channel pool for the discovery service which clears out the underlying client session similar to what `Kill Discovery Service.vi` does in our test suite. At least that way, the channel pool isn't still holding onto a stale reference that is no longer valid.

I'm not sure what this means for the measurement client library itself that you eventually plan to generate. Opening/closing a client session with each measurement call is pretty wasteful. However, I need to revisit the caching strategy employed by the current channel pool in light of this new information before I can feel good about using it more broadly than we already are.

Source/Generator/_MeasurementClient Utility/MeasurementClient Editor.lvlib
varshini-senthil: The access scope of the Helper VIs which have automated tests is set to community and for other helper VIs it is set to private. So, the access scope is set for individual VIs, instead of managing the scope using folder scope in a batch fashion.
image

Source/Runtime/Clients/Discovery V1/Destroy Client.vi
jasonmreding, line 1: All VIs in the repo should should be marked as source only ("Separate compiled code from source file"). If you haven't already, I would recommend you update your LV settings so this is the default configuration when creating new VIs. The configuration of the library should have done this by default so I'm not sure how you managed to get into this state.

jasonmreding, line 1: Consider not chaining `error in` to the underlying call and instead merging the two errors clusters after the call.

Source/Generator/_MeasurementClient Utility/Measurement Selection Icons.vi
jasonmreding, line 1: Mark as source only. Also, update the owning library properties so that by default all items created under it are source only.

Source/Generator/_Generator Helpers/Generator Helpers.lvlib
jasonmreding, line 1: Mark the library as source only.

Source/Generator/_MeasurementClient Utility/Validate User Input.vi
jasonmreding, line 1: Minor: Presumably you don't need to reset the display names of the measurements if you don't allow editing them to begin with. This is harmless but perhaps a bit confusing. Assuming this is correct, I think you could just delete this VI and call `Validate Client Names` directly.

jasonmreding (approved):

PR #575, opened 2024-08-19
Generate Client APIs for V2 Measurement Service

varshini-senthil:

As part of the Measurement Plug-In Client generator, we need to retrieve the metadata of the available v2 measurement services and also invoke the measure logic in the generated client. This requires the client APIs of the measurement service for communicating with the service. Thus, in this PR,
- Generated client APIs library `Measurement Service Client.lvlib` for the v2 measurement service.
- Updated the `Measurement Plug-In SDK Service.vipb` to install measurement service client APIs library as part of the package.
- Updated the `Regenerating Server and Client.md` file to include instructions for regenerating the V2 measurement client.

- To add the generated client APIs library `Measurement Service Client.lvlib` of v2 measurement service, as it is a dependency for the Measurement Plug-In Client generator. 
- To make the library available as part of the `Measurement Plug-In SDK Service` package.
- To update the `Regenerating Server and Client.md` file with the additional instructions.

None


Source/Runtime/Measurement Server.lvproj
jasonmreding, line 1: Adding comment here primarily to get a threaded conversation.

If you haven't already, please see these instructions for how to [generate the V2 server code](https://github.com/ni/measurement-plugin-labview/blob/main/docs/Regenerating%20Server%20and%20Client.md#measurement-service-base-v2).

I don't think this is the right approach. You have added these files as a new "client" library. However, this isn't really a client library in the same sense as our other client libraries. A client library is a higher level wrapper around what you get with the default generated code. This is just the generated code as is. See [here](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/119336/gRPC-Packaging-Guidelines#) for more information on packaging guidelines that were recently posted on our internal wiki.

What you have actually added here is closer to what we have been naming "api" libraries or packages. We have been placing these generated API packages underneath the "Source\gRPC\Generated APIs\" folder in the repo and under a matching directory structure underneath <vilib> when installed as package through VIPM. 

The rub here is that grpc-labview code generation isn't great IMO, and this is the first package where we have needed to generate both client and server code for distribution. If you try to generate both client and server code without using the default library suffixes assigned by the tool then code generation will fail. I have a slide set I can share with you if you like about what I think should be changed with the grpc-labview  code generation, but none of that can be done without breaking backward compatibility and/or considerable implementation effort. That generally leaves two choices:
1. Generate both client and server using the default library suffix. Then manually rename things afterwards to preserve compatibility with the library names we've already shipped for the server code.
2. Fix the grpc-labview code gen bug when generating both client and server code with a user defined suffix. Maybe make some other fixes like single sourcing the generated RPC messages between the client and server libraries. Then convince ourselves the server side changes are sufficiently encapsulated behind our SDK APIs that this doesn't break existing measurements.
3. Generate the client and server code independently from each other and ship them in separate libraries. This is unfortunate as the generated client library won't match any of the naming/patterns that all of the other shipping generated client libraries have.

Of these options, the third option is the least work and risk. That just means coming up with a library name that we can live with relative to the existing libraries. Existing libraries are:
- ni.measurementlink.discovery.v1.api.lvlib (only client code)
- ni.measurementlink.pinmap.v1.api.lvlib (only client code)
- ni.measurementlink.sessionmanagement.v1.api.lvlib (only client code)
- ni.measurementlink.measurement.v1.api.lvlib (only server code)
- ni.measurementlink.measurement.v2.api.lvlib (only server code)

Some of the options that most readily come to mind are:
- ni.measurementlink.measurement.v2.api.client.lvlib
- ni.measurementlink.measurement.v2.client.lvlib

If we ever create a generic measurement client library on top of this, then it would be named `NI Measurement V2 Client.lvlib` using the naming style established by our current client libraries. I forget now what you were planning to name the code generated client libraries that are service class specific. If I had to pick a name right now, I would probably go with `ni.measurementlink.measurement.v2.client.lvlib`, but I'm open to other arguments.

jasonmreding (changes requested):
Waiting for outstanding conversations to be resolved.

Source/Runtime/Measurement Server.lvproj
varshini-senthil: Thank you for pointing this out @jasonmreding, I have reverted the changes to `Measurement Server.lvproj` and generated client APIs for the V2 Measurement service. The generated APIs are added as part of a new folder `measurement client`, placed parallel to the existing client and server API folders.

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/ni.measurementlink.measurement.v2.client.lvproj
jasonmreding, line 1: If you have 2024 Q3 installed, please explicitly set the "Save version" in the project properties to "Editor version" as this setting tends to work best with the .lvversion file for the repo.

image

jasonmreding, line 1: For whatever reason, all of the Flat<->Rich generated VIs aren't configured to be source only. I recommend using the properties page to batch set all items in the project as source only before checking them in. I suspect this is a bug in the code generator since it only seems to happen with these VIs and that was a relatively recent feature addition to the generator.

Similarly, none of the icons for the Rich<->Flat VIs have the icon from the prototype/containing library. You should be able to force that from library properties page as well if you like. Looking through the other generated client libraries, it appears they have the same issue with the icons.

jasonmreding, line 1: So I'm going to change my mind slightly on my previous recommendation (sorry about that). Rather than have a parallel directory hierarchy `Source\gRPC\Generated APIs\ni\measurementlink\measurement client\v2`, I would prefer to keep the generated code for a proto package under the same folder. To do that, here is what I am proposing:
1. Use the existing project under `Source\gRPC\Generated APIs\ni\measurementlink\measurement\v2` for the server code to also generate the client code. You can add the prototype library to this project.
1. Continue generating the server code in the project according to the existing instructions.
1. Then generate the client code into the same project as a second pass.
   1. Name the library "Measurement Service"
   1. Name the suffix " Client"
   1. Only generate client code
1. After generation completes:
   1. Drag the generated client library under the `ni.measurementlink.measurement.v2.api.lvlib`.
   2. Remove the empty "Measurement Service Client" folder from the project.

The end project should look something like:

image

image

varshini-senthil: I observed that the Flat<->Rich and Rich<->Flat VIs of the other generated client libraries had the same icons set by the grpc-labview code generator. So, I updated the VI icon for `RPC Service` VIs and left the icon of these VIs unchanged.

varshini-senthil: I have updated the icon for the Flat<->Rich and Rich<->Flat VIs by applying the icon from the library properties.

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/gRPC Client Test.vi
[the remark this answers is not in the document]
varshini-senthil: I have updated the Service vipb to exclude the generated test VI for Measurement Service Client, but I see that the existing client libraries contain the test VI in the vi.lib.

image

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/ni.measurementlink.measurement.v2.client.lvproj
varshini-senthil: I am using LabVIEW 21 version, so this will not be applicable

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/gRPC Client Test.vi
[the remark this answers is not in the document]
varshini-senthil: I excluded the test VI and built the package; the build was successful. But, when the `ni.measurementlink.measurement.v2.api.lvlib` in the vi.lib location is opened after package installation, the `Measurement Service Client.lvlib` expects the `gRPC Client Test.vi`.
- If this is the case, we have to remove the test VI from the client library and then exclude from package. 
- Else, we can remove the VI from library and disk as well, since it is not used else where.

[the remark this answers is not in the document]
jasonmreding: > But, when the ni.measurementlink.measurement.v2.api.lvlib in the vi.lib location is opened after package installation, the Measurement Service Client.lvlib expects the gRPC Client Test.vi

I'm not sure I completely follow this. If the other VIs contained in the library are broken because of this, then I would say just forget about this and leave things as the generator produces them. If the only thing wrong is that you see a missing VI when viewing the library in the project pane, then I'm not that concerned about this.

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/ni.measurementlink.measurement.v2.client.lvproj
jasonmreding: You should install 24Q3. The feature that allows you to maintain the source code in a specific version of LV is really nice and minimizes the need to install multiple versions of LV on your machine. Anyway, since we ended up reusing an existing project rather than creating a new one, this comment no longer applies.

jasonmreding (approved):

Source/gRPC/Generated APIs/ni/measurementlink/measurement client/v2/gRPC Client Test.vi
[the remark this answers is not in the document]
varshini-senthil: The other VIs in the library are not broken because of the exclusion of generated test VI from vipb, but I am concerned with the warning shown, when the user opens the client library in `vi.lib` location, after `Measurement Plug-In SDK Service` package installation. So, I consider having the generated test VI in the vipb. I would like to know, if there is any particular reason to not include the VI.

image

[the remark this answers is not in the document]
jasonmreding: The reason I said to exclude it is because it is a test VI and doesn't make sense to deploy on the users' systems. With that said, there is a trade off between maintainability and remembering to have to do this each time we regenerate code and correctness. I don't think most users will ever go snooping around vilib on disk or load libraries from vilib in the IDE directly. Because of this, I can live with either excluding from the .vipb and living with the broken library item or deploying the test VI even though it will never be used (the file size isn't huge). I don't have a huge preference between those two options so I'll leave it up to you. I would rather not create more manual steps that have to be repeated each time we regenerate code.

PR #576, opened 2024-08-23
Add Pin Map Client to the Measurement Server project

MounikaBattu17:

As a part of Measurement Client work, we are adding Pin Map Client to support Pin Map related APIs.
Implements [User Story 2738656](https://dev.azure.com/ni/DevCentral/_workitems/edit/2738656): As a user, I want to have an option to register the pin map and support pin-based measurements

- Added new library `NI Pin Map V1 Client`
- Added below wrappers VIs to the library                                                    
    __________________
   - Create Client.vi                      
   - ReCreate Client.vi   
   - Destroy Client.vi                      
   - Pin Map Client FGV.vi        
   - Update Pin Map from XML.vi
   These files are moved from tests to this library (excluding Destroy Client.vi) and the access scope is set to `community`.
   _______________________
   - Register Pin Map.vi - This VI will register the given pin map with the pin map service. Created Measurement Client will use this VI, so we are planning to expose this VI in the functions palette. Hence the access scope is `public`.
  
- Added a helper VI `Read Pin Map File Contents. vi ` which will return the pin map file contents as a string.
- Added automated tests

Added automated tests


Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Pin Map Service.vi
MounikaBattu17, line 1: **Diff images:**
diagram-A-before-0
diagram-B-after-0

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Register Pin Map.vi
MounikaBattu17, line 1: **Diff images:**
diagram-A-before-0
diagram-B-after-0

Source/Runtime/Clients/Pin Map V1/Register Pin Map.vi
MounikaBattu17, line 1: Currently, this VI is placed directly under NI Pin Map V1 folder. But this placement might change when exposing this VI in the palette in upcoming PRs.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Pin Map Service.vi
jasonmreding, line 1: `Kill Session Management Service.vi` is still using the old service name as well. Can you please fix that while you are here?

Source/Tests/Tests.Runtime/Measurement Server/Pin Map/test_UpdatePinMapfromXML_StopPinMapService_UpdatePinMapfromXMLWorksWithoutError.vi
jasonmreding, line 1: I think a better name would be `test_ExistingClientConnection_StopPinMapService_UpdatePinMapFromXMLWorksWithoutError.vi`.

Source/Tests/Tests.Runtime/Measurement Server/Pin Map/test_RegisterPinMap_RegisterPinMapFile_PinMapRegistered.vi
jasonmreding, line 1: You should either:
- Use the FGV to get the existing client ID and then not call Destroy Client on the FGV (that's the whole point of the pool is to reuse the connection).
- Use the grpc-labview VI for closing the client session. You can either use the VI from the palette or the one that is code generated with the pin_map_service.proto file.

As is, you are leaking the client ID that you are explicitly creating and closing the connection from the channel pool instead.

image

Source/Runtime/Clients/Pin Map V1/Register Pin Map.vi
jasonmreding: I don't know that you should need to change the file location because you add it to the palette, but we can tackle that issue in a future PR.

jasonmreding, line 1: This should not be calling `Destroy Client` as part of the implementation. That defeats the purpose of the client pool.

image

jasonmreding, line 1: The `is registration successful?` output and logic doesn't seem necessary. If the registration is unsuccessful, that should be indicated by `error out`. As far as I can tell, the service will never return an empty string for the `pin_map_id` without also returning an error.

jasonmreding, line 1: If you plan to expose this publicly to users, then consider adding a VI description as well for context help purposes.

jasonmreding (changes requested):
Since this introduces public APIs, waiting to see the next round of changes. You will eventually need to update the vipb as well, but I'm OK if you are planning to do that as part of a future PR.

Source/Runtime/Clients/Pin Map V1/Register Pin Map.vi
MounikaBattu17: @jasonmreding With reference to [this](https://github.com/ni/measurement-plugin-labview/pull/571#discussion_r1716215002) comment in another PR for Discovery Destroy Client, I understand there is a bug in channel pool and there will a slow growing memory leak if we don't explicitly close the client. 

> Create a new public API that wraps the channel pool for the discovery service which clears out the underlying client session similar to what Kill Discovery Service.vi does in our test suite. At least that way, the channel pool isn't still holding onto a stale reference that is no longer valid.

So, In similar fashion for Pin Map client, I've created a public API that wraps channel the pool of pin map service to destroy the client. I've used this Destroy Client wherever we use Pin Map FGV to make sure we don't leave out any reference leaks from this VI as this a public API.

Please share your thoughts on this

MounikaBattu17:
> You will eventually need to update the vipb as well

Could you please clarify why updating VIPB is necessary for these changes? Is it because the `Register PinMap.vi` going to get added to the palette?

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Pin Map Service.vi
MounikaBattu17: Updated. The Diff images are
diagram-A-before-0
diagram-B-after-0

Source/Runtime/Clients/Pin Map V1/Register Pin Map.vi
MounikaBattu17: I've removed `is registration successful?` as `error out` is sufficient to tell whether registration is successful or not.

MounikaBattu17: Added.
image

jasonmreding: > I understand there is a bug in channel pool...

The bug is actually in grpc-labview. However, fixing the bug actually makes the use of the channel pool from within a measurement plugin useless. In other words, the channel pool only has efficacy across RPCs because of the bug.

In this case, the usage is coming from the client side and not the server side. That means the usage model is a bit different. In particular, I would expect typical client usage:
- Would not invoke VIs using call by reference which would create client connections. grpc-labview client connections that are created from a call by ref hierarchy will get cleaned up by LV when that call by ref hierarchy goes idle if the connection hasn't already been explicitly closed. However, because of the bug, the connection isn't actually cleaned up so there is a bit of a memory leak that can accrue. If you are using the channel pool, then the leak doesn't occur until the FGV for the channel pool goes idle. This is because the FGV will continue to use its existing connection reference, and because of the cleanup bug, the connection is actually still valid. Otherwise, the retry logic in the client wrappers would receive an error and immediately recreate a new connection each time. For measurement plugins, the FGV doesn't go idle until you stop the service, which is never for a deployed exe.
- Would involve VI hierarchies that start/stop frequently from interactive use. As opposed to a built exe for a measurement plugin which never stops once started.

So you have a few choices that I can see:
1. Never use the channel pool. Instead, users will have to manage the client ID if they want to reuse it across multiple operations. Alternatively, the API you offer could always create/close the connection each time. This might be reasonable for pin maps since I imagine the frequency of registration would be relatively low. 
2. Use the channel pool and live with the leaks with the expectation that the impact of the leaks will be minor until we can fix the bug and shore up functionality.
3. Use the channel pool and also expose the `Destroy Client` API publicly. Users would then be in charge of calling this API at the appropriate time if they are concerned about the memory leak. Basically the user would need to call this VI at the end of the scope in which they want to apply channel pooling.

Your current implementation makes use of the channel pool, but then immediately invalidates the channel pool which seems kind of pointless. I think it technically also creates race conditions where there could be other gRPC calls in flight on the channel, and then you end up tearing down the channel mid flight without the user really being aware that you are doing that. Also, the current implementation is also using the channel pool for the discovery client, and it isn't cleaning that up so it is being inconsistent. 

I was OK with explicitly calling `Destroy Client` for the generator use case because VIs used for tools menu plugins are already loaded into their own "sand boxed" context so you are completely isolated from other user code. That is not the case here. This code will go directly into the user's projects.

The other longer term option would be to just build channel pooling into grpc-labview at the C++ layer and get rid of all of this LV code which is attempting to do the same thing. Otherwise, the only real option I can think of as a fix is to augment the grpc-labview API to allow the caller to specify whether or not the connection should automatically be cleaned up. That would still leak a reference once per time the FGV goes idle if the caller didn't explicitly clean up the FGV before going idle.

MounikaBattu17: Thanks for the suggestion. Based on your feedback, I've removed `Destroy Client.vi` from Register Pin Map.vi and tests as it neglects the point of using channel pool. I went with third option and made the following changes:
1) Removed `Destroy Client.vi` from `Register Pin Map.vi`, not accounting for minor memory leaks.
2) Made `Destroy Client.vi` public so that users can use it to destroy clients whenever they intend to.

Please let me know if any additions are needed.

Source/Tests/Tests.Runtime/Measurement Server/Pin Map/test_RegisterPinMap_RegisterPinMapFile_PinMapRegistered.vi
MounikaBattu17: Used FGV to get the existing client ID and removed Destroy Client from the implementation.
image

jasonmreding (approved):

Source/Runtime/Clients/Pin Map V1/Recreate Client.vi
Prem-MysteryPro, line 1: @MounikaBattu17 
Can we resize the front panel window to view all the controls completely? (ctrl + space, ctrl + f)
image

Source/Runtime/Clients/Pin Map V1/Read Pin Map File Contents.vi
Prem-MysteryPro, line 1: @MounikaBattu17 
Are there any ways to identify if the BOM character is present and remove it? Because now we are always removing the first 3 characters.
image

Source/Runtime/Clients/Pin Map V1/Pin Map Client FGV.vi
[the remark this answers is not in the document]
jasonmreding: FGV stands for functional global variable. This design pattern only works if the VI is non-reentrant. It effectively provides a mutex around the execution of the VI/method which is needed for singleton like semantics.

Source/Runtime/Clients/Pin Map V1/Recreate Client.vi
jasonmreding: If anything, use Ctrl + U to do a front panel cleanup similar to block diagram cleanup. Unlike block diagram cleanup, this pretty much works as desired 100% of the time and is better than trying to arrange things manually. The only exception would be for user interface VIs where users are expected to interact with the front panel and then run manually.

Source/Runtime/Clients/Pin Map V1/Read Pin Map File Contents.vi
jasonmreding: I had missed this in my initial review. Is there any reason you are doing this? Out update pin map TestStand step type (or other pin map clients) don't try to remove the BOM and seem to work just fine. I guess maybe [File.ReadAllText](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.readalltext?view=net-8.0) is doing some of this for us in C#? You could also look at using `Read From XML File.vi` to see if automatically handles BOM for you.

image

Source/Runtime/Clients/Pin Map V1/Create Client.vi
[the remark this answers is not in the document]
MounikaBattu17: Updated the wire bends.

Source/Runtime/Clients/Pin Map V1/Recreate Client.vi
MounikaBattu17: Thanks for the shortcuts. I've updated it accordingly.

Source/Runtime/Clients/Pin Map V1/Read Pin Map File Contents.vi
MounikaBattu17: The pinmap file contents are read with additional BOM characters `ï»¿`, so I trimmed them from the final string. But now, I verified that even with BOM characters, pin map is registered successfully to the pin map service. Thus, I have removed the trimming.

Source/Runtime/Clients/Pin Map V1/Pin Map Client FGV.vi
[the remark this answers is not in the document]
MounikaBattu17: Thanks for the explanation!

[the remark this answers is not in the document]
Prem-MysteryPro: @jasonmreding 
Thanks for the explanation.
I understand that FGV is always meant for singleton pattern. I have seen some FGV in our repository enabled as clone. For example, `NI Session Management V1 Client.lvlib:Session Client FGV.vi`
image
Another example is `NI Discovery V1 Client.lvlib:gRPC Client Cache FGV Impl.vi`
image
I am thinking there is reason behind this (Channel pooling). Kindly educate us on this.

Source/Runtime/Clients/Pin Map V1/Read Pin Map File Contents.vi
Prem-MysteryPro: Thanks for the suggestions, Jason. We tried `Read From XML File.vi`, I guess the XML schema of `Write To XML File.vi` is different from our pinmap XML so while using the `Read From XML" we are getting empty string as output so went ahead with normal read from text file. Removing BOM is not required and hence we removed that like Mounika mentioned in above comment.
image

Source/Runtime/Clients/Pin Map V1/Pin Map Client FGV.vi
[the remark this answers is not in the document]
jasonmreding: The `Session Client FGV.vi` is an oversight or bug of sorts. It should really be using non-reentrant execution. The idea is there is a FGV per service that we use (discovery, pin map, and session management), and this FGV serves as a cache for the client connection used to communicate with that service. However, the underlying implementation of the 3 FGVs is effectively the same. This is why `gRPC Client Cache FGV Impl.vi` exists and is marked as a reentrant VI with pre-allocated clones. This means a dedicated data space for the clone VI will get allocated up front at each call site. The call site in this case is the 3 FGVs which wrap it. Each of these is then marked as non-reentrant so there is only one call site for that service instance in memory. So the inner reentrant FGV is actually the thing with the shift register which caches the client connection and the outer non-reentrant FGV is what ensures there is a single instance for that service connection and that access is managed in a race free manner (i.e. a singleton).

Using a reentrant outer FGV with shared clones isn't necessarily functionally wrong. It just means if there are parallel calls to the FGV, a new clone and data space will get allocated on demand so the call can be executed in parallel with an ongoing call. The new clone will have a new allocation for the shift register which will be unitialized, which means it will create a new client connection for each clone. So in effect, you are creating N copies of the connection rather than reusing the existing connection.

This is all a bit hackish to get around the fact that grpc-labview doesn't expose a first class concept of a grpc channel like other languages do. If you read up about gRPC performance, it will say that channels should be long lived and that you should reuse them rather than constantly creating/destroying them. The initial establishment of the http connection is generally pretty expensive, and it usually consumes a TCP socket from the OS. Doing this too quickly can lead to socket exhaustion until enough time expires for the TCP connection to exit the time wait stage. So lacking an actual channel object, we are doing the next best thing which is to reuse the client itself for each service. This would be akin to reusing the client stub in other languages where the stub itself takes a grpc channel. In C# for instance, a common pattern is to create/destroy a new client stub for each RPC but to reuse the channel or http handler object. An alternative to implementing our own caching/pooling in G code would be to make the grpc-labview C++ just always use channel pooling for us as part of the implementation. However, that has its own trade offs when the user has no say in what is happening.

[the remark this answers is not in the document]
Prem-MysteryPro: Thanks for the Detailed explanation, Jason.
In this case we will leave the pin map FGV as non-reentrant.

Prem-MysteryPro (approved):

PR #578, opened 2024-08-27
Measurement Client Generator | Unary and non-pin-based measurement support

varshini-senthil:

[User story 2738651](https://dev.azure.com/ni/DevCentral/_workitems/edit/2738651): [LabVIEW] As a user, I want to interact with the non-streaming & non-pin-based measurement service

- Added template client library `Measurement Client Template.lvlib` to `Measurement Generator Template.lvproj`. This library holds the `measure.vi` template for non-streaming measurement client support.
    - The `measure.vi` exposes an optional input `pin map id` to support pin-based measurements. This input will be supported after the deployment of `Register Pin Map.vi`.
    - The `pin_map_context.sites` in `measure.vi` is set with a default value (=0), thus multi-sites concept is not supported.
    - The `measure.vi` will return the last data in case of streaming measurements.
- Implemented Measurement Plug-In Client creation from the template for the selected V2 measurement services.
    - `Int32`, `Int64`, `UInt32`, `UInt64`, `Float`, `Double`, `Boolean`, `String` and 1D arrays of the base datatypes are supported by the client.
    - `Enum` datatype is also supported by the client.
    - `Path` and `IOResource` datatypes are supported as string type specializations.
    - `DoubleXYData` and `DoubleXYDataArray1D` are supported only as output datatypes.
- Implemented error handling to handle the errors during client creation.
    - Client will be created and added to the `Active project` only if all the steps in the process are successful. Else, the client will be removed from both the project and folder, to ensure the user does not receive a broken client.
    - The error messages for each client are populated at the end of client creation process, for the user to review.
    - The status is displayed to the user only if any of the client creation fails, else the dialog will be closed at the end of the process, skipping the status review step.
- Updated the `Measurement Plug-In SDK Generator.vipb` to deploy the template client library in a folder `_Measurement Plug-In Client Template` parallel to the `Create Measurement Client.vi` under the `project` folder.
 
- Add template client library `Measurement Client Template.lvlib` to `Measurement Generator Template.lvproj`, as it is a dependency for the Measurement Plug-In Client Generator.
-  Implemented `measure.vi` creation from the template for the selected V2 measurement services.
- To update the `Measurement Plug-In SDK Generator.vipb` to deploy template client library as part of the package as it is a dependency for the Client generator.

- Generated package for Measurement Plug-In SDK Generator for the new changes and tested.
- Created clients for the example measurements and tested.


varshini-senthil:
Error Handling:

image

- User can view the error message after the completion of client creation process, in case of any failure during the process.
- Else, this step is skipped, and the dialog will close.

varshini-senthil:
`List Available Measurements.vi`
Diff image:
Before:
image

After:
image

varshini-senthil:
`Validate Selected Client Names.vi`
Diff image:
Before:
image

After:
image

Source/Generator/_MeasurementClient Utility/Validate Selected Client Names.vi
jasonmreding, line 1: Consider having both `selected measurement indices` and `selected client names` bound to the right side of the connector pane to keep them next to each other and moving `warning?` to the top or bottom of conn pane.

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Management Service.vi
jasonmreding, line 1: This same change was made in #576. I would suggest reverting it to avoid the conflict.

Source/Generator/_Measurement Plug-In Client Template/controls/XY Data Control.ctl
jasonmreding, line 1: I didn't immediately see a reference to this typedef anywhere. Did you mean to add this?

Source/Tests/Tests.Runtime/Measurement Server/Helpers/Kill Session Management Service.vi
varshini-senthil: This change to `Kill Session Management Service.vi` was made to fix the pipeline failure, referring to the PR mentioned. So, I can revert the change and rebase this branch, once the mentioned PR is merged to main.

Source/Generator/_Measurement Plug-In Client Template/controls/XY Data Control.ctl
varshini-senthil: This control is added as a template, for creating controls of `DoubleXYData` datatype, using the `Create from reference` invoke node in `Create Cluster Controls.vi`.

Source/Generator/_MeasurementClient Utility/Create Cluster Controls.vi
jasonmreding, line 1: This VI utilizes `Get all Object Items.vi` from the JSONtext addon from JDP Science. Unless there is a really compelling reason to add this dependency, I would prefer to eliminate it. Otherwise, we will need to update our .vipb to list this package dependency, and we will need to update the instructions for this repo to indicate the need to install this package when editing the client generator. Can you not just use the built in JSON primitive/VIs that LV provides?

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
bkeryan, line 1: You should test the generated client with a C# service too. I tried the Python version with the C# GameOfLife example and it returns an error when unpacking the measurement configurations (maybe because it's missing type_url?)

bkeryan: Note that the correct type names are in the MeasurementSignature message: 
https://github.com/ni/ni-apis/blob/main/ni/measurementlink/measurement/v2/measurement_service.proto#L78

Source/Generator/_MeasurementClient Utility/List Available Measurements.vi
jasonmreding, line 1: I'm curious what made you decide to delete the call to destroy the client here?

Source/Generator/Create Measurement Client.vi
jasonmreding, line 1: Rather than use a `Bundle` node which assigns names based on the labels of the data sources, create a private typedef and use `Bundle by Name` instead. This is much more readable and generally more maintainable as usage of the type expands in scope.

Source/Generator/_MeasurementClient Utility/Get Measurement Metadata.vi
jasonmreding, line 1: There are at least two places where you have a "ni.measurementlink.measurement.v2.MeasurementService" string constant. Wrap the constant in a VI so you can single source the constant. If you are concerned about performance, you can inline the VI.

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
jasonmreding, line 1: nit: Consider renaming `client name` to `new library name`.

jasonmreding, line 1: You have a conditional disable structure with a case for `RUN_TIME_ENGINE==True`. These VIs should never run in the runtime engine, just the IDE.

jasonmreding, line 1: I don't know if it is a lot better, but it might be a little less brittle to anchor to the app directory and then construct relative paths from that rather than relying on the current VI being two directories below where you want to go.

image

Source/Generator/_MeasurementClient Utility/Check Client Status.vi
jasonmreding, line 1: Rather than output a "Success" or "Fail" string, use a Boolean or typedef enum instead. Either that or use the fact that the `error message` is empty to infer pass / fail rather than having to compare against string constants.

Source/Generator/_MeasurementClient Utility/Create Cluster Controls.vi
jasonmreding, line 1: This VI name is confusing as it creates a control that can have a variety of data types and not just a cluster. It also only creates a single control and not multiple controls. Consider renaming to `Create Control` instead.

jasonmreding, line 1: The downcast for the array control is harmless, but it isn't strictly necessary. Also, the down casts aren't used else where in the VI so it is a bit inconsistent.

jasonmreding, line 1: I don't understand the nested case structure where the `type` enum is wired to the selector of both cases. I think it makes the code harder to read and reason about. I would recommend merging the two case structures into one.

jasonmreding, line 1: Rather than generate a variant control for types the generator doesn't expect, I would just generate an error indicating the type isn't supported. Or do the pack/unpack VIs somehow work with a variant type?

jasonmreding, line 1: I would recommend deleting the default cases in the case structures where an enum type is wired to the selector unless you have a large number of cases that all map to the same logic. In general, this is a little safer since the case structure will break if the type definition for the enum is edited to add, remove, or change the name of any items. This forces you to make explicit decisions about how this should be handled rather than things getting routed to default cases without any intervention.

jasonmreding, line 1: For the numeric types, you can consolidate code by using the same `New VI Object` call for every case and writing a conversion VI that maps the `type` enum to the appropriate `LVNumericRepresentation` enum.

jasonmreding, line 1: Ideally enumerated types defined by the service would get generated as typedefs. However, I understand if you plan to defer that as a feature for now. However, it makes maintenance of code which utilizes the generated client library easier to maintain as the measurement API changes over time.

Source/Generator/_Measurement Plug-In Client Template/controls/XY Data Control.ctl
jasonmreding: I see where it is being use now. At a minimum, I think it should be moved on disk so that it is under the generator library and added to the client generator library. However, I think you should also consider referencing the `ni_protobuf_types_DoubleXYData.ctl` within the ni.protobuf.types proto package that is installed to vilib. I don't feel strongly about that though. I guess if the result isn't linked to the original typedef, then it works out the same either way, and I would just use that typedef rather than duplicate another one.

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
jasonmreding, line 1: IMO this should really be working off typedefs that are included as part of the template for the client library rather than anonymous cluster controls/indicators defined on the front panel of the `Measure.vi`. This starts to stray into larger conversations about the design of the template/generated library and API though.

Source/Generator/_MeasurementClient Utility/Check Client Status.vi
jasonmreding, line 1: It seems like both cases of the case structure should close the reference. Maybe just move the close outside the case structure.

image

jasonmreding, line 1: Is there a reason not to just delete the entire library directory when generation fails rather than explicitly checking for the library file and `Measure.vi` individually and deleting them?

jasonmreding, line 1: I'm not certain what type of string this will generate in regards to the display. However, I suspect things will generally work out better if you use the `Simpler Error Handler` or `General Error Handler` VIs to do this sort of thing for you. However, I guess you would have to play with this a bit to see what the difference in user experience is. I know those VIs at least have the ability to look up error descriptions for certain error codes, but maybe that never really comes into play here. I also thought about just storing off the error cluster directly and showing that. However, maybe that information is too "raw"? I didn't really see any specific errors thrown by the generator code so I'm not sure what sorts of error you really expect to see here and whether any of it will make much sense to the user.

image

Source/Generator/_MeasurementClient Utility/Create Measure VI for Client.vi
jasonmreding, line 1: This isn't really creating a VI, it is modifying or editing a VI. Consider renaming to something like `Update Measurement from Metadata`.

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
jasonmreding, line 1: This isn't storing default values reported in the metadata for the configuration parameters. I don't know if that was an oversight, is out of scope, or is intended to be implemented later.

Source/Generator/_Measurement Plug-In Client Template/measure.vi
jasonmreding, line 1: I would make the label for the service class constant visible since the generator searches based on that. I would also update the literal value to something like "#<replaced by generator>" to make it even clearer.

image

jasonmreding, line 1: This should be using the message type reported from the metadata. The same goes for the unpack. The unpack probably doesn't matter since I think grpc-labview only cares the name matches what you registered with it. However, as Brad pointed out in another comment, a C# grpc service will use this information to do its own unpacking, and if it doesn't match what is expected (what it reported as the message type in its metadata), then it is going throw an error.

image

jasonmreding, line 1: I'm not sure what the official stance is, but I would encourage you to never use `optional` terminals. Always use `recommended` instead. I'm referring to `pin_mad_id` here.

jasonmreding, line 1: I would personally construct the template library so there were at least three VIs (and maybe a fourth). Basically, I would break the existing VI into three sub VIs that would be public as well. The naming can be tweaked as desired:
- Create Measurement
- Measure (or maybe Measure Unary)
- Close Measurement (or maybe just Close)

`Create Measurement` would include the code for resolving from the discovery service and registering metadata for packing/unpacking with grpc-labview.

`Measure` would include the code for packing the data, starting the Measure call, looping on responses until done, unpacking the results, and then ending the Measure call. 

`Close` would just simply close the client connection.

You could then create a wrapper VI around all three of these VIs if you want (basically what you have now). I'm not sure what a good name for it would be in light of the other VIs. However, this approach would provide more flexibility to users of the generated library if they want to make multiple measurement calls and don't want the overhead of resolving, connecting, and disconnecting each time which can be pretty significant.

jasonmreding (changes requested):
Waiting for next round of changes.

Source/Generator/_MeasurementClient Utility/List Available Measurements.vi
varshini-senthil: The `Destroy Client` wrapper API is removed from the `List Available Measurements.vi`, as the same client will be used by the `Resolve Service` API to get the location details of the selected Measurement Services, to get the measurement metadata. I have included the wrapper API in the `Create Measurement Client.vi` to destroy the Discovery client before closing the generator dialog.
image

Source/Generator/_MeasurementClient Utility/Check Client Status.vi
varshini-senthil: - The Client generator creates a folder for the measurement client, parallel to the Active project, in the same name as the client. The created client library and the `Measure.vi` are placed in this new folder. The library and the `Measure.vi` are removed from project and disk, if the client creation process fails.
- The tool reuses the folder in disk, if a folder with the same name as the client is already present. So, if we delete the entire library directory, we might end up deleting some files already present in the folder, which were not created by the tool. So, to achieve the concept of not providing a broken client to the user and also not to delete existing files in the folder (not created by the tool), I went with this approach.
I would like to understand your perspective on this

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
varshini-senthil: The default value support for the Measurement Client will be covered as part of upcoming PR, currently, the tool provides unary support for non-pin-based measurement, for all datatypes mentioned in the description.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: Thanks for pointing this out @bkeryan

Source/Generator/_Measurement Plug-In Client Template/measure.vi
varshini-senthil: The above `Message Type` works well with LabVIEW and Python Measurement services. But C# Measurement Service expects the `Message Type` parameter for the `Pack to Any.vim` to be the message type got from the measurement metadata. Additionally, it expects the label of the `measurement configuration` cluster to be the same.

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
varshini-senthil: We actually thought of taking up the typedef creation for input and ouput clusters as part of upcoming PRs, considering it might need some effort. Since, it requires less modifications, I will update the template client library to have two typedefs `measurement configuration.ctl` and `measurement results.ctl` and modify `Measure.vi` and `Measure Stream.vi` (to be added) to reference the typedefs for input and output clusters.

Source/Generator/_MeasurementClient Utility/Create Cluster Controls.vi
varshini-senthil: Considering your suggestion in this [comment](https://dev.azure.com/ni/DevCentral/_git/ASW/pullRequest/751696#1720634987), we are planning to take up the typedef creation for enum datatype later

Source/Generator/_MeasurementClient Utility/Check Client Status.vi
jasonmreding: In general, I think the two options are to:
1. Assume sole ownership of the directory and all files within it. To be safe, you can fail generation if the folder already exists or prompt that you are going to blap it and ask it the user wants to proceed.
2. Meticulously track all files which the generator might have added and only delete those files as part of cleanup. You can optionally delete the folder if it is empty when you are done.

There are trade offs to both options. The first option might work a little better if you want to regenerate the client library from the latest version of the service and have the library updated in place. It's also a lot simpler to cleanup on failure. The second option is a lot more brittle as is from a maintenance perspective as you have to be vigilant to remember to update the list of files every time you change the template. You could probably fix this by programmatically calculating the list of files from the template rather than hard coding them. You will still need to figure out and handle regeneration in place if you want to support that. Also, regeneration from one version of the template vs. another gets more complicated with this approach since you would have to track the file set for every version of the template over time and somehow correlate that back to which version of the template the existing code was generated from.

I don't necessarily have a strong preference between the two options. I just don't like the hard coded list of files that can get out of sync with the template. Because of that, I probably have a slight preference for just deleting the entire folder and perhaps prompting if you want to be extra safe about it.

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
jasonmreding: You can either use the following as is or create a helper VI around it. Ultimately, I think you want the tools directory or some folder under that so creating a helper for your root tools directory my be nice.

image

varshini-senthil:
Raised a separate [PR](https://github.com/ni/measurement-plugin-labview/pull/587) to fix an issue with loading the gRPC client API library `Measurement Service Client.lvlib` as dependency for the tool generated measurement plug-in client in a new LabVIEW project.

Source/Generator/_Measurement Plug-In Client Template/Close Measurement.vi
dixonjoel, line 1: You seem to have a convention for underscore at the beginning of the directory name. Does this imply the contents are private? Or intended to be internal? I hadn't noticed this convention in the repo before. No big objections, but I just wanted to be clear about it. @jasonmreding is this common or any thoughts on this?

dixonjoel:
I noticed this VI though it's not included in the PR.
image

* Change the name to something like Get Font Color.vi
* Use 'color' for the spelling
* Use a button instead of an LED for the boolean input
* I prefer using 'is valid' for the input name as opposed to 'valid?' especially for inputs. But I'll leave this up to you

https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/7839/LabVIEW-Style-Guide
"Make sure controls look like controls and indicators look like indicators. For example, use buttons for Boolean controls and LEDs for Boolean indicators."

Source/Generator/_MeasurementClient Utility/Check Client Status.vi
dixonjoel, line 1: image

Collapse unused inputs

dixonjoel, line 1: This VI does a lot more than 'check' the status. I like a name something like 'Cleanup Client If Error.vi'. I would even move all of the error checking and error message generation to another VI outside of this, but I'll leave that up to you to decide.

It even saves the project, which you'd never expect 'Check Client Status' to do. So maybe a name like 'Save Project or Cleanup Generated Code.vi'.

jasonmreding: These are inputs to an invoke node and not a property node. They can't be collapsed.

dixonjoel (changes requested):
Waiting for comments to be addressed

Source/Generator/_MeasurementClient Utility/Create Control.vi
jasonmreding, line 1: You should not assume index 1 is always the enum annotation. Instead, you should search for the array for the entry with the matching `key`,

image

jasonmreding, line 1: You can collapse `TYPE_GROUP`, `TYPE_BYTES`, and `TYPE_UNKNOWN` into a single case.

Source/Generator/_MeasurementClient Utility/Parse JSON Enum Values.vi
jasonmreding, line 1: I'm OK with this, but technically we probably shouldn't assume the enum items are in ascending order (which is how the `strings[]` property works). For a LV based service, this should be guaranteed if the service uses the helper VIs we provide. However, for Python, C#, or other languages, this might not be guaranteed. Similarly, other languages could use sparse enums which aren't supported by LV. You'll have to decide what you want to do with that case. You can just error or ignore that enum all together. There should be a C# test service that has both contiguous and sparse enums if you want to test with that.

Source/Generator/_MeasurementClient Utility/Provided Interface Constant.vi
jasonmreding, line 1: Consider renaming to something like `V2 Measurement Service` or `V2 Measurement Service Constant`.

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
jasonmreding: Did you mean to call `Template Files Path.vi` here instead?

Source/Generator/_Measurement Plug-In Client Template/controls/Measurement Configuration.ctl
jasonmreding, line 1: Consider deleting the Boolean controls from the default typedef template. That will simplify the generator code slightly since you don't have to worry about first deleting the existing control before dropping in the new control constants. However, this might mean more of the template VIs are broken, and I'm not sure what other challenges that might present so you are welcome disregard this comment if you find it makes life more difficult.

Source/Generator/_Measurement Plug-In Client Template/Create Measurement.vi
jasonmreding, line 1: This API should not have any measurement configuration/results as parameters. Convert the controls to constants instead.

Source/Generator/_Measurement Plug-In Client Template/Close Measurement.vi
jasonmreding, line 1: Don't apply standard error handling to this VI. Just call the underlying VI which always executes on error.

Source/Generator/_Measurement Plug-In Client Template/Measure.vi
jasonmreding, line 1: There shouldn't be a `measure results` input to this VI. Convert to a constant instead. Then move `pin map id` to where this input was and rename `measurement results unpacked` to `measurement results`.

Source/Generator/_Measurement Plug-In Client Template/Run Client.vi
jasonmreding, line 1: Consider keeping the connector pane assignments the same as `Measure.vi`.

Source/Generator/_MeasurementClient Utility/Folder Reuse Warning.vi
jasonmreding, line 1: Consider updating the message text similar to the following:

The following folders already exist on disk:
Directory1
Directory2
Directory3

Generation of the measurement plug-in clients may overwrite or delete files in these directories. Do you wish to continue?

Source/Generator/_Measurement Plug-In Client Template/controls/Measurement Configuration.ctl
varshini-senthil: I am a little concerned about the template VIs being broken, so I am going with the existing template

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
varshini-senthil: Yes, I have created a helper VI `Template Files Path.vi` to get the path of the template control and library

Source/Generator/_MeasurementClient Utility/Parse JSON Enum Values.vi
varshini-senthil: Currently, the Client generator creates a contiguous enum control for a sparse enum. Returning an unsupported datatype error for this case, will cause the client creation to fail. So, I think we can go with the existing way of generating a contiguous enum for both types, which will avoid the failure of client creation.
I would like to know your thoughts on this?

varshini-senthil:
> Can you include a screenshot or video showing the process and result of the client generation?

Client Generator

Source/Generator/_MeasurementClient Utility/V2 Measurement Service Constant.vi
jasonmreding, line 1: Update icon based on rename. I think something like "V2 Meas" would work.

Source/Generator/_Measurement Plug-In Client Template/Close Measurement.vi
jasonmreding, line 1: Typical error chaining behavior for this sort of thing is to preserve or give precedence to the original error that was passed in rather than overwriting it with whatever error was produced by the "cleanup" VI.

Source/Generator/_Measurement Plug-In Client Template/Create Measurement.vi
jasonmreding, line 1: Consider making this VI reentrant so that multiple calls can be executed in parallel.

jasonmreding, line 1: You will need to decide how you want to handle versioning as part of the first release of the generator. As is, the client will always connect to the latest version of the service. However, there is no guarantee the signature of the latest version will match the signature for the version from which the client was generated. The other alternative would be to extract the version from the metadata and script that into the cluster constant the same as the service class.

Source/Generator/_Measurement Plug-In Client Template/Measure.vi
jasonmreding, line 1: Consider making this VI reentrant so that multiple calls can be executed in parallel.

Source/Generator/_Measurement Plug-In Client Template/Run Client.vi
jasonmreding, line 1: Consider making this reentrant so multiple calls can execute in parallel.

Source/Generator/_MeasurementClient Utility/Create Control.vi
jasonmreding, line 1: Using the Conditional Terminal on the for loop to end the iterations early and then using a last tunnel value without conditional terminal would be more efficient.

image
image

Source/Generator/_MeasurementClient Utility/Template Files Path.vi
jasonmreding, line 1: I'm curious if there is a reason why you reverted back to relative pathing from the current VIs directory rather building the path from the root application directory?

varshini-senthil: The application directory path provides the path to the LabVIEW application folder. But the client template library files won't be available in the `project` folder until the tool is deployed as a package, which will prevent us from running the Client generator from the repo. 
So, I have used app directory path to build the path to the XY Data control and `Current VI's path` to build the path to the template library.

Source/Generator/_MeasurementClient Utility/Parse JSON Enum Values.vi
jasonmreding: If the service is reporting a sparse enum, you can't just treat it as contiguous from the LV client side. Doing so would pass enum values to the server that are incorrect and/or undefined. Your options are:
1. Error
1. Don't generate an enum for the field definition. Just use I32/U32 instead.
1. Generate an I32/U32 ring control instead. 

I forget how or if grpc-labview handles this. You can create a simple proto file which defines a sparse enum and try for yourself. I'm not sure how common sparse enums are with some of our driver C APIs (i.e. grpc-labview). If you don't believe they are uncommon, you could certainly defer some of this work. However, just always generating a contiguous enum is definitely not correct.

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
jasonmreding: This VI is still constructing the path itself rather than calling `Template Files Path.vi`.

image

jasonmreding (approved):

Source/Generator/_MeasurementClient Utility/Copy Client Template Library.vi
varshini-senthil: Thanks for pointing this out, I will update.

varshini-senthil:
Hi Joel,
I've fixed all the comments and am awaiting your review to proceed with this PR.

dixonjoel (approved):

PR #579, opened 2024-08-28
Add ni-apis as submodule and regenerate discovery client LabVIEW code

pakdev:

The [discovery_service.proto](https://github.com/ni/ni-apis/blob/7228d2bdfe49dd05586111b8c76ed6c1e0228355/ni/measurementlink/discovery/v1/discovery_service.proto) in the [ni-apis](https://github.com/ni/ni-apis) repo is more updated than the one here. For an upcoming change, I wanted the [ServiceDescriptor.versions](https://github.com/ni/ni-apis/blob/7228d2bdfe49dd05586111b8c76ed6c1e0228355/ni/measurementlink/discovery/v1/discovery_service.proto#L111) property. Thus, I added ni-apis/protobuf as submodules and regenerated the LabVIEW gRPC bindings from the new proto files.

Because our proto files should stay up-to-date and our discovery client LabVIEW code is now up-to-date.

Ran run_tests.vi


Source/Protos/google/protobuf/any.proto
dixonjoel, line 1: These are used by v1 and v2 measurement_service.proto files. Have you regenerated all of the various generated code (per the README that you're updating) with this arrangement to make sure it still works?

pakdev: Gotcha. No, I only regenerated discovery_service.proto because it was the only one with functional changes. I'll try one of those and see if it fails w/o the google files and add them back if it does

dixonjoel:
@pakdev Waiting to approve until the checks all pass.

Source/Protos/ni-apis/README.md
bkeryan, line 1: GitHub should display the proto files as deletes, not adds or moves. Are you sure you didn't add the contents of the ni-apis repo to this repo?

The PR should look more like this: https://github.com/ni/measurement-plugin-python/pull/703/files

image

bkeryan:
> I'm not super fluent with submodules, but I think everything under the ni-apis folder comes directly from that repo and is showing because you're adding the submodule? So I think we don't need to review anything under there?

That looks wrong. When you added ni-apis as a submodule to measurement-plugin-python, GitHub displayed the submodule as a link and didn't display all of the files inside.

pakdev:
> I'm not super fluent with submodules, but I think everything under the ni-apis folder comes directly from that repo and is showing because you're adding the submodule? So I think we don't need to review anything under there?

Right

pakdev:
On second look, it appears I shouldn't have pushed the contents. I'll fix that

Source/Runtime/Clients/Discovery V1/Service Location FGV Impl.vi
pakdev, line 1: image

Source/Tests/Tests.Runtime/Measurement Server/Discovery/Fake Resolve Service.vi
pakdev, line 1: image

Source/Protos/ni-apis/README.md
pakdev: It's correct now, right?

Source/Protos/google/protobuf/any.proto
pakdev: Added google protos back. I also moved the location of the submodule from `Source/Protos/ni-apis` to `Source/Protos` because generation expected `Source/Protos/ni/...` to exist (I had to make the Import Path `Source/Protos` with the google protos existing again). A nice benefit is the docs don't need to updated

Source/Protos/ni-apis/README.md
bkeryan: Yes, thanks

Source/Protos/google/protobuf/any.proto
bkeryan: > Added google protos back. 

Where are they? I have 555e9b7a2beb289b7d7fde100afbfff55000d6fe checked out and I don't see any.proto.

```
PS C:\dev\measurement-plugin-labview> cmd /c dir /s/b *.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\grpcdevice\v1\session.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\pin_map_context.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\discovery\v1\discovery_service.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\measurement\v1\measurement_service.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\measurement\v2\measurement_service.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\pinmap\v1\pin_map_service.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\measurementlink\sessionmanagement\v1\session_management_service.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\protobuf\types\precision_timestamp.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\protobuf\types\waveform.proto
C:\dev\measurement-plugin-labview\Source\Protos\ni\protobuf\types\xydata.proto
```

> I also moved the location of the submodule from Source/Protos/ni-apis to Source/Protos because generation expected Source/Protos/ni/... to exist (I had to make the Import Path Source/Protos with the google protos existing again).

Attaching ni-apis as Source/Protos fails the "what if two people did this?" test: what if we needed ni-apis, ni-internal-apis, and googleapis? We can't overlay multiple submodules at the same location in the Git repo.

I think the solution is to put them at different locations and use the import path to make protoc treat them as a single, overlayed directory hierarchy. Does grpc-labview support that?

Would it be possible to put the submodules in /third_party like in our other Git repos?

> A nice benefit is the docs don't need to updated

There's nothing wrong with updating the docs.

pakdev: > Where are they?

Ah, another nail in the coffin for the `Source/Protos` approach. I added them in that directory locally, but to avoid the same mistake as before (adding the submodule's files), they did not get added. I'll attempt to add them to `third_party`

jasonmreding: > I think the solution is to put them at different locations and use the import path to make protoc treat them as a single, overlayed directory hierarchy. Does grpc-labview support that?

An alternative might be to map the submodule into just the `Source/Protos/ni` folder. However, I don't know enough about submodules to know how possible/easy that would be to pull off. I know Boost does some interesting things with their [submodules](https://github.com/boostorg/boost/tree/master/libs) so you can only pull a subset of their source. However, I really don't have any idea how that works. However, even then, it might be a little weird to have submodules under the `Source` directory.

Otherwise, I think Brad's approach to mapping them in as something like `/third_party/protos/<submodule>` might make the most sense. However, I think that means each submodule would have to be its own import directory/path when generating code. That is annoying, but it is workable. grpc-labview does now support multiple imports now. If we go this route, we'll need to update the instructions for regenerating client/server stubs to include this new import path in addition to `Source/Protos` assuming we keep the google proto files there along with `session.proto`.

image

pakdev: Decided to add `ni-apis` and `protobuf` to `/third_party` as we do in [measurement-plugin-python](https://github.com/ni/measurement-plugin-python/tree/main/third_party). Cloning protobuf adds an extra 100MB to your hard drive. This can be trimmed down if desired by doing the following:

```
cd third_party\protobuf
git sparse-checkout init --cone
git sparse-checkout set src/google/protobuf
git read-tree -mu HEAD
```

However, all top-level files will remain, and src/google/protobuf includes more than *.proto. Apparently --no-cone can use file wildcards, but it's [not recommended](https://github.com/ni/measurement-plugin-python/tree/main/third_party). Annoyingly, submodules do not provide a mechanism to do --no-checkout. Anyway, I could get into more details, but if you absolutely hate protobuf being another submodule, I'll remove it and manually copy the 3 proto files we need.

After doing this, I regenerated all the grpc-labview clients/servers and updated the documentation.

pakdev:
Not sure what's going on with these tests - might need some help tomorrow morning. Not able to reproduce these 1003 errors in my branch...

Source/Protos/google/protobuf/any.proto
bkeryan: > An alternative might be to map the submodule into just the Source/Protos/ni folder. However, I don't know enough about submodules to know how possible/easy that would be to pull off.

This won't work because the submodule contains its own `/ni` folder.

> Decided to add ni-apis and protobuf to /third_party as we do in [measurement-plugin-python](https://github.com/ni/measurement-plugin-python/tree/main/third_party). Cloning protobuf adds an extra 100MB to your hard drive.

That sounds like the right approach. 

> Anyway, I could get into more details, but the best solution would be for the ni github account to fork the protobuf repo and remove all the files we don't care about.

measurement-plugin-python doesn't need to include protobuf as a submodule because the `grpcio-tools` Python package includes the protobuf `.proto` files.

It could make sense for grpc-labview to redistribute the protobuf `.proto` files like `grpcio-tools` does.

bkeryan: > I know Boost does some interesting things with their [submodules](https://github.com/boostorg/boost/tree/master/libs) so you can only pull a subset of their source. However, I really don't have any idea how that works. However, even then, it might be a little weird to have submodules under the Source directory.

Boost's submodules don't support this use case either.

For example, you can't map https://github.com/boostorg/lexical_cast into your project's `/include/boost/lexical_cast` directory and specify `-I $(GIT_ROOT)/include` on the compiler command line because the lexical_cast repo contains an `/include/boost/lexical_cast` subdirectory of its own:
https://github.com/boostorg/lexical_cast/tree/develop/include/boost/lexical_cast

You would have to add a separate `-I` parameter for each Boost submodule, like we are doing here.

jasonmreding (approved):

pakdev:
> @pakdev Waiting to approve until the checks all pass.

Passing now

dixonjoel (approved):

bkeryan (approved):

PR #586, opened 2024-08-30
renovate: Fix Git submodule scheduling

bkeryan:

The previous scheduling config was not working because `third_party/ni-apis` is a `depName`, not a `packageName`.

At one point, `matchPackageNames` used to match against dep names, but this behavior has been removed. https://github.com/renovatebot/renovate/discussions/25233

Enable "Dependency Dashboard Approval", which prevents Renovate from creating a PR until someone approves via the [Dependency Dashboard](https://github.com/ni/measurement-plugin-labview/issues/408).

Disable Renovate PRs for Git submodules unless approved via the [Dependency Dashboard](https://github.com/ni/measurement-plugin-labview/issues/408).

When I made the same change in another repo, I checked the Renovate log and verified that it listed the right schedule info.

"Dependency Dashboard Approval" is untested.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -7,7 +7,7 @@
     },
     "packageRules": [
       {
-        "matchPackagePrefixes": ["third_party/"],
+        "matchDepNames": ["third_party/ni-apis", "third_party/protobuf"],
         "extends": ["schedule:monthly"]
       }
     ]
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,7 +8,7 @@
     "packageRules": [
       {
         "matchDepNames": ["third_party/ni-apis", "third_party/protobuf"],
-        "extends": ["schedule:monthly"]
+        "extends": ["schedule:monthly", ":dependencyDashboardApproval"]
       }
     ]
   }


.github/renovate.json
```diff
@@ -7,7 +7,7 @@
     },
     "packageRules": [
       {
-        "matchPackagePrefixes": ["third_party/"],
+        "matchDepNames": ["third_party/ni-apis", "third_party/protobuf"],
         "extends": ["schedule:monthly"]
```
bkeryan, line 11: From @dixonjoel in https://github.com/ni/measurement-plugin-labview/pull/582#issuecomment-2322331921

> This is an improvement over getting an update for each commit. So I suggest we check this in for now and discuss further.
>
> @jasonmreding and I discussed this and I think we're of the same mind that we want to just turn off auto-updating altogether for the ni-apis repo (and maybe the protobuf). The reason for this is we don't have an automated way to generate stubs and / or check for out-of-date stubs like Python does. So if we auto-update, we'll have .proto files that don't match the generated code.

This should prevent Renovate from creating PRs until you approve via the Dependency Dashboard.

```suggestion
        "extends": ["schedule:monthly", ":dependencyDashboardApproval"]
```

jasonmreding (approved):

PR #587, opened 2024-09-03
fix: decouple code generated gRPC server and client API libraries of V2 Measurement Service

varshini-senthil:

- Decoupled the code generated gRPC client library `Measurement Service Client.lvlib` from the server library `ni.measurementlink.measurement.v2.api.lvlib` of V2 Measurement Service.
    - This is done to fix the issue with loading the child client library `Measurement Service Client.lvlib` when a Measurement Plug-In Client is created in a new project, which sets the cursor icon to busy icon.
    - This issue arises as LabVIEW tries to load the parent server library `ni.measurementlink.measurement.v2.api.lvlib` as dependency, while created client only uses the APIs in child client library `Measurement Service Client.lvlib`.
    
  Issue

- Updated `Regenerating Server and Client.md` to reflect the changes.

- To decouple the gRPC generated client and server API libraries to resolve the issue with loading the client API library `Measurement Service Client.lvlib` as dependency in the LabVIEW project.

- Created measurement plug-in client in a new LabVIEW project and tested.


dixonjoel (approved):

docs/Regenerating Server and Client.md
jasonmreding, line 1: I don't understand what this PR is trying to fix. Can you elaborate? I don't understand how our use of a busy cursor would necessitate this type of change.

varshini-senthil: When measurement plug-in client is created in a new project using the Client generator, LabVIEW tries to load its dependencies into the project.
Since, the created client has its dependencies in the child client API library `Measurement Service Client.lvlib`, LabVIEW tries to load the parent library `ni.measurementlink.measurement.v2.api.lvlib` also. 
This causes the project window to hang, as LabVIEW keeps trying to load the dependencies and save the project.

jasonmreding: I don't really know what to say about this. I was able to reproduce this and was a bit befuddled when trying to characterize the problem. I went ahead a filed a [bug](https://dev.azure.com/ni/DevCentral/_workitems/edit/2846576) to LV. I guess we'll have to wait and see if we hear anything from them regarding this. I guess in the meantime, we can remove the nesting and change it back later if we hear of a work around/fix. If we have to release something without a resolution, then we should think about what if anything we want to change regarding the library name. For example, maybe we should add a "V2" to the front of the library name so we leave room for client libraries to future versions of the service API (or the past V1 version).

varshini-senthil: Sure @jasonmreding, I will remove the nesting and rename the client library as `Measurement Service Client V2.lvlib` (considering the point that the version information is usually appended at the end of the library name) for the meantime.

jasonmreding (approved):

PR #590, opened 2024-09-11
Measurement Client Generator | Default value support for the client

varshini-senthil:

- Updated the Client Generator to retrieve default values from the Measurement Plug-In and update the client with the values.
- Updated the representation of Enum control created in the client to U32, to provide support for more values in the generated Enum.

- To provide default value support in the generated Measurement Plug-In Client.
- To update representation of the generated Enum control in the client.

- Created clients for example measurements and tested.


varshini-senthil:
`Create control.vi`
Diff image:
image
image

varshini-senthil:
`Modify Client Controls.vi`
Diff image:
Before:
image
After:
image

varshini-senthil:
`Update Measurement from Metadata.vi`
Diff image:
Before:
image
After:
image

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
jasonmreding, line 1: Consider generating an error in the default case for the control name. I don't think we expect to ever hit this. If we ever change the name of the controls in the template and forget to update this, it would be nice if something failed and forced our hand.

Source/Generator/_Measurement Plug-In Client Template/Default Value Template/Get Measurement Default Template.vi
jasonmreding, line 1: This wire branch for the `gRPC ID` is non-standard. Use the "out" terminal from the preceding VI rather than branching the wire.

image

Source/Generator/_MeasurementClient Utility/Get Default Value.vi
jasonmreding, line 1: Close the reference to this typedef once you are done using it.

image

Source/Generator/Create Measurement Client.vi
jasonmreding, line 1: I think it would be cleaner if you moved the call to `Cleanup Default Value Template` into `Get Default Value` and then modify the conn pane of `Get Default Value` accordingly. Similarly, I would recommend moving the call to `Get Default Value` into `Get Measurement Metadata` to keep the implementation more self contained rather than split across multiple VI calls from the caller's perspective.

image

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
jasonmreding: I see that you are also using this VI to modify the typedef in the default value template. In that case, I would recommend adding a "configuration" case that is effectively a noop and then have the default case generate an error of some sort.

jasonmreding, line 1: Consider renaming this to `Update Control from Metadata`.

Source/Generator/_Measurement Plug-In Client Template/Default Value Template/Default Value Template.lvlib
jasonmreding, line 1: From a file system perspective, I find it a bit odd this is nested under the client template folder even though it is a sibling to that library in the project pane. I think I would prefer to either create a sibling folder to `_Measurement Plug-In Client Template` where this new template lives or create two sibling folders under `_Measurement Plug-In Client Template` similar to the server side templates. In the latter case, all of the existing client template files would move under a new folder.

jasonmreding (approved):

Source/Generator/_Measurement Plug-In Client Template/Default Value Template/Get Measurement Default Template.vi
varshini-senthil: The `CompleteMetadataRegistration.vi` is wired with the gRPC id out of the `Register Message Metadata.vi` instead of the gRPC id from the preceding `Register Enum Metadata.vi` because of the bug in the `Register Enum Metadata.vi`, as mentioned below:
image

Source/Generator/_MeasurementClient Utility/Get Default Value.vi
varshini-senthil: The reference to the typedef is closed inside the `Modify Client Controls.vi` after saving the VI for the changes.

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
varshini-senthil: This will be handled by the `Open VI Object Reference` node, as it will throw 1054 error stating that the particular object was not found. 
image

Source/Generator/_Measurement Plug-In Client Template/Default Value Template/Get Measurement Default Template.vi
jasonmreding: I don't know from which VI that screen shot was taken, but it looks to me like this has been [fixed](https://github.com/ni/grpc-labview/issues/312). You can see that if you just look at the block diagram and see the VI is now using shift registers for the tunnel. Unfortunately, the same problem exists for the `Register Message Metadata.vi`. I will submit a separate fix for that. Realistically, I don't think we will ever have zero messages. I guess if you create a proto file with only enum definitions and no service definitions, then you could run into this. However, that doesn't apply in this case.

Source/Generator/_MeasurementClient Utility/Modify Client Controls.vi
jasonmreding: The caller could still pass in a valid name for a control on the diagram that no longer matches any of the control names in the case structure. For example, if the code was ever refactored and changed the name of "measurement configuration" control, it would be easy to miss updating the case structure as well. In that case, it would be really difficult to detect/diagnose why the default value is no longer being saved on the control.

PR #591, opened 2024-09-16
[minor] update: generate error if case not found for a given control name

varshini-senthil:

- Added custom error string in the global variable VI `Error Strings Global Internal.vi`, to indicate that no case exists for the given control name in `Update Control from Metadata.vi`.
    - This update is to alert the user about the change in control name in the Client template through error generation.

- To update the global variable VI `Error Strings Global Internal.vi`.
- To update the `Update Control from Metadata.vi`, to generate error when no case found for the given control name.

- Modified the measurement plug-in client template and tested error generation by creating a client.


dixonjoel (approved):

Source/Runtime/Measurements/Helpers/Error Strings Global Internal.vi
jasonmreding, line 1: This error file is used by our runtime libraries to produce user facing error messages. I would prefer not to reference it from our generators as well. Instead, you should create a separate error file in that project if that is the route you want to go.

Some other thoughts:
- "No case found for the control name '%s'. Update case structure for the new control name and try again." reads more as a developer error and not as a user facing error. If this is something that could bubble up to the user, you might want to just say something like: "Internal error. No case found for the control name '%s'.".
- In past PRs, you made design trade offs which allowed you to run the generator from source without having to install any dependencies to the LV directory. Linking to the error VI from <vilib>is inconsistent with those decisions. Creating a new error file will resolve that.

jasonmreding (changes requested):
Waiting to remove dependency on error file used by runtime libraries.

Source/Runtime/Measurements/Helpers/Error Strings Global Internal.vi
varshini-senthil: I have reverted the changes to the global VI in the `Measurement Server.lvproj` and have added a new global VI `Generator Error Strings Global.vi` for the generators.

Source/Generator/_MeasurementClient Utility/Generator Error Strings Global.vi
jasonmreding, line 1: If the goal is to share this error file across all of the generator libraries, I would probably move the file on disk from `_MeasurementClient Utility` to `_Generator Helpers`. If this is intended to be specific to just the client generator, then I would move the VI within the project items pane so it is under the Helpers folder of `MeasurementClient Editor.lvlib`.

jasonmreding (approved):

PR #598, opened 2024-09-23
update: set default value for the pin map sites

varshini-senthil:

- Updated the `Measure.vi` in the client template to set a default value for the `pin map context.sites` input, if the pin map id is empty string.
    - When the measurement client is run for a pin-based measurement using IOResource name, sites value of the `pin map context` input should be explicitly set to 0. 
    context: https://dev.azure.com/ni/DevCentral/_git/983dfe62-3b16-436f-b2ef-d98a72840b64/pullrequest/698395

- To set default value to the pin map `sites` input of the measurement client, when the `pin_map_id` is empty string.

- Generated client with the updated template and verified if default value is set for the `sites` input, when the `pin_map_id` is empty string.


varshini-senthil:
`Measure.vi`
Diff image:
Before:
image

After:
image

dixonjoel (approved):

Source/Generator/_Measurement Plug-In Client Template/Client/Measure.vi
jasonmreding, line 1: I don't understand why this change or the change from the linked PR are necessary. It sounds like we are making concessions due to implementation details for our Python measurement framework. The [service API](https://github.com/ni/ni-apis/blob/main/ni/measurementlink/pinmap/v1/pin_map_service.proto#L163) treats an empty site list as equivalent to all sites. Why doesn't the same thing work here, especially if we aren't even using a pin map? It seems a little crazy to me to require clients to explicitly populate a site array with one element when they aren't using pin maps.

Avinash2Suresh: We're facing an issue in measurements that use Get_Connections. Let's take a LabVIEW DC-Power measurement, and assume we are using IO-Resource names (with no pinmap registered).

Due to the fix Joel made [here](https://dev.azure.com/ni/DevCentral/_git/983dfe62-3b16-436f-b2ef-d98a72840b64/pullrequest/698395), Instrument Studio will pass the sites as [0], while our LabVIEW/Python client will pass the sites as [].

image

In the `Reserve Sessions - NPin.vi`, we are framing the `PinMapContext` from the Measurement context, which then gets propagated to `Reserve Sessions with Observer.vi`.

image

The value of `PinMapContext` in the session reservation is framed from the `PinMapContext` set by the client(For sites -> [0] for IS Pro, [] for LV/Py client), not by the session reservation response. As a result, when calling the `Get_Connections.vi`, we encounter the 'No Connections match pin or relay name(s)' error. in `Validate Get Connections outputs.vi`

bkeryan: @jasonmreding I agree this sounds like a bug in the Python library. @varshini-senthil should file a bug and we should try to fix it.

Avinash2Suresh: I think we should raise a similar bug in the LabVIEW library as well.

jasonmreding: I would prefer to also update `Get Connections Impl.vi` so that it produces a site array of [ 0 ] when the `pin_map_id` from the pin map context is empty. Wrapping the inner most case structure in a sub VI with this logic would probably be cleanest. If you still want to include this in the client side generator so services don't have to redeploy with the latest runtime library, I'm OK with that. However, given things are still relatively new and this is a new feature, I don't think asking services to update for this bug fix is unreasonable either.

image

Avinash2Suresh: @jasonmreding I feel we can take up the service-side changes as a separate PR and focus on the client fix alone as part of this PR. Since the client workaround is needed to handle Python measurements until the bug is fully resolved in Python, this approach seems practical.

We can raise another PR for the server-side changes related to the Get Connections Implementations. If you agree, please help review this PR.

jasonmreding: I would prefer to at least have an issue linked to this PR for tracking of the service side issue if you want to manage it as two PRs. I would like to have a service side fix for this issue as part of the upcoming release to minimize the window of time clients have to be concerned about it.

jasonmreding (approved):

PR #599, opened 2024-09-24
fix: set sites value if pin map id is empty in Get Connections Impl.vi

varshini-senthil:

- Updated logic to set the `pin map context.sites` value to [0] if the array is empty and the `pin map context.pin_map_id` value is also empty string.

- To fix the issue mentioned in the [PR](https://github.com/ni/measurement-plugin-labview/pull/598).

- Tested by running the client and verified if the `pin map context.sites` value is set to [0] for empty `pin_map_id`.


varshini-senthil:
`Get Connections Impl.vi`
Diff image
Before:

image

After:

image

dixonjoel (approved):

Source/Runtime/Clients/Session Management V1/Session Reservation/Get Connections Impl.vi
jasonmreding, line 1: I might just always ignore the sites array passed in when the pin_map_id is invalid and always use site 0 internally, but I don't feel strongly about it. I think it's unlikely a client would ever do that.

jasonmreding (approved):

Source/Runtime/Clients/Session Management V1/Session Reservation/Get Connections Impl.vi
varshini-senthil: image

@jasonmreding, I would like to understand, if you are mentioning to remove the check for empty sites array and set the `pin map context.sites` value to [0], if the `pin map context.pin_map_id` value is an empty string?
- If so, I have added the additional check for empty sites array, to make sure we don't overwrite the value already set to `pin map context.sites` (considering the case of InstrumentStudio as client, where the value of sites array is internally set to [0], when the pin_map_id is empty)

image

jasonmreding: The original code I commented upon appeared like it would fail if a client passed an empty `pin_map_id` but specified a `sites` that wasn't empty and only contained site one for some reason. That's why I was suggesting you could only check for a valid `pin_map_id` rather than the `pin_map_id` is not empty *AND* the `sites` array is empty.

I don't understand the comment about overwriting the value already set to `pin map context.sites`. At this point in the code, it appears we are treating that as a read only field so I don't think there is any danger of overwriting anything. We already pass a modified sites array to down stream code.

Basically, what I'm suggesting is always ignore whatever the user passed in for sites when the pin_map_id is empty and make it just work regardless. However, as I said, I think the chances of a client ever doing this are pretty small so I don't feel strongly about it either.

PR #600, opened 2024-09-24
update: add FP cleanup for the generated client VIs

varshini-senthil:

- Implemented logic to programmatically cleanup the Front Panel of the generated client VIs.

- To clean up the FP of the client VIs generated by the Measurement Plug-In Client Generator.

- Manually tested by generating client and verified the FP cleanup


varshini-senthil:
`Update Control from Metadata.vi`
Diff image:
Before:
image

After:
image

varshini-senthil:
`Update Measurement from Metadata.vi`
Diff image:
Before:
image

After:
image

Source/Generator/_MeasurementClient Utility/FP Cleanup.vi
dixonjoel, line 1: Why not just Call by Reference? What is the asynchronous call for? Also add a comment for the 'options' parameter.
image

dixonjoel (approved):
Approved with suggestions.

Source/Generator/_MeasurementClient Utility/Update Measurement from Metadata.vi
jasonmreding, line 1: It is a bit weird that we are checking for and handling the `Measure.vi` twice (once inside the case structure and once afterwards). It's also a bit odd that we end up saving it twice. Is there a reason we just don't handle everything within the case structure? I'm guessing it is because we need the typedefs to be updated and saved before we cleanup the front panel for the `Measure.vi` or otherwise it won't layout with the final area of the typedef controls?

Assuming that's true, you're effectively dealing with saving VIs in dependency order which isn't an easy thing to implement. Here are some suggestions you might consider. I don't feel strongly about them though so I'll leave it to you as to whether you want to adopt any of them. If nothing else, I do think it would at least be worth a code comment explaining why things are structured the way they are.
- Pass Boolean out of the case structure instead of duplicating compare of VI name against "Measure.vi". This single sources things better at the expense of some readability.
- Do a first pass sort of the VI refnums passed in and process all typedef .ctls before any VIs. This at least guarantees any typedef edits which might impact front panel layout of VIs have been made and saved. This would allow you to keep the logic for each VI self contained inside the case structure. This still doesn't solve typedefs that are dependent on other typedefs, but that's not an use case we need to worry about currently and can be deferred until it becomes an issue.

Source/Generator/_MeasurementClient Utility/FP Cleanup.vi
jasonmreding, line 1: nit: I find the use of "Client" in the VI icon odd. I would personally just use something like "FP Cleanup", "Cleanup FP", "Clean Up FP", or "FP Clean Up".

jasonmreding (approved):

Source/Generator/_MeasurementClient Utility/FP Cleanup.vi
varshini-senthil: Thanks for pointing this out @dixonjoel. I have updated the asynchronous call with `Call by Reference` node and removed the option flag `x100`, as it is used when we collect the results of an asynchronous call with the `Wait On Asynchronous Call` node.

Source/Generator/_MeasurementClient Utility/Update Measurement from Metadata.vi
varshini-senthil: I have updated the implementation to pass the boolean out of the case structure, as per your first suggestion.

PR #604, opened 2024-09-26
[High Priority] fix: fetch measurement version info from `Resolve Service with Information.vi`

varshini-senthil:

- Created a wrapper for `Resolve Service with Information.vi`, client VI of Discovery service and added the client wrapper VI to the `NI Discovery V1 Client.lvlib`.
- Created an automated testcase for the client wrapper and added the testcase to `Discovery Tests.lvlib`.
- Updated the `Get Measurement Metadata.vi` in Client generator to fetch the measurement version information from `Resolve Service with Information.vi` instead of `Get Metadata.vi` of the measurement client.

- Currently, Client generator fetches the measurement version info from the `Get Metadata.vi` method of the measurement client. The `version` feature handling is enabled only in the 2024 Q4 version of `Discovery Service` and the latest `Measurement Plug-In SDK Service` packages of LabVIEW and Python.
- For Python measurements (which use `Measurement Plug-In SDK v2.0.0` package) which do not have the version parameter, `Discovery Service` internally sets the version parameter as 0.0.1 during measurement registration.
- So, fetching the version information using `Get Metadata.vi` from the measurement causes version mismatch between the version specified in the measurement `.service config` file and the version `Discovery Service` has assigned to the measurement during registration. This mismatch causes the generated measurement client to fail in resolving the service.
- This update is essential to provide backward compatibility for the client with the previous `Measurement Plug-In SDK Service` package.

**Context:** [#946 Measurement client Generator does not support backward compatability for Older libraries](https://github.com/ni/measurement-plugin-python/issues/946)

- Tested by creating and running client for Python measurements which use Measurement Plug-In SDK v2.0.0 package.


varshini-senthil:
`Get Measurement Metadata.vi`
Diff image:
Before:
image

After:
image

Source/Runtime/Clients/Discovery V1/Resolve Service with Information.vi
jasonmreding, line 1: Make the request input required. While making this edit, I would update the icon text to just say "Info" rather than "Infor" which is the common abbreviation.

Source/Generator/_MeasurementClient Utility/Get Measurement Metadata.vi
jasonmreding, line 1: This has the same issue as [here](https://github.com/ni/measurement-plugin-python/pull/949#discussion_r1777580555). You should either gracefully handle the error and impose the same requirement of IS 2024Q4 or implement a fallback solution.

jasonmreding (approved):

Source/Generator/_MeasurementClient Utility/Get Measurement Metadata.vi
varshini-senthil: I have addressed the comment, by checking for empty version (InstrumentStudio Pro version Q3) and `Method not implemented error` with code -1012 (InstrumentStudio Pro version < Q3) and informing the user about the requirement of version Q4 or above for the client.

`Get Measurement Metadata.vi`

image

`Create Measurement Client.vi`

image

varshini-senthil: @jasonmreding, I would like to get your review for this approach

PR #606, opened 2024-09-26
Delete broken wires in Get Services JSON Path.vi that were causing build issues with our build process.

ccaltagi:
- [X]  Automatically post PR comments with images for G code changes?

Delete broken wires in `Get Services JSON Path.vi` that were causing build issues with our build process.

Fix FlexLogger Measurement Plugin Support build issue.


Source/Runtime/Clients/Discovery V1/Get Services JSON Path.vi
jasonmreding, line 1: The layout/wiring of this diagram seems a lot worse than it was previously. I'm guessing you used diagram cleanup? It would be nice to restore it closer to its previous state from a layout perspective.

jasonmreding, line 1: Consider combining the two conditional structures into a single structure. I think that will simplify some of the error chaining.

jasonmreding, line 1: You have a VI analyzer check failure you will need to fix before you can submit. I'm guessing it's because of this?

image

I suspect if you wire the error in tunnel to a merge error to retain the original error, it will probably fix the issue. That might be a good idea to do regardless.

jasonmreding (approved):

Source/Runtime/Clients/Discovery V1/Get Services JSON Path.vi
ccaltagi: Fixed.

ccaltagi: Fixed

ccaltagi: Fixed.

PR #611, opened 2024-10-04
[High Priority] Measurement Client Generator | Stream response support for the measurement client 

varshini-senthil:

- Separated the logic inside the `Measure.vi` in the measurement client template library `Measurement Client Template.lvlib` into three VIs to get the streaming response of the measurement service in the caller VI, by looping the `Read Measurement.vi`. The new VIs are:
    - `Start Measurement.vi`
    - `Read Measurement.vi`
    - `Stop Measurement.vi`
    
- Updated the `Run Client.vi` to display the continuous response of a streaming measurement.
    - The modified `Run Client.vi` will support both non-stream and stream measurements.
    - The VI will provide the last value of the stream response when used as sub VI in a framework.
    - User has to replicate the logic inside the `Run Client.vi` in their framework, to get the complete stream response instead of using the `Run Client.vi` as a sub VI.

- Updated the `Template Files List.vi` to contain the list of files in the modified client template library.

- Updated the scripting logic in `Update Measurement from Metadata.vi` for the modified client template.

- Updated the execution property of `Close Measurement.vi` to re-entrant so that multiple calls can execute in parallel.

- To provide stream response support for the measurement client.

- Manually tested by creating measurement clients for the example measurements


varshini-senthil:
`Copy Client Template Library.vi`
Diff image
Before:
image

After:
image

varshini-senthil:
`Create Measurement Client.vi`
Diff image
Before:
image

After:
image

Source/Generator/_MeasurementClient Utility/Template Files List.vi
dixonjoel, line 1: Don't we need Close Measurement.vi?
image

dixonjoel:
Have you done any testing with a 'V1' measurement service or is this supposed to work with V1 services? It looked to me like you're always making a V2 measure call. Here's a couple examples that are V1 only.
image

Source/Generator/_Measurement Plug-In Client Template/Client/Start Measurement.vi
dixonjoel, line 1: Have you done any testing with a 'V1' measurement service or is this supposed to work with V1 services? It looked to me like you're always making a V2 measure call. Here's a couple examples that are V1 only.
image

dixonjoel: It may not be a requirement to work with V1 services. I think we'd probably be ok with that, just checking.

dixonjoel (approved):
Approved with a couple of minor outstanding questions open.

Source/Generator/_MeasurementClient Utility/Template Files List.vi
varshini-senthil: The `Template Files List.vi` contains the list of VIs that require modifications specific to a measurement service, which are to be done through scripting. Since `Close Measurement.vi` is a wrapper VI for `Destroy Client.vi`, no measurement service-specific modifications are needed, so it is not included in the list. The same applies to `Stop Measurement.vi` in the client template library.

Source/Generator/_Measurement Plug-In Client Template/Client/Start Measurement.vi
varshini-senthil: In line with the HLD, client support will be provided exclusively for V2 Measurement Services. Additionally, the client generator will list only the available V2 measurement services by calling the `Enumerate Services.vi` with the interface specified as “ni.measurementlink.measurement.v2.MeasurementService”.

image

Therefore, we are hardcoding the interface as “ni.measurementlink.measurement.v2.MeasurementService” in the client, ensuring that the client will be listed and generated only if it is a V2 measurement service.

jasonmreding (approved):

PR #612, opened 2024-10-04
[High Priority] Added cancel support to measurement client

MounikaBattu17:

This PR provides support for canceling an ongoing measure call and demonstrates how to use the cancel support in `Run Client.vi`.

- Added a new VI `Abort Measurement.vi` which cancels given `Call ID` using grpc `client cancel call.vi`.
- In `Run Client.vi` demonstrated the usage of the above new VI.
     - Added a new `stop client` button. Upon clicking on the button, the ongoing gRPC `Read Measure.vi` should be aborted.
    - Added a parallel event structure that captures the Value Change event of the stop client button where `Abort Measurement.vi` is called. This will be responsible for canceling the `Read Measure.vi` call, thereby stopping its while loop.
   - To terminate the stop while loop, created a custom stop client user event that will stop the while loop when triggered after measure call completion.
  - Suppressed all errors while generating stop client user event.
  - Using this support, we will get an Invalid argument error, indicating that the measure call was abruptly aborted. We are explicitly suppressing that error.

**Additional changes:**
- As we are not intending to provide the cancel button as a input to the `Run Client.vi`, we are not including the control in the connector pane. But if the control is not connected to the connector pane, the control will be positioned in the bottom , after programmatical Front panel cleanup. So, we are connecting the cancel button to the connector pane programmatically before cleanup and then performing the FP cleanup as usual. Finally, we disconnect the cancel button and save the VI for the changes in `Update Measurement from Metadata.vi`.
- In `Read Measurement.vi`, updated the measure success boolean to check for both success from Measure Read and error case also.

**Note:**
- We are proposing Run Client.vi as a stand alone example VI that demonstrates the usage of all VIs in `Measurement Client Template` library. So the cancellation support in `Run Client.vi` is again a demonstration of how we can stop a continuously running measurement. 
- `Run Client.vi` can still be used as sub VI for non-streaming measurements. But for streaming measurements and cancel support will have few modifications. 
     - For streaming measurements, the last streamed response will only be obtained. So to get all responses, users have to tweak the logic in Run Client accordingly.
    - For cancel support, users should tweak some logic and provide custom user events to Run Client.vi.

Manually verified the created client with cancellation support.


MounikaBattu17:
File Diff:

Run Client.vi: (modified)
image

image

Cancel Measurement.vi: {added)
image

dixonjoel:
Can you add a screenshot showing what the front panel of Run Client.vi looks like when cancelled? Is that the main VI the user runs to run the client?

dixonjoel:
Does Cancel Measurement need to be added to the Template Files List?

Source/Generator/_Measurement Plug-In Client Template/Client/Run Client.vi
jasonmreding, line 1: The read loop is going to stop on an error or upon a successful read. I think you want to invert the success check.

MounikaBattu17: Yes! Sorry, I missed to check that. I have reverted the change and updated the VI such that the read loop stops when the streaming is ended or if any error is encountered.

MounikaBattu17:
> Can you add a screenshot showing what the front panel of Run Client.vi looks like when cancelled? Is that the main VI the user runs to run the client?

Here below I've attached a small gif of how the measurement client (Game of Life Python example) could be used with cancel support.
Game of Life client


> Is that the main VI the user runs to run the client?
Yes! `Run Client.vi` demonstrates the usage of all VIs provided in the client library.

MounikaBattu17:
> Does Cancel Measurement need to be added to the Template Files List?

Cancel Measurement is an additional support to cancel any ongoing measure calls. Users can use the `Cancel Measurement.vi` to stop any streaming measurements or long-running non-streaming measurements. Similar to the existing start and read measurement functionalities, we've integrated the cancel option into the template file, enabling users to utilize it as demonstrated in the `Run Client.vi.`

Source/Generator/_Measurement Plug-In Client Template/Client/Run Client.vi
jasonmreding, line 1: Several things here:
- It doesn't make sense to have `cancel` on the connector pane. If it's true when you call `Run Client`, then you shouldn't have called it to begin with.
- I would change the name of the `cancel` control to `stop`. That seems more consistent with typical LV examples one would see.
- I think the overall template/example might need some more thought. Is the purpose to effectively serve as a shipping example or as reusable logic that can be called as a sub VI? If it's an example to demonstrate a concept (i.e. how you might program cancellation) that is only intended to be used as a top level VI, then having UI controls that aren't mapped to the connector pane and cancelling in a somewhat contrived way might be OK. However, if the goal is for somebody to be able to call/reuse this as a sub VI, then this will need some more work. In the latter case, I think you will need to take in a user defined event or something like that instead. Some other things that I think you will want to revisit regardless:
  - The read loop is always triggering cancel after the first iteration by signalling the user event. This will have to go away anyway if you decide to pass in a user event instead.
  - The unconditional clearing of errors in the read loop seems odd. Does a canceled call always produce an error from the read call?

jasonmreding (changes requested):
Waiting for next round of changes.

Source/Generator/_Measurement Plug-In Client Template/Client/Run Client.vi
Avinash2Suresh: >Is the purpose to effectively serve as a shipping example or as reusable logic that can be called as a sub VI?

I believe having it as a shipping example would make more sense. This way, the user can directly run the VI and take advantage of the cancel/stop button functionality. If they want to integrate this logic into their own framework, they can easily replicate or copy and paste it into their code.

On the other hand, providing it as a subVI might complicate things, as the user would need to handle passing events from the caller and triggering them when necessary. This could add unnecessary complexity.

@jasonmreding I would like to understand your opinion on choosing between these two approaches.

Avinash2Suresh: >I would change the name of the cancel control to stop. That seems more consistent with typical LV examples one would see.

I agree. However, since we already have a VI named `stop measurement`, using 'stop' for this as well might confuse the user. That's why we've chosen to refer to this as 'cancel' instead, to provide a clear distinction.

MounikaBattu17: > I would change the name of the cancel control to stop. That seems more consistent with typical LV examples one would see.

Renamed the button to `stop client`.


>  If it's an example to demonstrate a concept (i.e. how you might program cancellation) that is only intended to be used as a top level VI, then having UI controls that aren't mapped to the connector pane and cancelling in a somewhat contrived way might be OK.

Yes, we are proposing the Run Client as an example to demonstrate how to other VIs in the library. We added a button to the connector pane to facilitate proper positioning during front panel cleanup. However, since placing the button in the connector pane is inappropriate, we have updated the `Update Measurement from Metadata.vi` to position it correctly and disconnected the stop client button from connector pane.

> The read loop is always triggering cancel after the first iteration by signalling the user event.

Positioned the user event trigger outside the loop so that the completion of the Read Measurement will also terminate the Stop Measurement Client loop if it still running.

@jasonmreding  Please share your thoughts on this.

jasonmreding, line 1: -1 is the error code for `CANCELLED` and not `INVALID_ARGUMENT`.

jasonmreding, line 1: You may also want to stop the Read Measurement loop on a read error as well.

jasonmreding (approved):

Source/Generator/_Measurement Plug-In Client Template/Client/Run Client.vi
MounikaBattu17: Since we have already checked for errors and updated the `measure success` boolean based on the results in `Read Measurement.vi`, should we still include the error check here? The `measure success` boolean will already be `false` in the event of an error.
image

jasonmreding: I didn't realize that logic was already included in the Read VI. I think this is fine as is then.

dixonjoel (approved):

Prem-MysteryPro (approved):

PR #616, opened 2024-10-09
[minor] Add Sparse Enum support to Measurement Client

varshini-senthil:

- Currently, we generate `Unsupported data type` error for sparse enums, as it is not supported by LabVIEW. But measurements in Python and C# may contain sparse enums. Ex: [nidmm_measurement](https://github.com/ni/measurement-plugin-python/tree/main/examples/nidmm_measurement).
- So, updated the `Create Control.vi` to create a `Ring` control for Sparse enums and set both the item names and values to the control based on the metadata.
- Updated the `Parse JSON Enum Values.vi` to get the list of `enum items` and also `enum items and values`.
- Updated the controls in `Create Control.vi` and `Map Numeric Representation.vi` to remove coercion dots. Also, removed the `Merge Errors` node in `Create Control.vi` as it is redundant because of the standard error handling.

- To provide sparse enum support to measurement client.

- Manually tested by creating client for python measurements with sparse enums.


varshini-senthil:
`Create Control.vi`
Diff image

Before:
image

After:
image

varshini-senthil:
`Parse JSON Enum Values.vi`
Diff image

Before:
image

After:
image

Source/Generator/_MeasurementClient Utility/Parse JSON Enum Values.vi
dixonjoel, line 1: I don't see a reason to return the items and items and values in two different arrays. This is probably only called in a couple places, so I would remove the 'enum items' output and update callers to unbundle.

dixonjoel (approved):
Approved with a minor suggestion.

Source/Generator/_MeasurementClient Utility/Parse JSON Enum Values.vi
varshini-senthil: I've updated as per your suggestion.

varshini-senthil: The `Enum` control only expects the item names as an array of strings, while we can set both the item names and values as an array of clusters (string and value) for a `Ring` control. If we are getting only the `enum items and values` as the output from `Parse JSON Enum Values.vi`, we have to again iterate through the array of clusters, to get the array of strings for a `Enum` control.

I think using another for loop in the caller to unbundle each cluster from the `enum items and values` array to get the `enum items` is redundant, when we can get the same during the intial parsing inside `Parse JSON Enum Values.vi`. 

Would like to know your thoughts on this.

dixonjoel: Yeah that's fine. We can keep as is.

PR #631, opened 2025-02-04
feat: Create Wrapper VI for Measurement Client Generator Core Logic

varshini-senthil:

[Task 3008803](https://dev.azure.com/ni/DevCentral/_workitems/edit/3008803): Create wrapper VI and update Client Generator tool

- Implements a wrapper VI `Generate Client.vi` around the core logic helper VIs of the 'Measurement Plug-In Client Generator' tool.
- Adds a new library `NI Measurement V2 Client Generator Runtime.lvlib` to the `Measurement Generator Scripting.lvproj` project to hold the wrapper VI and its dependencies, which were moved from the existing `MeasurementClient Editor.lvlib`.
- Updates the existing 'Measurement Plug-In Client Generator' tool to use the wrapper VI for its core logic instead of using the helper VIs individually.

- To create a wrapper VI for the core logic of the Client Generator tool and a new library to hold the wrapper VI and its dependencies.
- To update the tool to use the wrapper VI for its core logic, simplifying the implementation.

- Built package and manually tested the wrapper VI and the Client Generator tool to ensure they are working as expected.


varshini-senthil:
_Code changes:_
**Changes to `MeasurementClient Editor.lvlib` library:**
- The helper VIs required for the wrapper VI are moved from the current `MeasurementClient Editor.lvlib` to the new library `NI Measurement V2 Client Generator Runtime.lvlib`.
- The access scope of the helper VIs which are also required in the existing 'Measurement Plug-In Client Generator' tool is updated to community scope for reuse in the tool.

varshini-senthil:
**Changes to `Cleanup Client If Error.vi`:**
- The `Cleanup Client If Error.vi` logic is updated to:
   - Removed the `error message` and `status` outputs and modified to pass out the original error that triggered the cleanup.
     _**Before:**_
     image
     _**After:**_
     image
   -  Remove the logic that converts the error cluster into an error message.
   **_Before:_**
    image
   **_After:_**
    image

varshini-senthil:
**Changes to `Create Measurement Client.vi`:**
- Updated the `Create Measurement Client.vi` to use the new wrapper VI (Generate Client.vi) and also added the required error handling logic for the tool as a subVI `Handle Error in Client Generation.vi`.
**_Before:_**
image
**_After:_**
image

varshini-senthil:
**Changes to `Copy Client Template Library.vi`:**
- Updated the `Copy Client Template Library.vi` to prevent losing the actual error source when '1051' error is encountered at the `Library: Save as Copy` invoke node.
_**Before:**_
image
_**After:**_
image

varshini-senthil:
**Changes to `Get Measurement Metadata.vi`:**
- Removed the `version < Q4?` boolean indicator and replaced the logic that provided a warning to the user through a dialog when the InstrumentStudio version was less than Q4 with logic to raise an appropriate error to indicate the same.

_**Before:**_
image

_**After:**_
image

varshini-senthil:
**Changes to `Template Files Path.vi` and client template files folder name:**
- Updated the `Template Files Path.vi` to align with the new deployment path for the client template files.

_**Before:**_
image
_**After:**_
image

- Renamed the client template files folder from `_Measurement Plug-In Client Template` to `Measurement Plug-In Client Template` to align with the measurement plug-in template files folder name, as the client template files folder will be deployed in parallel.

Source/Generator/Measurement V2 Client Generator Runtime/Generate Client.vi
jasonmreding, line 1: This applies to all VIs in the library, but I'm adding the comment to this file. I'm not sure what happened, but all of the icons in the library are 100% "local" VI icon rather than a combination of overlays. In particular, I would expect all of the text to be specified via the Icon Text tab of the icon editor rather than as pixels of the VI Icon layer.

image

Similarly, each VI in the library should have a `NI_Library` layer so that the icon for the library is automatically overlaid in the final rendering of the icon. You can see this if you look at VIs within the editor library.

image

You can then create/edit the library icon from the properties page of the library. It looks like you have already done this. I believe if you click the "Apply Icon To VIs", it will automatically add the `NI_Library` layer to all of the VIs in the library, but I'm not certain. 

image


I'm pretty sure you will have to "save all" after doing this though. You'll want to be careful about doing this with existing libraries though as I believe it will create an edit to every VI in the library regardless of whether or not it is already up to date and is synchronized to the library's icon. If you only want to apply the change to a single VI at a time, you can use the Edit -> Import Icon from Owning Library command instead.

image

jasonmreding, line 1: The icon text overflows a bit for this public icon. I would recommend just using "Gen Client" on two lines for this VI as I feel that is a pretty common abbreviation for icons. In general, you might want to audit all of the icons from the library once you are done for general readability/consistency of the icons. The public APIs are most important, but it can also go a long way towards readability/maintainability for future maintainers of the code as well.

jasonmreding: It looks like the new VIs you added to the other library have the same issue.

Source/Generator/Measurement V2 Client Generator Runtime/Copy Client Template Library.vi
jasonmreding, line 1: > The shift register for the error wire in the for loop clears the error source if the loop does not iterate even once. To prevent this behavior, which leads to the loss of the actual error source, a case structure has been added around the loop to check for errors before the loop executes.

This is a well known "gotcha" with LV code. The typical way to prevent loss of the error information is to simply convert the tunnels to shift registers on the for loop rather than wrapping everything in a case structure. If there are other reasons to skip execution of some portion of the code on error, introducing the case structure is fine. Otherwise, I would just convert to shift registers as the diagram is easier to read.

Source/Generator/Measurement V2 Client Generator Runtime/Template Files Path.vi
jasonmreding, line 1: I know you are copying what had been done before, but this is a bit non-standard way to manage file paths for <vilib>. Instead, you can just use the `VI Library` constant from the palette.

image

jasonmreding, line 1: With this change, you will be force to manually copy the template files to <vilib> when running from the project or have a previous version of the package installed which happens to include all of the same template files with the same template names. If you wanted to preserve the dev workflow without requiring this step, there are probably a few different things you could try:
- Query the Application's `Project.OwningProject` property. If this produces an error, you know it is not executing within the project context. In that case, you can infer you are running from the tool's menu application context and return the paths you have here. Otherwise, you can infer you are running from the source project context and return paths that are relative to this VI's location in the source hierarchy.
- Similar to above but query the current Application's `ContextName` property instead. If it is "My Computer" then you are executing from the source project and not from the <project> folder via the tool's menu application context.

I'm not sure if either of these will cause any trouble for tests or not. I haven't looked to see if the tests execute from the source project context or if they are loaded as "loose" VIs. I guess if it is a problem, it is probably already a problem as a result of these changes that you will need to address anyway.

jasonmreding, line 1: I don't know what the expectation of this feature is in regards to execution via LV Runtime Engine vs. the Dev Environment. There is no <vilib> for the runtime engine so if you want that to work, you will need to develop other strategies for how to deploy and locate the template libraries. That might be as simple as requiring the user to create a buildspec which manually copies the template libraries to a known relative location from the application's root directory. However, this could would still need to return the correct path for that deployment scenario. If this use case isn't something you need to support, I'm OK with that as well. I just wanted to point out in its current state, built apps can't utilize this API. Well... I guess they could, but they would need to copy"vi.lib\Plug-Ins\Measurement\Measurement Plug-In Client Template\..." directory into the root directory of their built app. If you use the `VI Library` constant, then the directory would be "Plug-Ins\Measurement\Measurement Plug-In Client Template\..."

jasonmreding (changes requested):
I would at least like to see the icons cleaned up to something more standard before approving.

Source/Generator/Measurement V2 Client Generator Runtime/Template Files Path.vi
varshini-senthil: The helper VIs of the tool use scripting and LabVIEW's scripting functions are generally not supported in built applications. So, we will not be able to provide LV Runtime Engine support for this feature.

varshini-senthil: I have updated the `Template Files Path.vi` to use the `VI Library` node instead of the App directory property node. Additionally, I have added a check for the owning project to enable running the tool from both the source code and the deployed location without requiring a new package installation, as suggested.

image

varshini-senthil: I have reverted the changes made to the VI to enable running the tool from both the source code and the deployed location because:

- Imagine a custom application that uses the `Generate Client.vi` wrapper VI to generate clients programmatically and the application-related VIs are maintained within a LabVIEW project.
- In this scenario, when the application runs and calls the wrapper VI (as a sub VI), the `Project.Owning project` will not throw an error since the calling VI (`Template Files Path.vi`) is owned by a project. This results in an incorrect path to the template files.
- I suspect the same issue will occur with the current application's `ContextName` property node.
- Since this is not the expected behavior, I am reverting the changes to this file.

Considering these downsides, I prefer to stay with the current implementation and either manually copy the template files to their deployment location `<vi.lib>\Plug-Ins\Measurement` or install the latest package to test the tool during development work. Also, this step is required only for the first time, and the developer does not need to repeat it for every change.

jasonmreding: That is a good point and something I didn't think about. One last thought, you could create a conditional symbol which is only defined in the source project. That would only be an issue if the end user defines the same conditional symbol in their project which should be extremely unlikely, and I believe it would also work when executing from the tool's menu application context. You would then just replace the case structure above with a conditional disable structure.

image

jasonmreding: That's a good point. I thought we were doing basic enough things that the scripting methods would be supported, but I guess anything involving project item related classes requires the full featured runtime engine which isn't something customers generally have access to.

Source/Generator/_MeasurementClient Utility/Handle Error in Client Generation.vi
jasonmreding, line 1: The font color for the first line of text is a little different from everything else. I assume this was a stray click at some point and not intentional.

image

jasonmreding (approved):

Source/Generator/Measurement V2 Client Generator Runtime/Template Files Path.vi
varshini-senthil: Sure, Jason. I have updated the project properties and the `Template Files Path.vi` to add a conditional disable symbol to check for the source project when building the path to template files. I have named the symbol `RUN_FROM_CLIENT_GEN_SOURCE` to make it unique for the `Measurement Generator Scripting.lvproj` and reduce the likelihood of other projects having a symbol with a similar name.

image

PR #632, opened 2025-02-04
refactor: Relocate 'Palette Files' folder and Update Build Spec

varshini-senthil:

[Task 3008804](https://dev.azure.com/ni/DevCentral/_workitems/edit/3008804): Relocate 'Palette Files' folder and update Build Specs

- Relocates the `Palette Files` folder from the `Source\Runtime` directory to the `Source` directory to make the `Measurement Plug-In SDK.mnu` file common for both Generator and Service packages.
- Updates the `Measurement Plug-In SDK Service.vipb` build spec and its Pre and Post Build Custom action VIs to align with the relocation.

_Context:_
- Currently, the `Measurement Plug-In SDK.mnu` menu file is located in the `Source\Runtime\Palette Files\Measurement` directory and is deployed as part of the `Measurement Plug-In SDK Service` package, as only the VIs from the Service package are exposed in the palette.
- As we are now introducing a new wrapper VI from the 'Measurement Plug-In SDK Generator' package into the palette, the `Palette Files` folder will be moved from the `Source\Runtime` directory to the `Source` directory. This change will make the `Measurement Plug-In SDK.mnu` file and the `Palette Files` folder common to both packages.

- To relocate the `Palette Files` folder to make the `Measurement Plug-In SDK.mnu` file common to both Service and Generator packages.
- To update the `Measurement Plug-In SDK Service.vipb` build specification and its custom action VIs to align with the relocation.

Built the package and tested it to ensure the changes are working as expected.


varshini-senthil:
**Code changes to `Pre-Build Custom Action Service.vi`**

_Before:_
image
_After:_
image

varshini-senthil:
**Code changes to `Post-Build Custom Action.vi`**
_Before:_
image
_After:_
image

Source/Palette Files/Measurement/Measurement Plug-In SDK.mnu
jasonmreding, line 1: It's not clear to me that you needed to move these palette files up a directory in the source hierarchy. I guess it doesn't necessarily hurt anything, but it does create a bit of noise in regards to the other changes. Is there something I am not seeing why this is required?

Source/Build Specs/Measurement Plug-In SDK Service.vipb
```diff
@@ -136,27 +136,21 @@
...
       <Destination_Overrides>
-        <Path>Runtime/Clients</Path>
+        <Path>Runtime/TestStand</Path>
         <Destination>0</Destination>
         <Additional_Destination>false</Additional_Destination>
         <Additional_Destination_Index>0</Additional_Destination_Index>
       </Destination_Overrides>
       <Destination_Overrides>
-        <Path>Runtime/TestStand</Path>
+        <Path>Palette Files/Measurement</Path>
```
jasonmreding, line 151: With this change, you now have two packages which are installing `Measurement Plug_In SDK.mnu`. I can't necessarily think of anything that will break because of this, but it is weird. I think it would be better if each palette file were owned by exactly one package.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
```diff
@@ -147,11 +159,71 @@
...
+        <Additional_Destination_Index>0</Additional_Destination_Index>
+      </Destination_Overrides>
+      <Destination_Overrides>
+        <Path>Generator/Measurement UId.png</Path>
+        <Destination>5</Destination>
+        <Additional_Destination>false</Additional_Destination>
+        <Additional_Destination_Index>0</Additional_Destination_Index>
+      </Destination_Overrides>
+      <Destination_Overrides>
+        <Path>Palette Files/Measurement</Path>
```
jasonmreding, line 222: With this change, you now have two packages which are installing `Measurement Plug_In SDK.mnu`. I can't necessarily think of anything that will break because of this, but it is weird. I think it would be better if each palette file were owned by exactly one package.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
```diff
@@ -21,10 +21,10 @@
       <Additional_External_Dependencies>ni_measurement_plugin_sdk_service &gt;=3.0.0.1</Additional_External_Dependencies>
     </Package_Dependencies>
     <Custom_Action_VIs>
-      <Pre-Build_VI/>
-      <Post-Build_VI/>
+      <Pre-Build_VI>Build Custom Actions\Pre-Build Custom Action Generator.vi</Pre-Build_VI>
+      <Post-Build_VI>Build Custom Actions\Post-Build Custom Action.vi</Post-Build_VI>
       <Pre-Install_VI/>
-      <Post-Install_VI/>
+      <Post-Install_VI>Install Custom Actions\Post-Install Custom Action.vi</Post-Install_VI>
```
jasonmreding, line 27: This is going to reprocess all palette files that the `Measurement Plug-In SDK Service` package already processed. That probably won't break anything, but it is weird. Since you haven't added any new palette files as part of this PR, it is unclear if you will actually need this post install action for this package.

IIRC, this custom step is part of our effort to single source the palette files used for the functions palette and for the right click context palettes. I believe associating a palette file with the library allows the right click context menu to get updated appropriately with the owning palette of the item. However, we ran into some weird VIPM issues with copying these files around the source directory as part of our build, and this was the work around we came up with to make everything work.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
```diff
@@ -96,6 +96,18 @@
...
         <Path/>
       </QuickDrop_Plugins>
+      <Additional_Destination>
+        <Name>Palette Files</Name>
+        <LLB>false</LLB>
+        <Preserve_Hierarchy>true</Preserve_Hierarchy>
+        <Path>&lt;menus&gt;\Categories</Path>
+      </Additional_Destination>
+      <Additional_Destination>
+        <Name>Client Generator Runtime</Name>
```
jasonmreding, line 106: This directory doesn't actually exist in this PR. Will this not break the package build or does VIPM just silently ignore this?

Source/Build Specs/Build Custom Actions/Post-Build Custom Action.vi
jasonmreding, line 1: I don't recall if VIPM requires these custom build/install action VIs to be source only or not. I wouldn't think so since I believe it executes within the dev environment. Assuming that is the case, can you update all of the VIs in the "Build Specs" directory to be source only so they work correctly with the .lvversion file at the root of the repo. If that doesn't work, it might be nice to add a readme.md file explaining VIPM prevents us from making these files source only which then requires they be edited with the editor for the minimum LV version supported by the repo.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding: Maybe this PR should wait until #631 completes and the palette file changes for the new palette have been made before submitting?

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: Currently, to reflect any changes made to the new subpalette menu file `Client Generator.mnu` (which will be added to the `Measurement Plug-In SDK.mnu` palette), we need to build and install both the Service package and the Generator package. This is because the `Measurement Plug-In SDK.mnu` file is deployed by the Service package, while the `Client Generator.mnu` file is deployed by the Generator package.

With the new approach of deploying the `Measurement Plug-In SDK.mnu` palette file as part of both packages, we will decouple the Service and Generator packages. This change will remove the dependency on building and installing the Service package for any changes made to the palette files to take effect.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: We will be creating a new palette file for the wrapper VI, which will be added as a subpalette to the `Measurement Plug-In SDK` palette. The deployment of the newly created subpalette file will require pre-build, post-build, and post-install custom action VIs. The pre-build and post-build VIs are required to single-source the subpalette files of the `Measurement Plug-In SDK` palette. The post-install action VI is required to fix the palette files' paths mangled during package installation.

As mentioned in the HLD, we are trying to reuse the `Post-Install Custom Action.vi` and `Post-Build Custom Action.vi` of the Service package in the Generator package, as they provide the same functionality. The `Post-Install Custom Action.vi` will relink the menu files of all libraries deployed under the `<vi.lib>\Plug-In SDKs` directory, to correct the palette files' paths that get altered during package installation. Since this reprocessing does not affect the existing libraries or menu files, we thought it would be efficient to reuse the same VI for the post-install action of the Generator package.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: > This directory doesn't actually exist in this PR. Will this not break the package build or does VIPM just silently ignore this?

The addition of a new destination (deployment location) in the Generator Build Specification will not break anything, as we already have many other unused destinations predefined in the build specs.

> Maybe this PR should wait until https://github.com/ni/measurement-plugin-labview/pull/631 completes and the palette file changes for the new palette have been made before submitting?

The PR https://github.com/ni/measurement-plugin-labview/pull/631 is completed, and the addition and modification of palette files will be addressed in a separate PR, allowing us to proceed with the current changes, as they are non-disruptive.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding: I don't understand this comment. The generator package has an explicit package dependency on the SDK package. That means the generator package can't be installed via VIPM without first installing the SDK package. How does this decouple anything?

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding: Yeah, I just think it would be nice if the build/install of each package only processed files that are actually managed by that package. Presumably that means you could use the bulk of the existing build/install VIs. You just might have to create some wrappers around the core logic so you can customize what files are processed with each build. I haven't looked at the VIPM API for these hooks in detail. I don't know if you would be able to perform conditional logic with a single VI using inputs that are already passed to the VI, or if you would have to create a top level wrapper per package in order to customize the behavior.

Source/Palette Files/Measurement/Measurement Plug-In SDK.mnu
varshini-senthil: Until now, we only had the VIs from `Measurement Server.lvproj` deployed in the palette as part of the `Measurement Plug-In SDK Service` palette. The `Measurement Plug-In SDK.mnu` file was also deployed as part of the Service package, as only the APIs from that palette were made available.

However, we are now adding a VI from the Generator package to the palette. This means that any changes made to the generator subpalette file (such as renaming the file) will require updating the `Measurement Plug-In SDK` palette file. Consequently, although the change is made to the generator subpalette file, since the `Measurement Plug-In SDK` palette file is deployed by the Service package, this will necessitate building and installing both the Service package and the Generator package for the changes to take effect.

To avoid this, we are making the `Measurement Plug-In SDK.mnu` file common to both packages. This change will eliminate the need to build and install the Service package for any updates made to the generator subpalette file. Therefore, we are moving the `Palette Files` folder (which contains the palette file) up a directory in the source hierarchy.

Source/Build Specs/Build Custom Actions/Post-Build Custom Action.vi
varshini-senthil: I made the custom build and install action VIs source-only by enabling the "Separate compiled code from source file" checkbox in the VI properties. I was then able to build and install the Generator and Service packages successfully. Since VIPM allows source-only files, I will proceed with updating the custom build and install action VI files accordingly.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: After further discussion, we believe it is straight forward to follow the existing method where the service package installs the parent palette file 'Measurement Plug-In SDK.mnu', and the generator package installs the new sub-palette file. We realized that having a single file installed through two different packages might introduce unnecessary complexity.

Therefore, the `Measurement Plug-In SDK.mnu` palette file will be deployed by the Service package and this palette file will be modified to include a placeholder for the new sub-palette. The new sub-palette file for the Client Generator wrapper VI will be deployed by the Generator package as a post-install custom action in the required location. Thus, we need not relocate the `Measurement Plug-In SDK.mnu` menu file.

Our initial proposal aimed to streamline the update process for the parent palette file. However, upon reevaluation, we believe that maintaining the current approach will be more straightforward and less confusing. Thank you for your feedback, kindly let us know if you have alternative suggestions.

jasonmreding: OK, I was going to say that you should only need to modify the `Measurement Plug-In SDK.mnu` once to include the new child palette. After that, you should be able to deploy and make changes to the child palette as part of the generator package independent of the service package. There are a couple of ways you can set that up. One approach is to set up an auto populating palette based on directory hierarchy. However, most of the guidelines I can still find seem to discourage that approach. The second approach is to reserve a specific spot in the parent palette and then use the `readonly.txt` marker file to suppress broken icons for the child palette when it is missing / not installed.

varshini-senthil: Sure Jason. As mentioned, we shall follow the existing method for the deployment of palette files where the service package will install the parent palette file 'Measurement Plug-In SDK.mnu', and the generator package will install the new sub-palette file. Considering the guidelines, we will go with the second approach of using the `readonly.txt` marker file to set up things, as per your suggestion.

We will take up the changes required for this setup in a separate PR to ensure clarity and ease of following the decisions, as the purpose and approach of the new PR differs from the current one.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: I am resolving this comment as it no longer applies based on the conclusion from this discussion: https://github.com/ni/measurement-plugin-labview/pull/632#discussion_r1942037407.

Source/Palette Files/Measurement/Measurement Plug-In SDK.mnu
varshini-senthil: I am resolving this comment as it no longer applies based on the conclusion from this discussion: https://github.com/ni/measurement-plugin-labview/pull/632#discussion_r1942037407.

Source/Build Specs/Build Custom Actions/Post-Build Custom Action.vi
varshini-senthil: We will take up this change of making the custom build and install action VIs of the `Measurement Plug-In SDK Service.vipb` build specification as source only in a separate PR to ensure clarity and ease of following the decisions.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: I have created a new [PR](https://github.com/ni/measurement-plugin-labview/pull/633) for the required changes. Hence closing this PR.

Source/Build Specs/Build Custom Actions/Post-Build Custom Action.vi
varshini-senthil: I have created a new [PR](https://github.com/ni/measurement-plugin-labview/pull/635) for the required changes. Hence closing this PR.

varshini-senthil:
I have created new PRs for the updated changes based on the discussions. Hence closing this PR.
PR Links: https://github.com/ni/measurement-plugin-labview/pull/633, https://github.com/ni/measurement-plugin-labview/pull/635

PR #633, opened 2025-02-12
feat: Add sub-palette file for Client Generator wrapper VI and Update Generator build spec

varshini-senthil:

- Adds the new sub-palette file `Client Generator.mnu` to the `NI Measurement V2 Client Generator Runtime.lvlib` library. This sub-palette file will hold the `Generate Client.vi` wrapper VI
- Updates the existing parent palette file `Measurement Plug-In SDK.mnu` to create a place holder for the new sub-palette file.
- Updates the `NI Measurement V2 Client Generator Runtime.lvlib` library to set the `Client Generator.mnu` file as the default palette file for the library.
- Adds custom post install and pre uninstall action VIs required for the `Measurement Plug-In SDK Generator` build spec. These action VIs are essential for the deployment of the `Client Generator.mnu` sub-palette menu file in the expected location.
- Updates the `Measurement Plug-In SDK Generator.vipb` to align with latest changes made to the source code.

- To add the new sub-palette menu file and to update the existing parent palette file `Measurement Plug-In SDK.mnu` file.
- To update the `Measurement Plug-In SDK Generator.vipb` to align with latest changes made to the source code and to add the custom action VIs required for the Generator package build.

Built the Service and Generator package with the changes and tested the packages to ensure the palette files are deployed and work as expected.


Prem-MysteryPro:
@varshini-senthil 
I hope we are planning to raise another PR to enable the `separate compiled code from source` option for the existing build or install action VI's.

varshini-senthil:
> @varshini-senthil I hope we are planning to raise another PR to enable the `separate compiled code from source` option for the existing build or install action VI's.

Yes @Prem-MysteryPro, I have raised the changes as a separate PR.

Prem-MysteryPro:
In the testing I hope we covered the scenario of installing the service package alone and made sure that plugins are working, and service-related palette are loading appropriately. Similarly, what will happen if I install older version of service package already and install latest version of generator package.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
Prem-MysteryPro, line 1: Should we bump up the dependency of the LV grpc and service package in dependency?

varshini-senthil:
> In the testing I hope we covered the scenario of installing the service package alone and made sure that plugins are working, and service-related palette are loading appropriately. Similarly, what will happen if I install older version of service package already and install latest version of generator package.

Yes, I tried with certain combinations of installation of the Service and Generator packages to see how the palette files will be affected. In the cases you have mentioned where we will install:
1. Only the service package - The `Measurement Plug-In SDK` palette will be visible along with its sub-palettes and the Service-related APIs as it is now (i.e., will be unaffected). 
_Note:_ The place holder created in the `Measurement Plug-In SDK` palette for the Client Generator sub-palette will not be visible from the Block diagram as the question mark for the missing palette will be hidden by the `readonly.txt` file. But, the question mark will become visible in the palette editing mode.

2. Older version of Service and latest version of Generator - In this case, the `Measurement Plug-In SDK` palette will not have the placeholder created for the Client Generator sub-palette file, as the palette file is not updated. So, the installation of latest Generator package will deploy the APIs and the Client Generator sub-palette file. But the sub-palette will not become available under LabVIEW's Functions palette.

varshini-senthil:
Image of the updated `Measurement Plug-In SDK` palette with the new `Client Generator` sub-palette:

image

Palette documentation:

image

image

Source/Build Specs/Install Custom Actions/Pre-Uninstall Custom Action - Generator.vi
Prem-MysteryPro, line 1: Are we doing this since the menu file is not removed by generator vip uninstallation automatically? If so what about the `Measurement V2 Client Generator Runtime` folder is it required?

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
Prem-MysteryPro, line 1: Can we clean up the free label for documentation and partially hidden constants and wires?

Prem-MysteryPro: Should we also rename to Post-Install Custom Action - Client Generator.vi?

Prem-MysteryPro:
> > In the testing I hope we covered the scenario of installing the service package alone and made sure that plugins are working, and service-related palette are loading appropriately. Similarly, what will happen if I install older version of service package already and install latest version of generator package.
> 
> Yes, I tried with certain combinations of installation of the Service and Generator packages to see how the palette files will be affected. In the cases you have mentioned where we will install:
> 
> 1. Only the service package - The `Measurement Plug-In SDK` palette will be visible along with its sub-palettes and the Service-related APIs as it is now (i.e., will be unaffected).
>    _Note:_ The place holder created in the `Measurement Plug-In SDK` palette for the Client Generator sub-palette will not be visible from the Block diagram as the question mark for the missing palette will be hidden by the `readonly.txt` file. But, the question mark will become visible in the palette editing mode.
> 2. Older version of Service and latest version of Generator - In this case, the `Measurement Plug-In SDK` palette will not have the placeholder created for the Client Generator sub-palette file, as the palette file is not updated. So, the installation of latest Generator package will deploy the APIs and the Client Generator sub-palette file. But the sub-palette will not become available under LabVIEW's Functions palette.

Then I think we might need to update the service dependency version to be latest. [Related comment](https://github.com/ni/measurement-plugin-labview/pull/633#discussion_r1954171662).
Let us request suggestion from Jason.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
varshini-senthil: > Can we clean up the free label for documentation and partially hidden constants and wires?

I have cleaned up the VI to remove any hidden constants and wires and the free label documentation of the VI as suggested.

varshini-senthil: > Should we also rename to Post-Install Custom Action - Client Generator.vi?

I have added the '- Generator' tag to the VI name to differentiate this VI from the 'Post-Install Custom Action.vi' of the `Measurement Plug-In SDK Service.vipb`. Since, this custom action VI is used by the SDK Generator VIPB as a whole, should we change the tag to '- Client Generator' to make it more specific for the Client Generator?

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: I agree with your ask to update the Service package version mentioned under the `Package Dependencies` section of the SDK Generator build spec as we require the installation of the updated SDK Service package in user's system for the `Measurement Plug-In SDK` palette file to get updated. As mentioned in this [comment](https://github.com/ni/measurement-plugin-labview/pull/633#issuecomment-2656192664), this is essential for the `Client Generator` sub-palette to become available in the `Measurement Plug-In SDK` palette.

@jasonmreding, please share your thoughts on this.

jasonmreding: Yes, I would suggest updating the minimum required version to the latest. While it's not technically necessary from a functional perspective, it probably leads to the least questions for the end user.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
jasonmreding: These actions are specific to a given vipb so I think the names should indicate the package and not a specific library within the package. In other words, I think the current name is good. If anything, you could rename the existing VI to have a " - Service" suffix.

Source/Build Specs/Install Custom Actions/Pre-Uninstall Custom Action - Generator.vi
jasonmreding, line 1: Consider either updating the code to handle multiple .mnu files or generating an error if more than one .mnu file is found. This goes for both the pre/post actions. I don't feel strongly about this so you are welcome to defer at your discretion.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding, line 1: I have mixed feelings in regards to deployment of the template libraries. On one hand, what you have done here is consistent with what existed previously. However:
- The fact that the service generator deploys anything under <vilib> is arguably incorrect.
- The fact that we are still deploying to `vi.lib\Plug-Ins\Measurement` which uses our old directory structure is odd. Everything else is deployed under `vi.lib\Plug-In SDKs` including the new client generator library. If we deployed everything under the same vilib root folder and keep the relative pathing the same as our source folder, then we could actually modify `Template Files Path.vi` so it always just returns a relative path and remove the conditional compile symbol and conditional disable structure.
- If we changed deployment of service template libraries to tools directory instead of vilib, we could also make similar changes for specifying pathing as previous bullet.

Given the above, I think it would make more sense to deploy the templates to the same directory as the runtime library. If you wanted to also do some cleanup around deployment of the service generator templates, I think that would good as well. However, you didn't make that mess so I won't require you to clean it up as part of this PR.

jasonmreding (changes requested):
Waiting to see next round of changes before approving.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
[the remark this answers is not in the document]
varshini-senthil: I moved the reusable code in the `Post-Install Custom Action.vi` of the Service and Generator packages into a sub VI and updated the Post-Install VIs to use this sub VI. After which I built the SDK Generator package, the build was successful, but the package produced an error during installation.

Error message:
Error in Package Installation

One more observation is that when I replaced the sub VI with its contents and built the package, both the build and installation were successful. So, I think Custom Action VIs used during package install and uninstall actions are expected to be simple VIs that do not call sub VIs.

Please share your thoughts on this and let me know if I am missing something in my approach.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil, line 1: When looking into the help for the VI Package Builder, I found a list detailing the requirements for custom action VIs:

image

As the highlighted line states, the VIs used for the install and uninstall actions of any package are expected to be part of the corresponding package source folder (i.e., Package's Source Directory). Currently, the Source Directory of the `Measurement Plug-In SDK Generator` package is set to `Source\Generator` directory, which makes the install and uninstall action VIs of the package external to its source as the VIs are located under `Source\Build Specs\Install Custom Actions` directory. So, should we consider either:
1. Updating the Source Directory of the `Measurement Plug-In SDK Generator.vipb` from `Source\Generator` to `Source` (or)
2. Creating a new folder named 'Install Custom Actions' under the `Source\Generator` directory and move the install and uninstall custom action VIs of the Generator package to this new folder

to ensure the install and uninstall custom action VIs are part of the package source folder? Please share your thoughts on this.
**_Note:_** The current location of the install and uninstall custom action VIs does not affect the SDK Generator package's build, installation or its functionality.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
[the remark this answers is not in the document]
jasonmreding: Sigh... given what I see from [this post](https://forums.vipm.io/topic/2482-subvis-in-prepost-install-vis/), I suspect VIPM isn't capable of handling a VI hierarchy for these extension points unless you package the sub VIs into their own package which seems crazy. I'm guessing VIPM copies the specified VIs to a different location on disk (probably temp folder) and doesn't do anything to try and find/copy dependencies of that VI.

The only other thing I can think of is to put everything into a single VI that is shared by both packages, and then use some input to recognize which package is being operated on and perform conditional logic based on that. If you use VIPM to generate a template VI for you, this is what it generates.

image

Given this, it seems like you might be able to use the "Package Name" from the variant that gets wired to a case structure to perform conditional package setup. So in effect, each case of that case structure effectively becomes a package specific sub VI and the rest of the diagram is effectively what would have been the shared sub VI. If none of that seems viable or better, then I can live with the code duplication.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding: If you want the ability to have shared installation VIs between the various packages, then you don't really have any choice other than to declare the source directory as the root source directory for each package and then exclude everything that doesn't apply. I wish the VIPM editor provided a better overview of what is and isn't included in the package.

Source/Build Specs/Install Custom Actions/Pre-Uninstall Custom Action - Generator.vi
varshini-senthil: I have updated the logic to handle multiple .mnu files as recommended.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: I have updated the 'Source Directory' of the Generator build specification from `Source\Generator` to the root source directory of the repository as we share the custom installation VIs between the Generator and Service package build specifications, as suggested.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action - Generator.vi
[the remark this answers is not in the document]
varshini-senthil: I have updated the `Post-Install Custom Action.vi` to use the variant input from VIPM to differentiate between the Service and Generator packages and perform the required post-install operations accordingly, as suggested.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: I have updated the deployment location of the Client template libraries from `vi.lib\Plug-Ins\Measurement` to `vi.lib\Plug-In SDKs` and the `Template Files Path.vi` to align with this change in deployment location. Additionally, I have removed the conditional disable symbol created in the `Measurement Generator Scripting.lvproj` as we will be using relative pathing (i.e., Current VI's path) in both source and deployed location to construct the path to the template libraries.

varshini-senthil: Sure Jason,

Given that we need to update the `Measurement Plug-In SDK Service` package for the upcoming release, which will, in turn, require an update of the minimum required version of the Service package specified in its dependent package build specs:

- For now, I shall update the minimum required version of the `Measurement Plug-In SDK Service` package specified in the `Measurement Plug-In SDK Generator` build spec to 3.2.0.1 (which is the latest available version).
- However, we will need to update the minimum required version once again to the new version (which could be 3.3.0.1) during the upcoming release (i.e., after the build of the Service package and before the build of the Generator package during the package build process for the release).

Please share your thoughts on this.

_Note:_ The minimum required version specified for the SDK Service package should be updated again to the latest version to ensure the update of the `Measurement Plug-In SDK Service` package installed on the user's system to the latest version when they install the upcoming SDK Generator package. This update is necessary because only then the `Measurement Plug-In SDK.mnu` file will be updated, making the Client Generator sub-palette available in the Functions palette.

varshini-senthil: Sure Jason,
- For now, I will update the minimum required version of the `Measurement Plug-In SDK Service` package mentioned in the 'Package Dependencies' section of the `Measurement Plug-In SDK Generator` build spec to 3.2.0.1, which is the latest available version of the Service package.
- Since we will be updating the `Measurement Plug-In SDK Service` package in the next release, we will also need to update the minimum required version of the Service package mentioned in the build specifications of the dependent packages (such as the Generator package) once again, to ensure the packages require this latest version of Service package. This update must be done before building those dependent packages as part of the next release work.

**Reasons for the update:**
- To ensure users install the latest Service package as an auto-dependency when they install the updated Generator package that will be provided in the next release.
- The installation of the latest Service package is necessary to update the `Measurement Plug-In SDK.mnu` palette file. This update is essential for the Client Generator sub-palette to appear in LabVIEW's Functions palette.

Please share your thoughts on this.

varshini-senthil:
I have updated the approach for deploying the Generator package menu files to follow the method used by the Service package. In this approach, the menu files are single-sourced before the package build through a custom pre-build action and then deployed to the respective location during installation. As part of this change, I have made the following updates:

- **Removed the `Pre-Uninstall Custom Action.vi`**, as it is no longer required for the new approach.
- **Updated the `Post-Install Custom Action.vi`** to remove the logic for deploying copies of the Generator package palette files. Only the logic essential for relinking the palette files to the libraries has been retained.
- **Updated the `Pre-Build Custom Action.vi`** to identify the package based on the `Product Name` obtained from the variant input provided by VIPM. The corresponding package's palette files are then copied to the `Source\Runtime\Palette Files\Measurement\_Measurement Plug-In SDK` folder to single-source the files.
- **Updated the Generator build specification** to deploy the palette files copied to the `Source\Runtime\Palette Files\Measurement\_Measurement Plug-In SDK` folder to the `<LabVIEW>\menus\Categories\Measurement\_Measurement Plug-In SDK` folder.

**Note:** This change in approach is to avoid the programmatic creation of the _Measurement Plug-In SDK folder and the copying of the palette files as post-install actions. This was necessary because, in some cases, the `<LabVIEW>\menus\Categories\Measurement\_Measurement Plug-In SDK` folder, created programmatically by the Generator package during installation, was retained on disk even after uninstalling both the Generator and Service packages.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding, line 1: Is there a reason you aren't excluding the entire "Runtime" folder? It looks like you are only including the palette files directory under it, but I don't understand why. That directory should only include the palette files managed by the other service package, right?

image

jasonmreding, line 1: It might be less confusing to just make the `Toolkit VIs` path (vilib path) the same as the `Client Generator Runtime` and possibly deleting the `Client Generator Runtime` in favor of the `Toolkit VIs` destination if that seems simpler to you.

jasonmreding: Yes, you will need to update the version of both packages and update the generator package so its minimum version for the dependency on the service package is the latest version. You can either do that as part of this PR or in a small follow on PR after this completes. I don't think there is any reason why this has to wait until release to do.

Source/Build Specs/Install Custom Actions/Post-Install Custom Action.vi
jasonmreding, line 1: If you have both the service and generator package installed and then upgrade the version of the service package, you are going to end up processing the .mnu files for the generator package.

image

This is probably harmless, and I'm not sure how much effort you want to throw at preventing this.  If you wanted to fix it, I think you would probably want to create a sub VI that returns just the generator .mnu paths and then exclude them from whatever is calculated here. If the deployed menu files for the generator package were in their own sub-directory, that might simplify things a bit. Again, I'm not sure how much effort you want to throw at this. I can live with what you have here.

jasonmreding, line 1: From you previous post:

413426473-133c0457-cc2a-4d30-8d98-453a69b99515

my assumption was this would also allow for the custom build/install VIs to contain sub VIs as long as they were also part of the source directory. Is that not that case? I'm OK with what you have here, but it might be a little cleaner long term to utilize sub VIs rather than having to utilize conditional code based on the package name and effectively have all of the logic in one big VI.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: - The `Measurement Plug-In SDK Service` package uses the `Measurement\_Measurement Plug-In SDK` folder inside the `Palette Files` folder as a placeholder to single source the palette files deployed by the package.

**Method followed by the Service package for palette file deployment:**
-  The pre-build custom action of the Service package creates a copy of all the menu files to be deployed by the package in the `Runtime\Palette Files\Measurement\_Measurement Plug-In SDK` folder, before the start of the package's build process.
- Also, the deployment location of the `Measurement\_Measurement Plug-In SDK` folder is configured to `<LabVIEW>\menus\Categories` folder in it's build spec.
- The post-build custom action of the package deletes this copy of the package files created in the placeholder folder (`Runtime\Palette Files\Measurement\_Measurement Plug-In SDK`) after completion of the package build process.
- This method allows the Service package to deploy the actual palette files to the `vi.lib\Plug-In SDKs` folder and a copy of the files to the `<LabVIEW>\menus\Categories\Measurement\_Measurement Plug-In SDK` folder, while only one palette file actually exists in the repository's source.
-  All the menu files deployed in this method are the files of the sub-palettes present in the `Measurement Plug-In SDK` palette.

> Is there a reason you aren't excluding the entire "Runtime" folder? It looks like you are only including the palette files directory under it, but I don't understand why. That directory should only include the palette files managed by the other service package, right?

- The palette file to be deployed by the Generator package is also the file of a sub-palette that will be made available under the same `Measurement Plug-In SDK` palette. So, a copy of this sub-palette file should also be deployed to the `<LabVIEW>\menus\Categories\Measurement\_Measurement Plug-In SDK` folder just like the Service package's palette files.
- So, I have configured the deployment location of the `Measurement` folder inside the `Palette Files` folder to `<LabVIEW>\menus\Categories` folder in the Generator build specification to follow the method used by the Service package for the deployment of its palette files.
image
image

Source/Build Specs/Install Custom Actions/Post-Install Custom Action.vi
varshini-senthil: I have updated the logic to exclude the `Measurement V2 Client Generator Runtime` folder which contains the Generator mnu files.
image

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
varshini-senthil: The `Toolkit VIs` destination is still required to configure the deployment location of the `Measurement Plug-In Template` files. This was the primary reason behind creating a new destination for the `Client Generator Runtime` files even when the Base Path `<LabVIEW\vi.lib` was same for both the destinations.

image

varshini-senthil: Sure Jason, I assumed that the package version update changes will be done as part of the release work. If it could be done now, I shall make the required changes after completion of this PR. Considering that the changes made to the source code after the release of the latest `3.2.0.1` version are more about new features and improvements, shall I update the version of the Service and Generator packages from `3.2.0.1` to `3.3.0.1`?

Source/Build Specs/Install Custom Actions/Post-Install Custom Action.vi
jasonmreding: OK, it wasn't clear to me if the sub VI linkage was broken because it will never work if not deployed as part of the package or if it was because it was outside of the source directory. Presumably VIPM must be doing something special to store off those individual VIs for later as we explicitly exclude the `Build Specs` folder from the deployment. I guess they just don't try to analyze the VI dependencies of that VI hierarchy and automatically pull any from under the same source folder with it as part of the build/packaging process. Anyway, I just wanted to verify this so resolving this thread.

Source/Build Specs/Measurement Plug-In SDK Generator.vipb
jasonmreding: Oh right, I forgot the server templates are part of this package as well and are deployed to vilib even though they aren't needed at runtime. If we ever clean that up and deploy to tools directory then this wouldn't be necessary.

jasonmreding: Yes, I would consider this a minor feature upgrade so 3.3.0.1 seems appropriate to me.

jasonmreding: I think I made this comment before I realized you changed to use the same pre/post build steps as the service package so this makes more sense. However, it does mean the top level palette/.mnu file is installed by both packages, right. I'm concerned that is going to cause weird upgrade/downgrade or install/uninstall issues. For instance, if you just uninstall the generator package, does it remove that .mnu file even though the service package is still installed. Or is VIPM smart enough that it does ref counting somehow and manages this?

varshini-senthil: No Jason, the top-level palette file will be installed only by the Service package. The Generator package will only deploy the `Client Generator.mnu` sub-palette file to the `menus\Categories\Measurement\_Measurement Plug-In SDK` location as the top-level palette file is excluded from the Generator package.

image

So, uninstalling the Generator package will remove only the sub-palette file it deployed from the corresponding location, leaving the top-level palette file unaffected.

PR #644, opened 2025-03-27
Fix Measurement Plugin vipb package version in main

Blahpapto1:

Measurement Vipb version update for SDK and SDK Example package to 3.3.0.1 to match the Generator and Service package.

Fix version of Measurement PlugIn vipb package before creating the new release branch.

N/A

### DIFF
--- a/Source/Build Specs/Measurement Plug-In SDK Examples.vipb
+++ b/Source/Build Specs/Measurement Plug-In SDK Examples.vipb
@@ -1,7 +1,7 @@
-<VI_Package_Builder_Settings Version="2020.1" Created_Date="2024-03-06 15:30:02" Modified_Date="2024-09-24 11:15:02" Creator="ksreekan" Comments="" ID="c05770638f35c26ca8d8ee10413ed162">
+<VI_Package_Builder_Settings Version="2020.1" Created_Date="2024-03-06 15:30:02" Modified_Date="2025-03-27 03:17:30" Creator="ksreekan" Comments="" ID="8a3f8cb3756b342419a791d5dfd64e0b">
   <Library_General_Settings>
     <Package_File_Name>ni_measurement_plugin_sdk_examples</Package_File_Name>
-    <Library_Version>3.2.0.1</Library_Version>
+    <Library_Version>3.3.0.1</Library_Version>
     <Auto_Increment_Version>false</Auto_Increment_Version>
     <Library_Source_Folder>..\Example Measurements</Library_Source_Folder>
     <Library_Output_Folder>..\..\Build Output</Library_Output_Folder>
--- a/Source/Build Specs/Measurement Plug-In SDK.vipb
+++ b/Source/Build Specs/Measurement Plug-In SDK.vipb
@@ -1,7 +1,7 @@
-<VI_Package_Builder_Settings Version="2020.1" Created_Date="2024-03-06 15:09:40" Modified_Date="2024-09-24 11:15:18" Creator="ksreekan" Comments="" ID="6ebdd931bcb35629bbf67d4bba1b02cf">
+<VI_Package_Builder_Settings Version="2020.1" Created_Date="2024-03-06 15:09:40" Modified_Date="2025-03-27 03:17:55" Creator="ksreekan" Comments="" ID="f8fff25bcfb46515b7c60b79de9eb80c">
   <Library_General_Settings>
     <Package_File_Name>ni_measurement_plugin_sdk</Package_File_Name>
-    <Library_Version>3.2.0.1</Library_Version>
+    <Library_Version>3.3.0.1</Library_Version>
     <Auto_Increment_Version>false</Auto_Increment_Version>
     <Library_Source_Folder>..</Library_Source_Folder>
     <Library_Output_Folder>..\..\Build Output</Library_Output_Folder>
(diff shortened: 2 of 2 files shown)


Source/Build Specs/Measurement Plug-In SDK.vipb
jasonmreding, line 1: I just noticed this and some of the other .vipb files don't have the `.lvversion` file excluded. Is this file actually getting installed by these packages?

Source/Build Specs/Measurement Plug-In SDK Examples.vipb
jasonmreding, line 1: I guess I'm always a bit torn by these sorts of changes. None of our examples have changed since the last release, and we are not changing any of the dependencies or minimum version requirements of our dependencies. So why are we creating a new release of this package which is published to VIPM? The same goes for the `Measurement Plug-In SDK` package which is just a meta package which exists to simplify installation so users can pick a single top level package to install which will pull in all of the required dependencies (from NIPM perspective it would be the equivalent of a store product). Is there some practical/functional motivation for making this change or was it just part of some release checklist? I guess I can see wanting all of the version numbers to match for a given release, but we already have the ni.protobuf.types package in the repo which won't match these versions.

I guess I can see a couple of strategies here:
1. Always bump all package versions so they match and release all packages each release.
   - With this approach, perhaps we should also bump the minimum version of dependencies to match the new version? This would mean the metadata package would require upgrade of all dependent packages when upgraded. However, it would still allow individual dependencies to be upgraded beyond its current version unless we locked it down to an exact version match. 
3. Only bump versions and rebuild/release packages with changes.
   - With this approach, we would never release a new version of the meta package unless we changed the number of dependencies or wanted to enforce a different minimum version of dependencies. VIPM would simply show available upgrades of the dependency packages without requiring an upgrade of the meta package itself.

@dixonjoel any thoughts on this? I'm sure there are other variations of these strategies we could utilize as well.

dixonjoel: Setting aside the minimum dependency question, I like the clarity and simplicity of just bumping the versions to be the same each release and releasing whether we have made changes or not. This avoids somewhat the need to carefully keep track of what package versions are compatible with each other. Typical expectation would be just to download and install all four packages with the same version from a new release.

So basically your option 1.

Protobuf types package will be different yes, but that may eventually go to its own repo anyway. We will always include a compatible package of this with each release.

On to version dependencies. In general, I think we should only bump the minimum dependent version when needed, e.g. if the generator needs new service package code added in the release, we'd bump it's minimum service version to the newest one. This way, if the user happens to only download the new generator package, it would require them to get a new service package.

Not sure if that answers all the questions, but for me it means this PR is good and the release will include all 3.3.x pacakges.

Source/Build Specs/Measurement Plug-In SDK.vipb
dixonjoel: @Blahpapto1 please exclude the .lvversion file from the SDK package like this and test to ensure it's not installed.
image

Agree @jasonmreding ?

dixonjoel (approved):

Source/Build Specs/Measurement Plug-In SDK Examples.vipb
jasonmreding: I posted a question to our internal LV users forum and results are mixed. The things that were pointed out were:
1. [OpenG](https://www.vipm.io/package/openg.org_lib_openg_toolkit/) has a similar meta package (but much larger). It has very few releases and appears to always just pull the latest version of the dependency.
1. What happens if the user wants to downgrade the meta package and have all of the packages move as one? With our current approach (and OpenG's), this doesn't work. An upgrade or downgrade of the meta package only changes that package because we typically never change the oldest compatible version. If we changed our approach to always specify exact dependency versions from the meta package then user could upgrade/downgrade that package to control all versions collectively (which were presumably all tested against each other for that release). I haven't tested this, but I believe requiring an exact version from the meta package means any time a new version of a dependency is released, then a new version of the meta package must also be released. Either that or the user would have to first uninstall the meta package so they could then install the packages individually in order to get fine grained control of the package versions being used.

I personally kind of like the second option even though it doesn't match OpenG. I also tend to prefer not re-releasing packages that haven't changed, but I'm not willing to die on that hill. For the second option, there is also a nuance of whether we use >= or = for the version of that release. Using <= would allow for bug fixes / upgrades to a single dependency without having to re-release the meta package. However, I'm not sure how compelling that is vs. always having to release a new version of the meta package as well.

Source/Build Specs/Measurement Plug-In SDK.vipb
jasonmreding: > Either way do we want it installed or no? 

No.

> What are the implications of this?

I'm not entirely sure. I would have to go reread some of the LV help to understand the algorithm used again. I was afraid we were installing to the root of vilib or root folder we install to inside vilib. In that case, my concern is that when open those vilib VIs the file would cause LV to try and keep the loaded version at that version in the editor, which would generally be bad. I would rather exclude the file and not have to think about it. Since the package was already being modified, this seemed like a good time to fix it.

Blahpapto1: > @Blahpapto1 please exclude the .lvversion file from the SDK package like this and test to ensure it's not installed.

Will test this change. Trying to setup my machine for building and installing these packages but I'm currently getting some error while installing Labview 2021.

image

dixonjoel: If you're running an installer directly off \\argo, you may try downloading the installer locally first.

Source/Build Specs/Measurement Plug-In SDK Examples.vipb
dixonjoel: For the sake of getting this release out sooner, I suggest we leave things as they are and for this release, we make all the versions the same and do not change any dependencies. We should still prevent the install of the .lvversion file as mentioned in the other thread.

jasonmreding: I am OK with that for now. We can look at some of the other improvements in the background moving forward.

Source/Build Specs/Measurement Plug-In SDK.vipb
Blahpapto1: Updated SDK vipb package to exclude `.lvversion`. Also, tested it out on installing it does not add the `.lvversion` file in that location.

PR #648, opened 2025-04-28
Adding support for 2D arrays of double and string

jasonmreding:

- Adds support for 2D arrays of double and strings for LV based measurements. 
- Updates scope example to utilize 2D array of doubles rather than N 1D arrays. The example has also been updated to demonstrate proper cancellation since the example uses a trigger, and the user might want to abort waiting on the trigger.
- Updates the template UI demonstrating supported data types.

- Creates parity between LV and Python of supported data types.
- Adds an example which demonstrates proper cancellation support.

- Added some automated tests for valid/invalid array types.
- Hand tested scope example to verify it behaves as expected.


Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Wait Until Cancelled or Complete.vi
jasonmreding, line 1: I wrestled quite a bit with how we should exemplify and recommend users to implement cancellation. I ended up with an implementation that uses a timed loop and channel wires as that seemed to create the simplest diagrams. However, I recognize these are two language constructs that probably aren't used that often so I'm open to feedback on this.

Other options I considered were an event structure with user event, occurrences, and queues. All of them would require some form of polling without modifications to grpc-labview to expose an API where the user could register a user event which would get fired when cancellation is requested. This coupled with a second user event for when the measurement completes normally would allow for a solution that is completely event driven without any need for polling. However, I decided not to pursue that route until getting additional feedback about what seemed best.

jasonmreding: I converted the timed loop to a while loop since I realized the channel wire is using an occurrence under the hood anyway. That simplifies things a bit more and the diagram is less cluttered.

searly: Looks better without the timed loop. Would using the Tag channel type instead of Stream make sense? Since this is an example I would add brief VI descriptions to the two new helpers, everything else on the MeasurementLogic.vi diagram has context help.

Source/Runtime/Measurements/Helpers/LabVIEW Data Type to Measurement Data Type.vi
searly, line 1: I'm curious why the 1D array case doesn't need to be filled in like the 2D case.

jasonmreding: gRPC inherently supports repeated data (or 1D arrays), so at this level the only thing we really care about is flagging the data type as repeated. The caller then calls this same VI again with the array element type to see if that scalar type is supported. Currently, if we support the scalar type, we also support the repeated or 1D version of that type. This implicit coupling in the logic isn't the best when viewed from just the perspective of this VI, but it was how the code started. If you have any thoughts for how to improve it, I can take a crack at refactoring it.

searly: oh I see, makes sense. I guess this could just recursively call itself instead of having the caller know to call it again.

searly (approved):

Source/Generator/Measurement Plug-In Template/UI/Measurement UI.vi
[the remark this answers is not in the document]
jasonmreding: Updated.

Source/Runtime/Measurements/Helpers/LabVIEW Data Type to Measurement Data Type.vi
jasonmreding: OK, the diffs will probably be terrible, but I've updated the logic so it recursively calls itself and removed some unused indicators. After the refactor, I collapsed/deleted a sub VI layer since the refactor resulted in a single VI call on the diagram with all of the controls/indicators wired to the single remaining sub VI call.

Source/Example Measurements/NI-SCOPE Acquire Waveform/NI-SCOPE Acquire Waveform/Wait Until Cancelled or Complete.vi
jasonmreding: I added documentation to the helper VIs. 

Use of a tag vs. streaming vs. messenger channel is subjective I suppose. I personally prefer destructive data reads with what I would term as commands being sent across the channel. You only get that with streaming channels. I also view this as strict peer to peer communication and you really only get that with streaming channels as well since you are not allowed to branch the wire. Based on that I am resolving this comment thread. If you feel strongly about a different approach, reopen the thread and we'll discuss it further.

PR #651, opened 2025-05-12
Updating measurement template to better support parallelism and cancellation

jasonmreding:
### Description

Updates the measurement template used by the measurement generator to better support parallelism and cancellation.

### Implementation

- Added cancellation VIs to the template project and updated TODO documentation for the template to explain the purpose for these VIs.
- `Measurement Logic.vi` and new cancellation VIs support reentrant execution by default so measurement calls can run in parallel.
- Audited existing examples and updated their reentrancy settings so they are all reentrant and support parallel calls.

### Testing

Generated a new measurement with the updated template/generator and verified the default implementation of the template example works as expected such that multiple measurement calls are executed in parallel.


Source/Generator/Measurement Plug-In Template/Measurement/Wait Until Cancelled or Complete.vi
dixonjoel, line 1: image
The 1ms default interval for polling seems pretty low to me. Have you checked the CPU while polling for cancel? Maybe, e.g. 10 ms is better?

dixonjoel: As an aside, I think this is the first time I've ever seen the 'pipe / channel' used in LV.

dixonjoel (approved):
A couple minor suggestions

Source/Example Measurements/Game Of Life/Game Of Life/Helpers/Next Generation.vi
searly, line 1: Needs front panel cleanup. "births" and "deaths" are on the wrong side.

searly (approved):

Source/Generator/Measurement Plug-In Template/Measurement/Wait Until Cancelled or Complete.vi
jasonmreding: > The 1ms default interval for polling seems pretty low to me...

The polling interval impacts both the checking of cancellation and normal completion. So in effect, the polling interval is the minimum amount of time it will take a `Measure` call to complete. That is why I originally set it so low. With the latest push, I've updated to using an event based channel wire in order to signal normal completion, and set the default polling interval for cancellation to 100 ms.

PR #652, opened 2025-05-16
Adding automated tests to verify measurement calls execute in parallel or in sequence based on reentrancy

jasonmreding:

### Justification

We do not currently have any automated tests which exercise the entire software stack for our measurement server. This PR adds tests which perform end to end tests from client to server in order to validate multiple `Measure` calls are scheduled for execution serially or in parallel as appropriate based on their reentrancy settings.

### Implementation

- Added new acceptance test folder and library containing measurement service tests.
- Added new test measurement service which continually counts upward by one until stopped.
- Added new test measurement client library to simplify calls to above measurement.
   - Added retry loop with timeout when:
      - Resolving through the discovery service. There is an inherent race condition when starting up the client and server in parallel as to which will happen first so we need to tolerate not being able to resolve the service for some period of time as the server is initialized. 
      - As part of "warmup" VI which should be the first client call with each test. Again, there are inherent race conditions between initialization of the server and the client where the client might receive `UNAVAILABLE` gRPC status codes until the server is sufficiently initialized to start accepting client requests. This warmup loop ignores these errors and retries until no error is received or the specified warmup timeout expires.
- Added tests that:
   - Verifies multiple measure calls are able to run in parallel if the `Measurement Logic.vi` for the service is marked as reentrant. 
   - Verifies multiple measure calls run sequentially if the `Measurement Logic.vi` for the service is marked as non-reentrant. This cannot be verified without a certain amount of uncertainty. However, it at least verifies multiple parallel calls from the client execute without error which cannot happen unless the server dispatches the calls in a serialized fashion.
- Renamed "Unit Tests" folder to "Unit" for consistency.

### Testing

New tests pass on the PR pipeline.


Source/Tests/Tests.Runtime/Measurement Server/MeasurementServerTests.lvlib
searly, line 1: I would use a Select node over a case structure here.

searly: This is regarding Assert True.vi

Source/Tests/Tests.Runtime/Measurement Server/Acceptance/Measurement Service Tests/Continuous Counting Measurement/Get Plugin Paths.vi
searly, line 1: I guess you can't just set the reentrancy setting on the logic VI here instead of having the two wrappers? Does it get purged from memory and reloaded?

jasonmreding: I didn't think it would let me set reentrancy while reserved, but it seems to work OK. I always find it a bit weird to invoke a VI marked as reentrant as non-reentrant, but LV allows it (but not the other way around). Anyway, I've updated to setting the reentrancy programmatically and deleted a number of support VIs that are no longer needed with that strategy.

PR #657, opened 2025-07-08
feat: Add support for Measurement Remote Execution by populating deployment_target

AishwaryaLakshmiSS:

This Pull Request implements support for the [Feature 3069888](https://dev.azure.com/ni/DevCentral/_workitems/edit/3069888): Execute measurement plug-ins using remote instruments with Session Management​.
It enables accessing instrument resources hosted on a remote test bench machine from developer machines. To achieve this, the `Enumerate Compute Nodes` API that returns the remote test bench URL as a compute node is implemented in the `DiscoveryClient` and integrated into the service clients:

- gRPC Device Server Activation service
- Session Management service
- Pin map service

These clients use the `Enumerate Compute Nodes` API to detect remote compute nodes and populate the `deployment_target` accordingly. As a result, all service requests are resolved and sent to the appropriate remote services instead of the local host. If no remote compute nodes are found, the system falls back to the current local behavior.

This change is essential for enabling seamless execution of measurement plug-ins using remote instrument resources. By allowing service clients to resolve and communicate with remote services transparently, developers can work remotely without changing existing workflows.

TODO: Add automation tests.

### Additional references:

- [Pull Request 948366](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/948366): HLD to execute measurement plug-ins using remote instruments with Session Management​
- [Pull Request 1004993](https://dev.azure.com/ni/DevCentral/_git/ASW/pullrequest/1004993): Service clients: Populate the respective clients with the deployment target


Source/Tests/Tests.Runtime/Measurement Server/Discovery/test_EnumerateComputeNodes_StopDiscoveryService_EnumerateComputeNodesWorkWithoutError.vi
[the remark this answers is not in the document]
Prem-MysteryPro: I don't think the test is valid. We are asserting if we have zero compute nodes. In pipeline we don't have any remote connections so it would pass. But in reality different users might have different remote connection like plugin library & remote hardware.

Source/Runtime/Measurement Server.lvproj
jasonmreding, line 1: I generally try to not submit changes to project files where only the listed dependencies change in order to keep reviewers focused and avoid clutter in the revision history. Not a big deal, just something you have to micromanage with LV development and source control. Maybe one day these won't be treated as dirtying project changes.

Source/Runtime/Clients/Discovery V1/Get Deployment Target.vi
jasonmreding, line 1: Make this a shared reentrant VI.

jasonmreding, line 1: Rename `deployment_target` terminal to `deployment target` as LV code does not generally use snake case (nor does the code in this repo).

jasonmreding:
> TODO: Add automation tests.

I am OK for now relying on manual testing of these changes. The changes are small enough that I think the risk of introducing a regression failure is relatively small compared to the dev cost of spinning up a testing framework that can reliably test the functionality in an automated fashion.

If you want to create automated tests, here are some thoughts that might be useful:
- You can use the existing `Kill Discovery Service.vi` and `Restart Discovery Service.vi` to launch new instances/clusters of the discovery service. This would presumably allow to kill the service, change environment variables, and then relaunch the service such that the new environment variables take effect and various features are enabled/disabled. You could then reverse this process during test cleanup to put the system back into the original state.
- The test image is still running against InstrumentStudio 24.8. If you need new discovery service features, you would have to update the [SALT state](https://dev.azure.com/ni/DevCentral/_git/ni-central?path=/eng/iac/cfg_mgmt/salt/srv/salt/measlinkbot/win/measurementlink_install.sls&version=GBmain&line=5&lineEnd=6&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) for the test image and kick off the pipeline that rebuilds the image and deploys it to AWS before you could submit any new tests that rely on those features.
- You will likely have to write your own test utilities for registering/unregistering compute nodes.
- If you want to test using two local instances of the discovery service running on different ports, you will likely need to do that for one of them outside the use of the FGV which is used to cache the client connection. Many of those VIs are marked as private to the library so you would either need to duplicate some of the logic in the test utility or move them to community scope so the test libraries can access them. Another option would be to loopback on the same instance. This is simpler but not as representative of the real world use case you are trying to test.
- I don't know I will be of much help if you really want to test communication between two different hosts. Someone more familiar with use of containers in a CI pipeline might be more helpful. I imagine you might also be able to spin up a second ASW AMI, but this outside the scope of my experience so I probably wouldn't be able to offer up much help beyond what you could find yourself through researching on your own.

jasonmreding (approved):

PR #708, opened 2026-07-13
[releases/3.5] Bump dependencies to grpc-labview >= 1.7.0.1 and build numbers to 3.5.4.1

dixonjoel:

Bumps the versions in the package build specs to make our packages depend on grpc-labview >= 1.7.0.1. There were some breaking changes and we now need this latest version for our services not to be broken. Also bumping versions to 3.5.4.1 in anticipation of a release that will happen shortly.

Unbreak our services.

PR


Source/Build Specs/Measurement Plug-In SDK.vipb
```diff
@@ -17,9 +17,9 @@
   </Library_General_Settings>
   <Advanced_Settings>
     <Package_Dependencies>
-      <Additional_External_Dependencies>ni_measurement_plugin_sdk_examples &gt;=3.0.0.1</Additional_External_Dependencies>
-      <Additional_External_Dependencies>ni_measurement_plugin_sdk_generator &gt;=3.0.0.1</Additional_External_Dependencies>
-      <Additional_External_Dependencies>ni_measurement_plugin_sdk_service &gt;=3.0.0.1</Additional_External_Dependencies>
+      <Additional_External_Dependencies>ni_measurement_plugin_sdk_examples &gt;=3.5.4.1</Additional_External_Dependencies>
```
jasonmreding, line 20: I don't think the version bumps for these dependencies are strictly needed, but I don't know that it really hurts anything either. In general, I'm not sure if there is a ton of value in releasing these meta packages over and over again or not without any range restrictions. I believe VIPM will automatically install newer dependencies if they are available anyway. So I don't think it really buys you anything unless the number of dependencies/packages changes from one release to another. With that said, I'm OK leaving this as is if you like.

jasonmreding (approved):

Source/Build Specs/Measurement Plug-In SDK.vipb
dixonjoel: Ok. I'll leave as is.

